PR #2, opened 2016-09-15
[master] Add .gitignore

petricm:
Add a standard .gitignore plus the additional exclusion of directories that result from in-source builds


petricm:
Added, all above 'Build and install directories' are the github defaults for a c++ project.

PR #23, opened 2016-11-01
clang linker dynamic lookup of undefined symbols

gaede:


MarkusFrankATcernch:
Frank,

We cannot simply blind-out the errors in the build. The missing operators are in the code.
The question is if they are part of the correct library. This has to be sorted out.

Markus

andresailer:
Then we should add `-Wl,--no-undefined` to the compiler flags for dd4hep

PR #25, opened 2016-11-02
Clang 3.9 for gitlab CI

petricm:
This is just provisional, so that you are able to run clang on gitlab CI. I have done it now with root 6.08-rc1


petricm:
This PR also sanitizes the warnings from external packages, based on the patch provided by @andresailer

PR #43, opened 2016-11-17
rewritten MegatileLayerGrid class

danieljeans:
Rewritten MegatileLayerGridXY class, for use with the new SECAL05 drivers in lcgeo.
New version rather simpler, the complexity has been moved to the detector drivers.

note:
This is not backward compatible, and will break the one lcgeo detector driver (SEcal04Hybrid...) which uses this segmentation; it can be replaced by the new SEcal05 driver.


petricm:
Please check the [build](https://travis-ci.org/AIDASoft/DD4hep/builds/176645348) it is not compiling.

danieljeans:
Sorry, I was compiling against an older version.
I have fixed this in the latest commit (adding an extra constructor).
The check failed, but it seems to be because of taking too much time. My code seems to have compiled OK, as far as I can tell.
Regards,
Daniel.

Issue #30, opened 2016-11-04
DD4hep not working on El Capitan and MacOS Sierra


petricm:
Due to the [System Integrity Protection (SIP)](https://support.apple.com/en-us/HT204899), as well as preventing anyone from modifying system files, protected binaries no longer inherit linker environment variables. Meaning that setting the `DYLD_LIBRARY_PATH` will have no effects on spawned sub shells unless one [turns off (SIP)](http://www.macworld.com/article/2986118/security/how-to-modify-system-integrity-protection-in-el-capitan.html).[[1](http://dmtn-001.lsst.io/en/master/)][[2](http://stackoverflow.com/questions/39927235/alternative-for-the-dyld-library-path-trick-since-mac-os-10-11-el-capitan-with-s)]

petricm:
Just as reference Aliroot does not work either on El Capitan and Sierra [[1](https://dberzano.github.io/alice/install-aliroot/prereq-osx/#system_integrity_protection)], due to the same reasons and they suggest turning off SIP.

petricm:
One more addition, any system binary in:
/bin
/sbin
/usr
/System
will exhibit striping of DYDL_* however, except for /usr/local. Which mean if you copy /bin/sh /bin/bash /usr/bin/env to some custom location and add it at the beginning of PATH, you will partially get back the desired behavior.

Normal SIP behavior:
$ (export DYLD_FOO=foo; env) | grep foo
$
with binaries copied to new location and added to path
$ (export DYLD_FOO=foo; env) | grep foo
$ dyld: warning, unknown environment variable: DYLD_FOO
$ DYLD_FOO=foo

This still does not resolve the build problem, but might help.

petricm:
The above fix does not work, because MakeRootMap calls EXECUTE_PROCESS, and this does not really open a new shell, but executes the child process using operating system APIs directly. All arguments are passed VERBATIM to the child process. Thus no intermediate shell is used.

petricm:
Just a small update on this. The problem from the DD4hep side has been resolved, we can now "propagate" the DYLD_LIBRARY_PATH, however this only uncovered a problem in CLHEP cmake config, which does not set rpath.

petricm:
The description of the underlying problem
https://its.cern.ch/jira/browse/CLHEP-139

gaede:
Did you try and build CLHEP with 
cmake -D CMAKE_MACOSX_RPATH=ON ..

petricm:
[...]

PR #48, opened 2016-11-29
Drop ROOT 5 and fix problems on macOS

petricm:
This drops supports for ROOT 5 and removes the FindROOT.cmake. The setup is done via ROOTConfig.cmake which you now have to additionally provide via -DROOT_DIR=$ROOTSYS.

CMake files have been rewritten to use the new ROOTConfig.cmake functionality. This includes a new treatment for teveDispaly, which does not only depend on Eve but olso on Eve, Gui, Graf3d and RGL this was not done properly previously. 

This rewrite is now also partially macOS Sierra compatible, the compilation does not complain about missing libs and the basic tests work. This assuming you are using a newer version of CLHEP. The CI has been updated to 2.3.4.2. which has proper rpath treatment. 

It seems that the C++ part is now mac compatible the only problem remains the python DDG4 interface etc.. Where there is a problem with the on the fly ACLiC compilation and you get 

```
Error in <TMacOSXSystem::FindDynamicLibrary>: libDDCore[.so | .dll | .dylib | .sl | .dl | .a] does not exist in .:/cvmfs/clicdp.cern.ch/software/ROOT/6.08.00/x86_64-mac1012-clang80-opt/lib::/usr/local/lib:/usr/X11R6/lib:/usr/lib:/lib:/lib/x86_64-linux-gnu:/usr/local/lib64:/usr/lib64:/lib64:
+--------------------------------------------------------------------------------------------------------+
|  Failed to load DDG4 library:                                                                          |
|  DDG4.py: Failed to load the Geant4 library libDDCore:                                                 |
+--------------------------------------------------------------------------------------------------------+
```


petricm:
A bit of progress now it's only 
```
>>> from ROOT import gSystem
>>> gSystem.SetDynamicPath(os.environ['DD4HEP_LIBRARY_PATH'])
>>> gSystem.Load("libDDG4Plugins")
cling::DynamicLibraryManager::loadLibrary(): dlopen(/Users/lcdci/DD4hep/lib/libDDG4Plugins.0.19.dylib, 9): Library not loaded: @rpath/libCLHEP-2.3.4.2.dylib
  Referenced from: /Users/lcdci/DD4hep/lib/libDDG4Plugins.0.19.dylib
  Reason: image not found
```
I don't know why the libraries not loaded automatically

petricm:
This now has all the C++ problems solved with correct rpath installation on macOS as described [here](https://cmake.org/Wiki/CMake_RPATH_handling#Always_full_RPATH). The first pass-though of cmake generated DD4HEP_LIBRARY_PATH which is not dropped in sub-processes. And thisdd4hep.sh generates it also when initialized. The python objects DDD4hep.sh and DDG4.sh have been updated to search in this new path for dependencies. This brings us to: 66% tests passed, 35 tests failed out of 104. And the reasons of the failed tests is the missing DYLD_LIBRARY_PATH in python when things get loaded look [here](https://gitlab.cern.ch/mpetric/DD4hep/builds/125164).

petricm:
It seem to be partially solved. As I don't find any other solution it will be for the time being required that if you want to run DD4hep on mac that you will have to provide another python which is not the system python. Under this condition DD4hep+TinyXML work, there some minor issue with Xerces + ACLiC, but does not really prevent you from using the thing.

petricm:
Closes #30

PR #56, opened 2016-12-08
add graphical material scan utility

danieljeans:
new utility to produce 2-d scans of material (as TH2F in a TFile).
maps of x0, lambda, and individually for the various material types.
range in (x,y,z), slice axis, number of bins etc are set on command line.

### DIFF
--- a/UtilityApps/CMakeLists.txt
+++ b/UtilityApps/CMakeLists.txt
@@ -24,6 +24,8 @@ dd4hep_add_executable( print_materials src/print_materials.cpp USES DDRec )
 #-----------------------------------------------------------------------------------
 dd4hep_add_executable( materialScan    src/materialScan.cpp USES DDRec )
 #-----------------------------------------------------------------------------------
+dd4hep_add_executable( graphicalMaterialScan src/graphicalMaterialScan.cpp USES DDRec )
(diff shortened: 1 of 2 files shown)


UtilityApps/CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ dd4hep_add_executable( print_materials src/print_materials.cpp USES DDRec )
 #-----------------------------------------------------------------------------------
 dd4hep_add_executable( materialScan    src/materialScan.cpp USES DDRec )
 #-----------------------------------------------------------------------------------
+dd4hep_add_executable( graphicalMaterialScan src/graphicalMaterialScan.cpp USES DDRec )
```
petricm: To be explicit, this depends on ROOT, so there should be USES DDRec ROOT

MarkusFrankATcernch:
The SVN keyword is not really a problem. We should only slowly, but surely remove them
whenever we stumble across them....

UtilityApps/CMakeLists.txt
danieljeans: Thanks for the feedback. Excuse my ignorance about the workflow: is this a request for me to fix it, or a notification that you have done so?

PR #64, opened 2016-12-14
use DD4HEP_LIBRARY_PATH and full lib path on mac

gaede:
This fixes running python simulation programs with plugins on macos 10.12.

### DIFF
--- a/cmake/thisdd4hep.sh
+++ b/cmake/thisdd4hep.sh
@@ -45,9 +45,11 @@ dd4hep_add_library_path()    {
     if [ @USE_DYLD@ ];
     then
         if [ ${DYLD_LIBRARY_PATH} ]; then
-	    export DYLD_LIBRARY_PATH=${path_prefix}:$DYLD_LIBRARY_PATH;
+            export DYLD_LIBRARY_PATH=${path_prefix}:$DYLD_LIBRARY_PATH;
+            export DD4HEP_LIBRARY_PATH=${path_prefix}:$DD4HEP_LIBRARY_PATH;
         else
             export DYLD_LIBRARY_PATH=${path_prefix};
+            export DD4HEP_LIBRARY_PATH=${path_prefix};
         fi;
     else
         if [ ${LD_LIBRARY_PATH} ]; then
@@ -63,6 +65,10 @@ dd4hep_parse_this ${BASH_ARGV[0]} DD4hep;
 #
 # These 3 are the main configuration variables: ROOT, Geant4 and XercesC
 # --> LCIO & Co. are handled elsewhere!
+
+if [ -z $ROOTSYS ]; then
+    export ROOTSYS=`dirname @ROOT_DIR@`
+fi;
 export Geant4_DIR=@Geant4_DIR@;
 export XERCESCINSTALL=@XERCESC_ROOT_DIR@;
 #
--- a/GaudiPluginService/src/PluginService.cpp
+++ b/GaudiPluginService/src/PluginService.cpp
@@ -162,7 +162,9 @@ namespace Gaudi { namespace PluginService {
       const char sep = ';';
 #else
 #ifdef APPLE
-      const char* envVar = "DYLD_LIBRARY_PATH";
+      //fg: on macos >10.11 we cannot rely on DYLD_LIBRARY_PATH any more
+      //    and therefore use consistently DD4HEP_LIBRARY_PATH instead
+      const char* envVar = "DD4HEP_LIBRARY_PATH";
       const char sep = ':';
 #else
       const char* envVar = "LD_LIBRARY_PATH";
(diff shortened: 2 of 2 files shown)


petricm:
Please explain in more detail, what was not working, since all the test work in the previous state.

petricm:
This breaks builds on the check here mac https://gitlab.cern.ch/mpetric/DD4hep/builds/144030

gaede:
This PR fixes running DDG4 simulations via python on the mac, e.g. with ./examples/./DDG4/examples/SiDSim.py  or - more importantly - with ddsim from lcgeo.
This is not included in the current tests ( and probably should be).
Will check why my fix seems to break the tests ...

cmake/thisdd4hep.sh
```diff
@@ -63,6 +65,10 @@ dd4hep_parse_this ${BASH_ARGV[0]} DD4hep;
 #
 # These 3 are the main configuration variables: ROOT, Geant4 and XercesC
 # --> LCIO & Co. are handled elsewhere!
+
+if [ -z $ROOTSYS ]; then
+    export ROOTSYS=`dirname @ROOT_DIR@`
```
petricm: This might not be ok, since this can be actually $ROOTSYS/cmake. When you do cmake -DROOT_DIR=.... this can  be absolute path to the ROOTConfig.cmake

gaede:
exactly - and `dirname @ROOT_DIR@` removes the last path directory, i.e. '/cmake' which is why this works. All tests passed on my Mac and all tests passed on CI - until timeout.

petricm:
Yes, but when you specify -DROOT_DIR=\<prefix\> [cmake should search](https://cmake.org/cmake/help/v3.0/command/find_package.html) in 
```
<prefix>/                                               (W)
<prefix>/(cmake|CMake)/                                 (W)
```
How do you know which is the case did the user specify?

gaede:
I tried -DROOT_DIR=\<prefix\> and it did not work, so I concluded
that ROOT_DIR should always be $ROOTSYS/cmake.

Anyways, do you have a better solution for initializing ROOT in thisDD4hep.sh ?


----- Original Message -----

petricm:
Best would be that the ROOTConfig.cmake would set the base folder, but since it doesn't I think it's best to infer ROOTSYS from a variable that is set by ROOTConfig.cmake, e.g. ROOT_Core_LIBRARY and do a regex replace /lib/libCore.* . This is then not dependent on the user input.

gaede:
I agree. It would be much better if ROOTConfig.cmake would set the base folder -  in fact the iLCSoft standard of putting the PackageConfig.cmake in the package root folder seems much cleaner and more robus, as then one does not have to do this kind of gymnastics.

petricm:
The default settings on UNIX systems for find_package are not in the base folder, please check [here](https://cmake.org/cmake/help/v3.0/command/find_package.html). It should be in:
```
<prefix>/(lib/<arch>|lib|share)/cmake/<name>*/          (U)
<prefix>/(lib/<arch>|lib|share)/<name>*/                (U)
<prefix>/(lib/<arch>|lib|share)/<name>*/(cmake|CMake)/  (U)
```
So iLCSoft standard is contrary to the CMake Unix standard location. Unless you think the Windows location is the way to go? I don't think that this export ROOTSYS issue is resolved.

gaede:
We have adopted CMake for iLCSoft much before anything like a 'Unix' standard even existed.
And in this particular case, I am happy to for once acknowledge that Windows is best ;-)
Putting the PkgConfig.cmake in the package root folder is the simplest and most robust solution.
PS: please feel free to provide a more generic solution to find the ROOTSYS as you suggested above.

petricm:
There is already a more robust solution in the function [fill_dd4hep_library_path](https://github.com/AIDASoft/DD4hep/blob/master/cmake/DD4hepBuild.cmake#L1442)

petricm:
You said:
`We have adopted CMake for iLCSoft much before anything like a 'Unix' standard even existed.`
The first [commit](https://github.com/iLCSoft/iLCUtil/commit/7a790da144a755e75c5099d810958714225620b8) to iLCUtil  is from 2011 this is 3 years after CMake version 2.6 is release where already this standard is introduced. 

But the more important point is that DD4hep was started after CMake has standardized this, thus DD4hep should follow this standard. The version of CMake that SLC6 came with had also already this standard.

gaede:
As you seem to be so keen on discussing the history of CMake in iLCSoft: here is the first commit of CMake files for Marlin [Apr 23, 2007](https://github.com/iLCSoft/Marlin/commit/016c88ff8b75af5dc458227f947141a12f84994b) - including the MarlinConfig.cmake.
So we have indeed used CMake before most (all?) other groups in HEP and before any 'standards' had been introduced ...

Apart from the fact that this is completely off-topic in this PR that dealt among other things with the location of ROOT and the ROOTConfig.cmake ( also 'non-standard'), I still believe that our approach of putting the config files in the package root dir makes more sense that the Unix 'standard'. The reason is that we are not installing iLCSoft in system directories such as /usr/lib/ or /user/local/lib but in dedicated self-contained directories - together with the source code. In such a scenario it is much simpler to deal with the configuration files in the root directory, as then you only need one variable to define everything for the given package - as this discussion and previous issues we had with CLHEP and Geant4 cmake config files we had in the past show.

petricm:
The only thing that is important is that DD4hep is younger then the standard, so it should respect it. 

Regarding Marlin, as long as you use a version of CMake lower then 2.6 it's ok not to respect the standard.

petricm:
CMake 2.4.3 released 20 Sep 2006 (which is before Marlin), already searched in lib/\<name\> check [this](https://github.com/Kitware/CMake/blob/v2.4.0/Source/cmFindPackageCommand.cxx#L287), it was just badly documented.

PR #65, opened 2016-12-14
add map cellID of surface <--> vector of cellID of neighbouring surfa…

simoniel:
…ces for ZPlanar and ZPetalDisk data struct

### DIFF
--- a/DDRec/include/DDRec/DetectorData.h
+++ b/DDRec/include/DDRec/DetectorData.h
@@ -1,6 +1,7 @@
 #ifndef DDRec_DetectorData_H_
 #define DDRec_DetectorData_H_
 
+#include <map>
 #include <bitset>
 #include <ostream>
 
@@ -155,6 +156,8 @@ namespace DD4hep {
       } ;
 
       std::vector<LayerLayout> layers ;
+      std::map<DD4hep::long64 , std::vector<DD4hep::long64 > > mapNeighbours;
+
 
     } ;
     typedef StructExtension<ZPlanarStruct> ZPlanarData ;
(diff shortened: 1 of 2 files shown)


gaede:
I would prefer a more generic solution/class that deals with topological surface associations and not put this into DetectorData that so far is completely independent of the surface concept.
What exactly should be stored in this map - and where is this information going to be filled ?
Can this information maybe be created by the SurfaceManager class ?
Do you not foresee neighbour relation ships between surfaces of different detectors ?

simoniel:
the map is filled in lcgeo in the detector driver with cellID information of each cell and vector of cellID of neighboring cells. It is nice to have it there because to find the neighbouring surfaces you  need to know the logic the cellIDs have been filled with, so in the detector drivers I can fill my information inside pre-existing loops (and in case of change of the filling logic is everything in one place). In principle this map object can also not belong to DetectorData and be a different type (i can define it directly in lcgeo?). At the moment the neighbouring surfaces are the ones on the same layer, we are considering to add also the ones on the next layer. As for your request for a more generic implementation, do you have anything specific in mind?

gaede:
Ok, if I understand this correctly, the data in the map is somewhat algorithm specific. It should then clearly not be added to the existing DetectorData classes.  In order to continue,  we could simply add a dedicated struct, e.g:
~~~ 
/** holds maps of nearest neighbour surfaces in the same and next (inner/outer ?) layer
 *  of a tracking detector...
*/
struct NeighbourSurfacesStruct {
  std::map<DD4hep::long64 , std::vector<DD4hep::long64 > > sameLayer ; 
  std::map<DD4hep::long64 , std::vector<DD4hep::long64 > > nextLayer ;
} ;
typedef StructExtension<NeighbourSurfacesStruct> NeighbourSurfacesData ;
~~~

A more generic solution would be a topology mapping that works across detectors and provides all the surfaces that could be reached from given surface ( assuming  a given B-field and momenta cuts ).
This is clearly more difficult and not so easy to define (and implement).

Something like the above should allow you to continue your patrec developments for now.

PR #68, opened 2016-12-20
Add Multi-threading conditions example, Fix other examples

MarkusFrankATcernch:


MarkusFrankATcernch:
Marko,

could you please have a look to the changes of cmake/DD4hepBuild.cmake?
I am not sure if I got it right.

cmake/DD4hepBuild.cmake
```diff
@@ -71,11 +71,22 @@ macro(dd4hep_set_compiler_flags)
     add_definitions(-DDD4HEP_USE_STDCXX=11)
   endif()
 
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wshadow -Wformat-security -Wno-long-long -Wdeprecated")
```
petricm: This line is too much. Since it at the top [line 38](https://github.com/AIDASoft/DD4hep/blob/357a977587ae6783be0946223518850b53ac0d42/cmake/DD4hepBuild.cmake#L38) you set the flags you want, then they are checked if the compiler supports them and added to CMAKE_CXX_FLAGS ([line 40-48](https://github.com/AIDASoft/DD4hep/blob/357a977587ae6783be0946223518850b53ac0d42/cmake/DD4hepBuild.cmake#L40))

cmake/DD4hepBuild.cmake
```diff
@@ -71,11 +71,22 @@ macro(dd4hep_set_compiler_flags)
...
 
-  IF(("${CMAKE_CXX_COMPILER_ID}" EQUAL "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" EQUAL "Clang"))
+  CHECK_CXX_COMPILER_FLAG("-fdiagnostics-color=always" FLAG_COLOR)
```
petricm: This `fdiagnostics-color=always` is obsolete since it's already added to line 38 at the top. And it preferred that it is set to auto, due to compatibility with compilation inside emacs.

cmake/DD4hepBuild.cmake
```diff
@@ -71,11 +71,22 @@ macro(dd4hep_set_compiler_flags)
...
+   
+  IF( "${CMAKE_CXX_COMPILER_ID}" EQUAL "Clang" )
     SET(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-undefined")
   ENDIF()
 
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-undefined -pthread")
+  endif()
+
```
petricm: Regarding pthread, I would suggest using the [CMake function](https://cmake.org/cmake/help/v3.0/module/FindThreads.html) `find_package (Threads)` and the do `target_link_libraries (DDApp ${CMAKE_THREAD_LIBS_INIT})`. Along the lines discuses [here](http://stackoverflow.com/questions/1620918/cmake-and-libpthread) and [here](http://stackoverflow.com/questions/5395309/cmake-and-threads)

MarkusFrankATcernch:
Not so sure about the pthread issue....
You normally also need to compile with -pthread, not only link.
-pthread is said to do more than only "-lpthread".

petricm:
Of course you need to compile and link with pthread. It was only a suggestion regarding the detection of the presence of the threading library.

cmake/DD4hepBuild.cmake
```diff
@@ -71,14 +73,20 @@ macro(dd4hep_set_compiler_flags)
...
+    set ( CMAKE_CXX_FLAGS           "${CMAKE_CXX_FLAGS} -pthread")
+    SET ( CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -pthread")
+  elif ( CMAKE_THREAD_LIBS_INIT )
+    SET ( CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} ${CMAKE_THREAD_LIBS_INIT}")
+  endif()
 
-  IF(("${CMAKE_CXX_COMPILER_ID}" EQUAL "GNU") OR ("${CMAKE_CXX_COMPILER_ID}" EQUAL "Clang"))
-    SET(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--no-undefined")
-  ENDIF()
+  if( "${CMAKE_CXX_COMPILER_ID}" EQUAL "Clang" )
```
petricm: You forgot to re-enable `-Wl,--no-undefined` for GNU

MarkusFrankATcernch:
Forget about travis. On gitlab all passed....

PR #71, opened 2016-12-22
fix tests for new elements treatment

gaede:


petricm:
Hi, 
is the elements.xml created as a dump from ROOT? If so, could you please add a comment field somewhere with the reference to the ROOT version.

PR #72, opened 2016-12-23
 prepare for v00-20 (release notes and version )

gaede:
- still need resolution for failing tests
  - or revert to old treatment of elements ...
 - release notes created from
    git log --pretty=format:"%aN %ad %n  - %s" --date=short
   => should have squashed commits with more
   meaningful commit notes in the future ...


gaede:
disabled the deletion of the TGeo element table for now, so that the example tests work again
 => should be fixed after the holidays

PR #73, opened 2017-01-10
call find_package for CLHEP in DDRec if needed

gaede:
treatment of CLHEP with cmake has again changed in CLHEP 2.3.4.3 ( needed by geant4 10.03)

### DIFF
--- a/DDRec/CMakeLists.txt
+++ b/DDRec/CMakeLists.txt
@@ -18,6 +18,16 @@ dd4hep_package(DDRec
   INSTALL_INCLUDES include/DDRec)
 
 #---Add Library-------------------------------------------------------------
+
+#fg: for CLHEP 2.3.4.3 we need o call again the find_package
+if(DD4HEP_USE_GEANT4)
+  if(NOT Geant4_builtin_clhep_FOUND)
+    find_package( CLHEP REQUIRED )
+    message( STATUS "!!! calling find_package( CLHEP REQUIRED ) !!!" )
+  endif()
+endif()
+
+
 if(DD4HEP_USE_GEAR)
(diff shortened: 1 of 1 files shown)


petricm:
Why is this only affecting DDRec and not other parts of DD4hep? And this if you look at Travis breaks 10.2.2. if I am not mistaken.

gaede:
I don't know why it only affects DDRec - maybe it is sufficient that find_package is called once ?
In any case I tested on SL6 w/ CLHEP 2.3.4.3  (this is where the problem first occurred) and on Macos. Can you check why your Travis build fails ?
It should either not call find_package(CLHEP) because geant4 was built with internal CLHEP or - if not - the path to CLHEPConfig.cmake muste be given...

petricm:
Can you please try to move `add_subdirectory(DDRec)` after  `include(${Geant4_USE_FILE} )` because the Geant4 use file propagates CLHEP I think this is the issue. 

If you use Geant you should not provide -DCLHEP_DIR=.... in cmake command since the use file propagates how and which CLHEP was used. In principle the same also hold for XercesC since Geant 10.3 if my bug reports was properly resolved (https://bugzilla-geant4.kek.jp/show_bug.cgi?id=1877) but I did not have time to check.

gaede:
Moving add_`subdirectory(DDRec)` after `include(${Geant4_USE_FILE} )` causes a different problem with cmake: `By not providing "FindDDREC.cmake" in CMAKE_MODULE_PATH ...`.

Anyways, I will close this PR as even though I can build DD4hep with it, I get problems in pretty much all dependent packages ( missing `CLHEP::CLHEP IMPORTED target`  ... ).

This needs more thought on how to best resolve the issue. 
Will go back to the previous CLHEP for now.

PR #83, opened 2017-02-01
Fixed installation destinations of cmake and doxygen files.

whit2333:
The cmake file goes in `${CMAKE_INSTALL_PREFIX}/cmake`
rather than polluting the installation prefix.

The documentation goes into `${CMAKE_INSTALL_PREFIX}/share/DD4hep`
I was not sure if it should be `share/DD4hep/doc/html` or
`share/DD4hep/html`, but I thought the former was more clear since the
`share/DD4hep` directory can hold other things beyond documentation.
For example, "examples" would go into `share/DD4hep/examples`, hence the extra `doc` directory.

### DIFF
--- a/cmake/DD4hepDoxygen.cmake
+++ b/cmake/DD4hepDoxygen.cmake
@@ -10,7 +10,7 @@ if(DOXYGEN_FOUND)
                     COMMENT "Generating API documentation with Doxygen" VERBATIM)
 
   install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/html
-    DESTINATION doc
+    DESTINATION share/DD4hep/doc
     )
 
 else()
--- a/cmake/DD4hepMacros.cmake
+++ b/cmake/DD4hepMacros.cmake
@@ -76,7 +76,7 @@ MACRO( GENERATE_PACKAGE_CONFIGURATION_FILES )
                 CONFIGURE_FILE( "${PROJECT_SOURCE_DIR}/cmake/${arg}.in"
                                 "${PROJECT_BINARY_DIR}/${arg}" @ONLY
                 )
-                INSTALL( FILES "${PROJECT_BINARY_DIR}/${arg}" DESTINATION . )
+                INSTALL( FILES "${PROJECT_BINARY_DIR}/${arg}" DESTINATION cmake )
                 #IF( EXISTS "${_current_dir}/MacroCheckPackageLibs.cmake" )
                 #    INSTALL( FILES "${_current_dir}/MacroCheckPackageLibs.cmake" DESTINATION cmake )
                 #ENDIF()


petricm:
As much as I agree with you on this matter, the developers have not reached a consensus on this topic on one of the last developers meetings. I would suggest the following as a compromise:

- introduce a cmake flag `SYSTEM_INSTALL`
- if this flag is set to ON the install uses the standard install location as in your PR
- do not install the cmake files into the `cmake` folder but rather  `lib<arch>/DD4hep`
- the default value for `SYSTEM_INSTALL` should be `OFF`

If you implement a solution along these lines we can merge it.

whit2333:
I prefer `lib/DD4hep` too.  I guess `SYSTEM_INSTALL` could be done but I really only want to encourage good defaults. :)

Right now it looks like ilcsoft threw up cmake files in my `/usr/local` and I thought this project would be a good place to discourage this behavior.

gaede:
ilcsoft does not install anything in /usr/local unless you tell it to - in fact the standard (default) installation is with the source ...

whit2333:
In source builds/installs are another issue. IMO they are a bad idea.  
Keeping the code, build, and install separate makes life some much more 
pleasant.

The default prefix should always be /usr/local. All the excessive cmake 
macros can be avoided if done this way and one needs only manage a 
single directory added to LD_LIBRARY_PATH, PATH, etc...


On Wed, Feb 01, 2017 at 09:06:59AM -0800, Frank Gaede wrote:
>ilcsoft does not install anything in /usr/local unless you tell it to - in fact the standard (default) installation is with the source ...
>
>
>
>> On 1 Feb 2017, at 17:55, Whitney Armstrong <notifications@github.com> wrote:
>>
>> I prefer lib/DD4hep too. I guess SYSTEM_INSTALL could be done but I really only want to encourage good defaults. :)
>>
>> Right now it looks like ilcsoft threw up cmake files in my /usr/local and I thought this project would be a good place to discourage this behavior.
>>
>> —
>> You are receiving this because you are subscribed to this thread.
>> Reply to this email directly, view it on GitHub, or mute the thread.
>>
>
>
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/AIDASoft/DD4hep/pull/83#issuecomment-276716773

-- 
Whitney R. Armstrong

petricm:
Please let's not discuss iLCSoft configurations and installations here. 

It would be of added value to have an option/flag to install the cmake stuff and doc to the "standard" location as describes in my previous comment. If it turns out that this does not break any other dependent software, we might revisit the topic on a future developers meeting.

gaede:
I agree and support your proposal:
  
> -   introduce a cmake flag SYSTEM_INSTALL
> - if this flag is set to ON the install uses the standard install location as in your PR
> - do not install the cmake files into the cmake folder but rather lib<arch>/DD4hep
> - the default value for SYSTEM_INSTALL should be OFF

whit2333:
Checkout my branch `system_install` https://github.com/whit2333/DD4hep/tree/system_install

Not sure if I should submit a PR on this yet...

petricm:
Please make PR and put WIP in the title, we can then discuss the implementation in the PR.

PR #84, opened 2017-02-03
Evolve alignments and conditions

MarkusFrankATcernch:
If you use alignments and conditions be careful.
-- Save your work before the update!
-- There are backwards incompatible changes!


MarkusFrankATcernch:
Gitlab worked! finally ...

PR #90, opened 2017-02-10
Fix coverity defects in DDRec

andresailer:
Note: using boost class to restore the ostream state, DDRec now depends on boost


MarkusFrankATcernch:
Andre,
we merge this one too?
Markus

gaede:
looks good to me.

PR #93, opened 2017-02-10
Generate multiple vertices

gaede:
Allow event readers to create more than one vertex per event - this should be possible as most generator formats allow to specify more than one event vertex.
The current implementation still only creates one vertex per event for HepMC, LCIO and HepEvt.
The new feature will first be used for Guinepig files with non-prompt e+e- pairs.


MarkusFrankATcernch:
Can we merge this request ?

andresailer:
Fine for me

PR #96, opened 2017-02-13
Fix example DDG4/examples/readHEPMC.py and allow access to STL contai…

MarkusFrankATcernch:
…ners from DD4hep python interface


MarkusFrankATcernch:
Gitlab worked. Accept merge request.

gaede:
thanks.

PR #98, opened 2017-02-14
System install [WIP]

whit2333:


petricm:
Please just rebase your changes regarding the install location on the AIDASoft/DD4hep master without NoSegmentation.

gaede:
I would prefer to have a discussion on these changes in the next developers meeting in order to understand, if we want to got ahead with them. Also, it seems that they do not compile right now ...

PR #105, opened 2017-02-17
Update to allow the externalization of the XML and the ComponentPrope…

MarkusFrankATcernch:
…rty sub-packages


MarkusFrankATcernch:
Gitlab worked with all tests....

PR #107, opened 2017-02-23
InputAction: Add parameter map property

andresailer:
Add parameter map property to pass arbitrary parameters to event readers if necessary

Then we can do something like this in an eventReader class
``` C++
Geant4EventReaderGuineaPig::setParameters( std::map< std::string, std::string > & parameters ) {
  // int m_part_num
  _getParameterValue( parameters, "ParticlesPerEvent", m_part_num, -1);
}
```


gaede:
Excellent. Can you also provide an example for how to use this with python/ddsim ?

andresailer:
Sure in python it will be somewhere
``` python
gen.Parameters =  {"ParticlesPerEvent":"10"}
```
However, we can wrap something around the dictionary to have an explicit parameter in ddsim itself.
The GuineaPigReader needs to be adapted as well, obviously.

PR #110, opened 2017-02-24
multiple vertices per interaction continued

gaede:
- change Geant4PrimaryInteraction::VertexMap to
  std::map<int,std::vector<Vertex*>>

### DIFF
--- a/DDG4/include/DDG4/Geant4Primary.h
+++ b/DDG4/include/DDG4/Geant4Primary.h
@@ -100,11 +100,11 @@ namespace DD4hep {
       Geant4PrimaryInteraction& operator=(const Geant4PrimaryInteraction& c) = delete;
 
     public:
-      typedef Geant4Particle                    Particle;
-      typedef Geant4Vertex                      Vertex;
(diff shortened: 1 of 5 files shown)


andresailer:
Did you check that this will produce the exact same result with existing stdhep files?

gaede:
@andresailer: Of course I ran some tests with existing stdhep files.

@MarkusFrankATcernch can you have a quick look, if you are happy with these changes ?
It was a bit confusing that sometimes the mask is used as key in `Geant4PrimaryInteraction::VertexMap` and sometimes simply a counter w.r.t inserted vertices.

MarkusFrankATcernch:
Looks OK. If the intended functionality is preserved can only be done with tests.

At the end what matters is: 
In the event of multiple inputs identified each identified by its mask, 
it must be possible to uniquely associate also the outputs (hits) to the corresponding
primary interactions. This was the intention for the masks and must be preserved.

gaede:
Ok, I will merge this then. We need to of course run tests...

PR #115, opened 2017-03-02
Fix problem with DetElement transformations.

MarkusFrankATcernch:


MarkusFrankATcernch:
Gitlab worked for all platforms.
Andre, could you please re-run your test job to confirm that the problem disappeared?
Frank, please check if the deprecated messages from DD4hep internal calls disappeared.

andresailer:
Yes, the surfaces are created at the right place and with the correct inner/outer material again and all the hits are confirmed to be on the surface.

gaede:
Also the deprecated calls have disappeared - incl. the external one from ```DetElement.localToWorld()``` - is this expected ?

MarkusFrankATcernch:
DetElement::localToWorld should still issue a warning:

/// Transformation from local coordinates of the placed volume to the world system
bool DetElement::localToWorld(const Position& local, Position& global) const {
  DD4HEP_DEPRECATED_CALL("DetElement","DetElement::nominal() member functions",__PRETTY_FUNCTION__);

MarkusFrankATcernch:
Is there a release existing with the wrong implementation?

If yes, we should obsolete it before someone spends significant amounts of time to understand the funny behaviour, since this was a quite severe bug.

gaede:
Yes, we had a HEAD release from 23.02 and have been chasing 'issues' with the reconstruction in this release.  I will create a new one with the fix and then we run some tests...

andresailer:
Was there any one change responsible for the issue, or was it a really complex combination?

MarkusFrankATcernch:
This was it:

-    a->worldTrafo = parent.nominal()->worldTrafo;
-    a->worldTrafo.MultiplyLeft(&a->detectorTrafo);
+    //a->worldTrafo = parent.nominal()->worldTrafo;
+    //a->worldTrafo.MultiplyLeft(&a->detectorTrafo);
+    a->worldTrafo = a->detectorTrafo;
+    a->worldTrafo.MultiplyLeft(&parent.nominal()->worldTrafo);

The 2 matrices were multiplied in the wrong order.....

andresailer:
Ok, thanks.

gaede:
So just to confirm: I still get the Depracted warning from the external code in ```bool DetElement::localToWorld(const Position& local, Position& global) const```.
(Something has gone wrong in my previous test).
Do we need to change this function call - and if so, how ?

MarkusFrankATcernch:
Very simple:
DetElement::localToWorld(local,global)  ==> DetElement::nominal().localToWorld(local,global)

The problem is that introducing the shortcut at the level of the DetElement would be very misleading in the presence of real alignment constants

andresailer:
By the way we can use `[[gnu::deprecated]]` to emit warnings about deprecation at compile time.
works with gcc and should work with clang as well (http://clang.llvm.org/docs/AttributeReference.html#deprecated-gnu-deprecated) (just [[deprecated]] in c++14)
e.g.:
```
      [[gnu::deprecated]]
      bool localToParent(const Position& local, Position& parent) const;
```
produces
```
warning: ‘bool DD4hep::Geometry::DetElement::localToParent(const Position&, DD4hep::Geometry::Position&) const’ is deprecated [-Wdeprecated-declarations]
  _det.localToParent( local, global);
                                   ^
In file included from DD4hep/HEAD/DDRec/include/DDRec/Surface.h:6:0,
                 from DD4hep/HEAD/DDRec/src/Surface.cpp:1:
DD4hep/HEAD/DDCore/include/DD4hep/Detector.h:408:12: note: declared here
       bool localToParent(const Position& local, Position& parent) const;
            ^~~~~~~~~~~~~
````

PR #119, opened 2017-03-09
UserParticleHandler: fix logic for long lived particles with pre-assi…

andresailer:
…gned decays that end in the calorimeter

fixes an issue where the MonteCarlo history becomes inconsistent with the history from the generator because the original particle entry is not updated


petricm:
@shaojunlu does this solve the problems you observed?

PR #120, opened 2017-03-10
Fix various issues mentioned in this weeks meeting.

MarkusFrankATcernch:


DDCore/include/DD4hep/objects/VolumeManagerInterna.h
```diff
@@ -54,7 +54,7 @@ namespace DD4hep {
       /// Placement identifier
       VolumeID     identifier;
       /// Ignore mask of the placement identifier
-      [[gnu::deprecated("This member variable might get axed if it is not used, please tell us if you do")]]
+      //[[gnu::deprecated("This member variable might get axed if it is not used, please tell us if you do")]]
```
petricm: Why is just commented out?

DDCore/src/plugins/VolumeMgrTest.cpp
```diff
@@ -81,99 +84,177 @@ VolIDTest::VolIDTest(LCDD& lcdd, DetElement sdet, size_t depth) : m_det(sdet) {
...
+      //bool is_sensitive = place.volume().isSensitive();
+      //if ( is_sensitive || !child_ids.empty() )  {
+        checkVolume(detector, place, child_ids, child_chain);
+        //}
+      walkVolume(detector, place, child_ids, child_chain, depth+1, mx_depth);
     }
   }
 }
-
+#if 0
```
petricm: Whya re you keeping this in the code? This was replaced with walkvolume?

DDCore/include/DD4hep/objects/VolumeManagerInterna.h
andresailer: Did you forget to uncomment this, or is the `mask` no longer being phased out?

DDCore/src/plugins/VolumeMgrTest.cpp
MarkusFrankATcernch: This test will be further elaborated in the future, but I do not yet know exactly what should go there. So let's keep it.

DDCore/include/DD4hep/objects/VolumeManagerInterna.h
MarkusFrankATcernch: This is intentional. The mask very likely cannot go away.

andresailer: Then you could have just completely removed the line.

PR #134, opened 2017-03-24
Set verbose true for G4EmSaturation to printout Birks coefficient.

shaojunlu:
Added printout information for user to check the Birks Law usage at ddsim/Geant4 runtime.
It will printout only once in the first event like this.
### G4EmSaturation::FindBirksCoefficient Birks coefficient for G4_POLYSTYRENE  0.07943 mm/MeV

BEGINRELEASENOTES
- Set verbose true for G4EmSaturation to printout Birks coefficient.

ENDRELEASENOTES

### DIFF
--- a/DDG4/src/Geant4StepHandler.cpp
+++ b/DDG4/src/Geant4StepHandler.cpp
@@ -114,9 +114,10 @@ G4ThreeVector Geant4StepHandler::globalToLocalG4(const G4ThreeVector& global)  c
 /// Apply BirksLaw
 double Geant4StepHandler::birkAttenuation() const    {
 #if G4VERSION_NUMBER >= 1001
-  static G4EmSaturation s_emSaturation(0);
+  static G4EmSaturation s_emSaturation(1);
 #else
   static G4EmSaturation s_emSaturation();
+  s_emSaturation.SetVerbose(1);
 #endif
 
   double energyDeposition = step->GetTotalEnergyDeposit();


andresailer:
Please fill the releaseNotes block

shaojunlu:
I have added the release notes, and rebased to the current master HEAD.

DDG4/src/Geant4StepHandler.cpp
```diff
@@ -125,6 +125,7 @@ double Geant4StepHandler::birkAttenuation() const    {
   const G4Track* trk = step->GetTrack();
   const G4ParticleDefinition* particle = trk->GetDefinition();
   const G4MaterialCutsCouple* couple = trk->GetMaterialCutsCouple();
+  s_emSaturation.SetVerbose(true);
```
andresailer: a) SetVerbose takes an int, http://www-geant4.kek.jp/Reference/10.02.p02/classG4EmSaturation.html
b) It would be better to set the argument in line 117 to `1`, and only call `SetVerbose(1)` for older geant versions. by putting the call in the if/else pre-processor block.
Then we can avoid calling this function for every step at least in newer geant4 versions.

shaojunlu: I have updated it as you suggested.
Should I keep this pull request as the current?
Or should I remove this pull request, and generate a new one to include the last update only?

andresailer: `SetVerbose` is still called with a boolean instead of an int.
You could "squash" the two commits into one and force push to this branch, e.g., by using `git rebase --interactive`

shaojunlu:
Now I have the updated pull request after squash.

andresailer:
Thanks!

PR #140, opened 2017-03-30
clarify documentation for CartesianFields

gaede:
BEGINRELEASENOTES
- clarify documentation for CartesianField and implementations
- make clear that  void fieldComponents() has to add
   the new field to the given field vector
ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/Fields.h
+++ b/DDCore/include/DD4hep/Fields.h
@@ -30,7 +30,7 @@ namespace DD4hep {
     // Forward declarations
     typedef Position Direction;
 
-
+    /// Base class describing any field with 3D cartesian vectors for the field strength.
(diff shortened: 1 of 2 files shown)


MarkusFrankATcernch:
All fine with me. But the doxygen format looks wrong:
I Think it MUST be like this:

  /// Handle class describing any field with 3D cartesian vectors for the field strength.
  /** More blabla

The first line is the short description, which doxygen puts into the 
class list....

gaede:
Added requested change - will merge

MarkusFrankATcernch:
Just for the sake of it:
Whenever you get hold of a header, which does not have this one line explanation,
please add it. And, of course remove the Id$ fields ....

Thank you!

Issue #141, opened 2017-03-31
MCTruth link broken in multisegmentation/collection simulation


gaede:
The MCTruth link for the simulated hits seems to be broken when simulating with two different segmentations and collections.
The example we used is defined here:
https://github.com/iLCSoft/lcgeo/blob/master/ILD/compact/ILD_o4_v01/HCalBarrel_o4_v01_01.xml#L25-L40
This nicely produces two collections with different segmentations from the same shower.

However, when visualizing the Hcal barrel hits - color-coded by particle we get:
![pion_multisegcol](https://cloud.githubusercontent.com/assets/14929044/24540881/547f097a-15f5-11e7-88a1-2d21f5f5a56e.jpg)
whereas from a normal simulation it looks like this:
![pion_ahcal](https://cloud.githubusercontent.com/assets/14929044/24540939/8e2a242a-15f5-11e7-8de1-b6a51e44a840.jpg)

Are we missing something in the configuration (definition of the sensitive Actions) - or is this a bug ?

MarkusFrankATcernch:
Difficult to say if this is a bug or a feature.....
Can this be reproduced using only stuff from DDDetectors - ie without lcgeo?
Then I could debug it....

gaede:
I knew that you were going to say this ;-)
Can you remind me of what the easiest way is to run a simulation in python w/o the ddsim script from lcgeo ?

MarkusFrankATcernch:
Examples are here:
examples/ClientTests/scripts  and
examples/CLICSiD/scripts/CLICSid.py

You should be able to do it simply in python.

gaede:
Could you also try and run our actual code on lxplus ( and also debug it there) ?

- init iLCSoft
- copy a steering file

		mkdir testdir ; cd testdir
		. /cvmfs/clicdp.cern.ch/iLCSoft/builds/nightly/x86_64-slc6-gcc62-opt/init_ilcsoft.sh 
		cp $lcgeo_DIR/example/steeringFile.py  .


- add the following line to the steering file to simulate a single pion

		SIM.gun.particle = "pi+"
		

- run the simulation with a 'normal' ILD simulation:

[...]

PR #143, opened 2017-03-31
Add Geant4EventAction to dump standard DDG4 hits

MarkusFrankATcernch:
BEGINRELEASENOTES
-- Add new test for multi segment multi collections segmentations using a calorimeter endcap.
   See also: https://github.com/AIDASoft/DD4hep/issues/141, which is still unresolved,
   but seems not to be directly related to the Monte-Carlo truth handling.
-- Side effect: add Geant4EventActions to dump hits and particles
-- Fix a linker problem for unicode tags.
ENDRELEASENOTES


MarkusFrankATcernch:
All gitlab platforms finished successfully all tests. Remerge and ignore travis.

Issue #126, opened 2017-03-15
Add quality bit information to SimTrackerHit output 


dhynds:
Correctly set bit in SimTrackerHit (being added) to identify hits where the corresponding MCParticle has not been kept in the output collection.

andresailer:
needs ILCSOFT/LCIO#11

andresailer:
Done

PR #145, opened 2017-04-07
LCIO: Add setting of ProducedBySecondary bit for SimTrackerHits

andresailer:
if the hit is produced by a particle that is not kept in the MCParticle history we set the bit

added oid (original ID) to Geant4Particle class for faster lookup, as this ID is stored with the hit
Otherwise we need a complicated and time-consuming logic to parse the equivalent map if this hit really belonged to that track

Needs ilcsoft/lcio#21 to work, but compiles also against older LCIO versions

BEGINRELEASENOTES
- LCIOOutput: Add setting of ProducedBySecondary bit for SimTrackerHits if the hit is produced by a particle that is not stored in the MCParticle collection, needs lcio 2.8

ENDRELEASENOTES

### DIFF
--- a/DDG4/include/DDG4/Geant4Particle.h
+++ b/DDG4/include/DDG4/Geant4Particle.h
@@ -105,7 +105,9 @@ namespace DD4hep {
       typedef std::set<int> Particles;
       /// Reference counter
       int ref = 0;           //! not persistent
-      int id  = 0, originalG4ID = 0, g4Parent = 0, reason = 0, mask = 0;
+      int id  = 0;
+      int originalG4ID = 0;  //! not persistent
+      int g4Parent = 0, reason = 0, mask = 0;
       int steps  = 0, secondaries = 0, pdgID = 0;
       int status = 0, colorFlow[2] {0,0};
       char charge = 0, _spare[3] {0,0,0};
--- a/DDG4/include/DDG4/Geant4Particle.h
+++ b/DDG4/include/DDG4/Geant4Particle.h
@@ -105,7 +105,7 @@ namespace DD4hep {
       typedef std::set<int> Particles;
       /// Reference counter
       int ref = 0;           //! not persistent
-      int id  = 0, g4Parent = 0, reason = 0, mask = 0;
+      int id  = 0, originalG4ID = 0, g4Parent = 0, reason = 0, mask = 0;
       int steps  = 0, secondaries = 0, pdgID = 0;
       int status = 0, colorFlow[2] {0,0};
       char charge = 0, _spare[3] {0,0,0};
--- a/DDG4/lcio/LCIOConversions.cpp
+++ b/DDG4/lcio/LCIOConversions.cpp
@@ -137,6 +137,17 @@ namespace DD4hep {
         lc_hit->setMCParticle(lc_mcp);
         lc_hit->setPosition(pos);
         lc_hit->setMomentum(hit->momentum.x()/CLHEP::GeV,hit->momentum.y()/CLHEP::GeV,hit->momentum.z()/CLHEP::GeV);
+
+#if LCIO_VERSION_GE( 2, 8 )
+        auto particleIt = pm->particles().find(trackID);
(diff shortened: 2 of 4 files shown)


andresailer:
For #126

petricm:
@gaede @andresailer shouldn't this be in the last tag?

gaede:
I guess no one has ever reviewed/merged this, so it has not been included in any (pre-)release tag ...

andresailer:
I was hoping @MarkusFrankATcernch would have a look as I am changing one of the main classes for DDG4.

MarkusFrankATcernch:
This is the solution to a symptom. The real issue is deeper.
In principle this feature should be extracted in an event action and saved in the Geant4Particle object
as a flag. This lookup in the particle map for the original G4Track-id should be done centrally.
At the time of the lookup it is also clear if the hit was produced by a secondary or not.

What is happening here is, that every output action (LCIO, ROOT,....) effectively must do this lookup explicitly and setting the options. This in the long run will not help, but rather increase confusion.
I suggest we separate this functionality into an event action and then also the output actions will become simpler.

andresailer:
I don't think I understand what you want to store where, and which things to do centrally?
What do you want to store in the `Geant4Particle`? Or do you mean the CalorimeterHit object?

MarkusFrankATcernch:
The hits typically must be patched. This is code similar to this (here ROOT output):

        for(size_t i=0; i<nhits; ++i)   {
          Geant4HitData* h = coll->hit(i);
          Geant4Tracker::Hit* trk_hit = dynamic_cast<Geant4Tracker::Hit*>(h);
          if ( 0 != trk_hit )   {
            Geant4HitData::Contribution& t = trk_hit->truth;
            int trackID = t.trackID;
            t.trackID = m_truth->particleID(trackID);
          }
          Geant4Calorimeter::Hit* cal_hit = dynamic_cast<Geant4Calorimeter::Hit*>(h);
          if ( 0 != cal_hit )   {
            Geant4HitData::Contributions& c = cal_hit->truth;
            for(Geant4HitData::Contributions::iterator j=c.begin(); j!=c.end(); ++j)  {
              Geant4HitData::Contribution& t = *j;
              int trackID = t.trackID;
              t.trackID = m_truth->particleID(trackID);
            }
          }

Take the last 4 lines:
here, if t.trackID != trackID ==> hit from secondary 
I think this is correct.
Now this should be handled centrally.

For LCIO it is similar, except that the lookup is not intrusive as for ROOT, where parallel streams are currently not possible, since the above code may not be executed twice.
Or do I misunderstand the problem?

andresailer:
```
t.trackID != trackID ==> hit from secondary
```
This is what I tried in the beginning. But this would only work if you patch the trackID that belongs to the hits. (is that what your first sentence refers to?)
The hits have the originalTrackID (which I want to add to the Geant4Particle). The originalTrackID is not the same as the trackID, even for particles  that are kept, because the `rebaseParticles` changes the trackID used in the `m_truth->particleID` map.

MarkusFrankATcernch:
OK. I understand now.
Then your patch is fine. But please do not call it oid. Nobody will remember what it really means.
Call it something more descriptive: g4ID, g4id, etc. similar to g4Parent which is the G4 parent's id....

andresailer:
I would use `originalTrackID`?

MarkusFrankATcernch:
originalG4ID ?

andresailer:
Sold!

andresailer:
I already change to `originalG4ID`

MarkusFrankATcernch:
Forgot to add: This should NOT be ROOT persistent (has no meaning afterwards)

Hence declare it like this:
int   originalG4ID         //! not persistent


see "ref" as an example....

andresailer:
Should I change the order of the members? Does it have to be its own statement?

MarkusFrankATcernch:
It needs to be a separate statement. Just split the lines....

andresailer:
add the `not persistent` comment

PR #151, opened 2017-05-05
Use standard cmake way of defining package version

andresailer:
These changes simplify the setting of the package version, so that the version only has to be defined in one place. This allows more automatic tagging of dd4hep, which needs to have the `PACKAGE_VERSION_*` variables be set in the main CMakeLists file.

Also change the variables in DDSegmentation to cmake like PACKAGE_VERSION_* (Will adapt tagging scripts to modify this file automatically as well and change ddsegmentation version via tagging script when making the tag on Monday...)

BEGINRELEASENOTES
- Use cmake to create Version.h file to contain DD4hep version information and macros
- Change the way DD4hep package version is defined and set standard cmake variables for this purpose

ENDRELEASENOTES

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ build/
 # CMake config files
 *Config.cmake
 *ConfigVersion.cmake
+Version.h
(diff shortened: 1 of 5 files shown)


MarkusFrankATcernch:
I believe this is OK like this. The only question, which came to my mind was:
Does DDSegmentation really need it's own versioning? 
In principle we anyhow at some point wanted to merge it into DD4hep.

andresailer:
No ddsegmentation doesn't need its own version. But changing the package structure was too much to do until Monday.

PR #154, opened 2017-05-22
 ignore NANs in Geant4EventReaderGuineaPig

gaede:
BEGINRELEASENOTES
- protect against NANs in Guineapig pairs files in Geant4EventReaderGuineaPig
- make INFO printout more consistent  with DD4hep style

ENDRELEASENOTES


DDG4/plugins/Geant4EventReaderGuineaPig.cpp
```diff
@@ -182,10 +182,24 @@ Geant4EventReaderGuineaPig::readParticles(int /* event_number */,
...
 
+    m_input  >> Energy
+	     >> betaX   >> betaY >> betaZ
+	     >> posX    >> posY  >> posZ ;
     
-    m_input >> Energy
-	    >> betaX   >> betaY >> betaZ
-	    >> posX    >> posY  >> posZ ;
+
+    if( Energy != Energy ||
```
andresailer: maybe better to use isnan: www.cplusplus.com/reference/cmath/isnan/

Issue #131, opened 2017-03-23
Fix IDDecoder so that DetElement is not needed for conversion


petricm:
At present the conversion from cell to position only works if a DetElement is associated to every sensitive.

Example [this](https://github.com/AIDASoft/DD4hep/blob/master/DDRec/src/IDDecoder.cpp#L74) function.

petricm:
Resolved with #155

PR #155, opened 2017-05-31
[WIP] add CellIDPositionConverter

gaede:
Do not merge this PR yet !

For discussion:
 - I implemented the cellID to position conversion based in the discussions at the
   DD4hep mini-workshop w/o using any of the deprecated members in the VolumeManager
     - however the VolumeManagerContext::placement would be useful, as otherwise the Readout has to be searched recursively every time, see:
https://github.com/gaede/DD4hep/blob/cellid_position_converter/DDRec/src/CellIDPositionConverter.cpp#L49-L65
            - any good reason to not keep the placement ?
- now also implemented a first version of the cellID(position) lookup that seems to (mostly work) 

BEGINRELEASENOTES
- add new class DDRec::CellIDPositionConverter
      - replaces DDRec::IDDecoder
      - implement positionNominal(CellID id) and cellID(postion)
     - prepare for using alignment map by separating transforms to DetElement and daughter volume
     - don not use deprecated methods/members in VolumeManager
- add test_cellid_position_converter.cpp
- add VolumeManagerContext::toElement
     - transform from sensitive volume to next DetElement
ENDRELEASENOTES

### DIFF
--- a/DDRec/src/CellIDPositionConverter.cpp
+++ b/DDRec/src/CellIDPositionConverter.cpp
@@ -83,6 +83,7 @@ namespace DD4hep {
 
 
 
+
     CellID CellIDPositionConverter::cellID(const Position& global) const {
 
       CellID result(0) ;
--- a/DDRec/src/CellIDPositionConverter.cpp
+++ b/DDRec/src/CellIDPositionConverter.cpp
@@ -1,0 +1,289 @@
+
+#include "DDRec/CellIDPositionConverter.h"
+#include "DDRec/API/Exceptions.h"
+
+#include "DD4hep/LCDD.h"
+#include "DD4hep/VolumeManager.h"
+
+namespace DD4hep {
+  namespace DDRec {
+
+    using Geometry::DetElement;
+    using Geometry::LCDD;
+    using Geometry::PlacedVolume;
+    using Geometry::Readout;
+    using Geometry::Solid;
+    using Geometry::VolumeManager;
+    using Geometry::Volume;
+    using Geometry::SensitiveDetector;
+    using Geometry::Position;
+    using std::set;
+
+
+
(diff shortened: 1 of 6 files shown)


petricm:
I think on one meeting we concluded that #131 , does this meet this requirement?

gaede:
@petricm yes, this PR fixes #131, i.e. the lookup of the position for a given sensitive cellID works also in setups where there is no DetElement directly assigned to the sensitive volume.

DDRec/src/CellIDPositionConverter.cpp
```diff
@@ -83,6 +83,7 @@ namespace DD4hep {
 
 
 
+
```
petricm: Did you forget to stage? This commit is just a new line.

gaede:
@MarkusFrankATcernch could you please have a look at the implementation of the cellID<->position lookups if this is the way it should work ?
This is without the alignment yet, but the transformations are already split between the DetElement worldTrafo and the sensitive volume to DetElement transformantion, so that we can easily replace the det.nominal() with the corect one, once the AlignementMap will be available...

DDRec/src/CellIDPositionConverter.cpp
gaede: @petricm read the commit message ;-)

petricm: Oh, I understand this is `git commit --amend`

gaede: yes, but you are not supposed to use this after a push ...

andresailer: It is ok to use it as long as your branch is not merged. Those two commits shoud be squashed.
And you can fix the typo in the next commit message

PR #157, opened 2017-06-02
Fix id position conv

gaede:
BEGINRELEASENOTES
-  bug fix in test_cellid_position_converter
     - with this no tests for position from cellID lookup should fail
- re-implement ```CellIDPositionConverter::cellID(pos)```

ENDRELEASENOTES

### DIFF
--- a/DDRec/src/CellIDPositionConverter.cpp
+++ b/DDRec/src/CellIDPositionConverter.cpp
@@ -4,6 +4,8 @@
 #include "DD4hep/LCDD.h"
 #include "DD4hep/objects/VolumeManagerInterna.h"
 
+#include "TGeoManager.h"
+
 namespace DD4hep {
   namespace DDRec {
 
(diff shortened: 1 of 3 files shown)


DDRec/src/CellIDPositionConverter.cpp
```diff
@@ -83,43 +85,107 @@ namespace DD4hep {
...
+    //   // 	PlacedVolume p = geoManager->FindNode( global.x() , global.y() , global.z() ) ;
+
+    //   // 	std::cout << " -> found: " << p.name() << "  valid: " << p.isValid() << " sensitive: "
+    //   // 		  <<  (p.isValid() ? p.volume().isSensitive() : false ) << std::endl ;
+	
+    //   }
+	
+    //   return result ;
+    // }
+
```
petricm: Can you please remove the old code not just comment out

gaede: Actually, I would like to keep the commented out code in, as this is an alternative way of computing the cellID from the position (and we might want to come back to this later ...)

andresailer: Use the history for these kind of things.

PR #160, opened 2017-06-08
Add AlignmentsNominalMap to fake alignment conditions access.

MarkusFrankATcernch:
BEGINRELEASENOTES
Add a new class AlignmentsNominalMap, which behaves like a ConditionsMap and handles alignment entries. 
The AlignmentsNominalMap is not a conditions cache per se. This implementation 
behaves like a conditionsmap, but it shall not return real conditions to the user,
but rather return the default alignment objects (which at the basis are conditions 
as well) to the user. These alignments are taken from the DetElement in question
Alignment DetElement::nominal().

The basic idea is to enable users to write code "as if" there would be conditions
present. This is important to ease in the lifetime of the experiment the step from
the design phase (where obviously no conditions are taken into account) to a more
mature phase, where alignment studies etc. actually are part of the 
"bread and butter work".

Added a corresponding example in examples/AlignDet:
$>   geoPluginRun -volmgr -destroy -plugin DD4hep_AlignmentExample_nominal \
         -input file:${DD4hep_DIR}/examples/AlignDet/compact/Telescope.xml

   Access the DetElement nominal conditions using the AlignmentNominalMap.
   Any use of DDCond is inhibited.
   1) We use the generic printer, which during the detector element scan accesses the
      conditions map.
   2) We use a delta scanner to extract the nominal deltas from the DetElement's 
      nominal alignments
   3) We use a ConditionsTreeMap to perform the alignments re-computation.

===> Coming next on this channel: 
Update the Volume manager to use ConditionsMap objects to resolve position computations.

ENDRELEASENOTES

### DIFF
--- a/DDAlign/src/AlignmentsCalib.cpp
+++ b/DDAlign/src/AlignmentsCalib.cpp
@@ -86,7 +86,7 @@ AlignmentsCalib::_set(DetElement detector, const Delta& delta)   {
     if ( !src_cond.isValid() )   {
       except("AlignmentsCalib",
              "++ The SOURCE alignment condition [%016llX]: %s#%s is invalid.",
-             key.hash, detector.path().c_str(), Keys::deltaName);
+             key.hash, detector.path().c_str(), Keys::deltaName.c_str());
     }
   }
   // Add the entry the usual way. This should also check everything again.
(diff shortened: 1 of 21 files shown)


MarkusFrankATcernch:
All tests on gitlab worked fine....

PR #163, opened 2017-06-13
Add namespace dd4hep

gaede:
BEGINRELEASENOTES
- add header file DD4hep/dd4hep.h
      - define  a convenient namespace dd4hep containing all public namespaces in DDh4ep
            - all DD4hep classes can be called as ```dd4hep::AClass```
      - define a typedef for the LCDD class ```DetDescription```
- use dd4hep.h in test_cellid_position_converter

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/dd4hep.h
+++ b/DDCore/include/DD4hep/dd4hep.h
@@ -1,0 +1,90 @@
+// -*- C++ -*-
+//==========================================================================
+//  AIDA Detector description implementation for LCD
+//--------------------------------------------------------------------------
+// Copyright (C) Organisation europeenne pour la Recherche nucleaire (CERN)
+// All rights reserved.
+//
+// For the licensing terms see $DD4hepINSTALL/LICENSE.
+// For the list of contributors see $DD4hepINSTALL/doc/CREDITS.
+//
+// @author F.Gaede, DESY
+// @date June, 2017
+//==========================================================================
+#ifndef dd4hep_NAMESPACE_H
+#define dd4hep_NAMESPACE_H 1
+
+
+#include "DD4hep/LCDD.h"
+#include "DD4hep/DD4hepUnits.h"
+
+
+
+/// The top level namespace in DD4hep 
+namespace DD4hep{
+
+  /// holds all classes relevant for the geometry description
+  namespace Geometry{}
+
+  /// simulation classes
+  namespace Simulation{}
+  
+  /// dedicated view into the geometry for reconstruction
+  namespace DDRec{}
+  
+  /// alignment module
+  namespace DDAlign{}
+
+  /// alignment classes 
+  namespace Alignments{}
+  
+  /// condition handling
+  namespace Conditions{}
+
+  /// segmentations
+  namespace DDSegmentation{}
+
+  /// utilities for boost::spirit
+  namespace Utils{}
+
+  /// general utilities
+  namespace Utilities{}
+  
+  /// xml stuff
+  namespace XML{}
+
+}
+
+
+namespace DDSurfaces{}
+
+
+
+/** 
+ * Convenient namespace that combines all public namespaces in DD4hep into one global namespace.
+ */
+
+namespace dd4hep {
+  
+  using namespace DD4hep ;
+  using namespace DDSurfaces ;
+  
+  using namespace DD4hep::Geometry ;
+  using namespace DD4hep::DDRec ;
+  using namespace DD4hep::Alignments ;
+  using namespace DD4hep::Conditions ;
+  using namespace DD4hep::DDSegmentation ;
+  using namespace DD4hep::DDAlign ;
+  using namespace DD4hep::Utils ;
+  using namespace DD4hep::Utilities ;
+  using namespace DD4hep::Simulation ;
+  using namespace DD4hep::XML ;
+
+
+  /// define a better type name for the main detector description class
+  typedef LCDD DetDescription ;
+
+}
+
+
+#endif
--- a/UtilityApps/src/test_cellid_position_converter.cpp
+++ b/UtilityApps/src/test_cellid_position_converter.cpp
@@ -12,10 +12,9 @@
 //==========================================================================
 
 // Framework include files
-#include "DD4hep/LCDD.h"
+#include "DD4hep/dd4hep.h"
 #include "DD4hep/DDTest.h"
 
-#include "DD4hep/DD4hepUnits.h"
 #include "DD4hep/BitField64.h"
 #include "DDRec/CellIDPositionConverter.h"
 
@@ -28,11 +27,8 @@
 #include <sstream>
 
 using namespace std ;
-using namespace DD4hep ;
-using namespace DD4hep::Geometry;
-using namespace DD4hep::DDRec ;
-
 using namespace lcio;
+using namespace dd4hep ;
 
 
 static DDTest test( "cellid_position_converter" ) ; 
@@ -72,11 +68,11 @@ int main_wrapper(int argc, char** argv ){
   
   std::string inFile =  argv[1] ;
 
-  LCDD& lcdd = LCDD::getInstance();
+  dd4hep::DetDescription& detDesc = dd4hep::DetDescription::getInstance();
 
-  lcdd.fromCompact( inFile );
+  detDesc.fromCompact( inFile );
 
-  CellIDPositionConverter idposConv( lcdd )  ;
+  CellIDPositionConverter idposConv( detDesc )  ;
 
   
   //---------------------------------------------------------------------
@@ -124,8 +120,8 @@ int main_wrapper(int argc, char** argv ){
 
       std::string cellIDEcoding = col->getParameters().getStringVal("CellIDEncoding") ;
       
-      DD4hep::BitField64 idDecoder0( cellIDEcoding ) ;
-      DD4hep::BitField64 idDecoder1( cellIDEcoding ) ;
+      BitField64 idDecoder0( cellIDEcoding ) ;
+      BitField64 idDecoder1( cellIDEcoding ) ;
 
       int nHit = std::min( col->getNumberOfElements(), maxHit )  ;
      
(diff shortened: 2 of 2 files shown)


petricm:
I don't think we should encourage such usage of namespaces. We should rather stick to the paradigm that **each thing can only be acceded in one way and one way only**. Also this might cause problems down the road with potential clashes of functions. Furthermore you raised the issue on several occasions, that the "excessive" usage of typedefs makes the life of users complicated as you don't know form where things are coming. Aren't you doing the same thing here with this?

gaede:
- we have a proliferation of namespaces in DD4hep that make it really tedious for the users to write (and read) code 
         - most of them are not really needed and they are not in synch with the components (sub-packages) 
- this is addressed with combining all the public (i.e. non-helper namespaces)  into one **short** namespace **dd4hep**
        - this will make it easier to write *and* read code
        - I don't think we have any name clashes across DD4hep namespaces ( and also should not)  
        - we have used the same approach in lcio with the **lcio** namespace and that has worked nicely for many years
- concerning the typedef for LCDD:
         - this is a proposal of how we can repair the unfortunate decision to name the central class in DD4hep  *"Linear Collider Detector Description"* 
         - potentially better would be to rename the class entirely and introduce a typdef ```LCDD``` for backward compatibility

MarkusFrankATcernch:
We should not have namespaces, which only differ in capitalization.
Why not use DD4hep directly ?

namespace DD4hep  {
  using Geoemtry::DetElement;
  using Geometry::xyz;
  .....
}

Would this not also work?

petricm:
* I completely agree with your diagnosis that the number of namespace is to high.
* I don't agree with your cure. If there are to much namespaces we should reduce the number of namespaces, especially if they are as you say not really needed and they are not in sync with the components (sub-packages). Wouldn't you agree that they should rather be in-sync and this is a serious issue?
* My reference to typedefs was a more general comment that the code contains a large number of typedefs and introducing such a super namespace is the same as introducing even more hard to understand typedefs.

gaede:
We already have namespaces **DD4hep** and **dd4hep** for a long time, where **dd4hep** is only used for the units so far.
My personal preference would be to use lower case **dd4hep**, as this is faster to write and does not overlap with the existing capital **DD4hep** one...

andresailer:
I disagree with the notion that there are too many namespaces.
https://www.python.org/dev/peps/pep-0020/#id3
"""Namespaces are one honking great idea -- let's do more of those!"""

I think there are too many cases where namespaces are simply ignored because of the `using` statements, which make it impossible to figure out where things are coming from, well that is impossible in dd4hep now anyway.

Code is `write once, read often`, faster writing does not make better code.

whit2333:
My 2 cents

> I think there are too many cases where namespaces are simply ignored because of the using statements, which make it impossible to figure out where things are coming from, well that is impossible in dd4hep now anyway.

I found this to be a *significant* learning barrier. In most cases the abuse of  `typedef ` in each class header just made things an order of magnitude more difficult to navigate. IMO it is better to leave things explicit in the header and then use `using` in the source. This way I know if the thing is a `std::vector` without having to hunt down the typdef.


Also, the middle finger on my left hand is becoming quite strong from navigating the directories. :)

jlingema:
I also felt like I should chime in: I agree that the problem is not the amount of namespaces but rather the amount of typedefs and using statements. This change would make things even harder to figure out. (Just as an example: Try to figure out from Doxygen that DD4hep::Geometry::XYZ is really the same as the dd4hep::XYZ you saw in some example code.)

With regards to the `LCDD` vs `DetDescription` I agree that it would be nice to have this change but would rather rename the class and add a typedef for backward compatibility but directly deprecate it.

gaede:
Ok, it seems that this rather straight forward fix to the namespace proliferation issue ( yes, this actually is an issue with ~30 (!) namespaces in DD4hep) is more complicated and needs some  more discussion in the next DD4hep developers meeting. 
Just as input to this discussion and in reply to some of the postings:
- almost all of us agree that there are either too many namespaces and/or typedefs in DD4hep that make it hard to find a given or needed class
- as there are so many namespaces most user code has many using declarations in the beginning
       - this then makes it even harder to know whether a class is actually from dd4hep or any other package
- with one global namespace **dd4hep** one can easily avoid any using statements and simply prepend every class with dd4hep::  without too much typing
       - this clearly requires that there are no name clashes inside DD4hep 
       - but sadly there are a few, e.g. DD4hep::Geometry::Segmentation vs. DD4hep::DDSegmentation::Segementation,  DD4hep::Geometry::DetElement vs DD4hep::XML::DetElement
       - could ( and should be fixed ) in a rather straight forward way by renaming classes accordingly
       - this would address the comment from @jlingema about knowing whether DD4hep::Geometry::XYZ is the same as dd4hep::XYZ, as there can only be one class of theat name
       - implementing this properly, however would require some work as one has too find all class name clashes and fix them ...

- if we start to address this issue seriously however, we might use it as an opportunity to clean up the namespaces, API and code organization altogether, e.g.:
         - review the package sub-structure 
                    - e.g. get rid of DDSegmentation/DDSurfaces 
         - organize namespaces according to the package sub-structure
                     - have one public namespace per sub-package/module and optionally an internal one
         - remove all ( as many as possible) extra typedefs
         - cleanup the API  (remove un-needed classes/methods etc
         -   ... and more ...
- any volunteers ?

MarkusFrankATcernch:
After some investigations and some thinking I came to some conclusion.

1) Which namespaces do users actually use ? Here only 2 come to mind:
   DD4hep::XML and DD4hep::Geometry.
   These 2 need attention. DD4hep::Simulation is OK, well separated
   and does not interfere. In fact, hardly anybody is actually directly
   working with these, except simulation managers.
2) I entirely oppose Frank's proposal of adding a new file with "using"
   statement in a new *dd4hep* namespace. Though it seems attractive,
   because of a very low work-involvement, this is nothing but a single
   source of confusion. Even the idea of having 2 namespaces,
   DD4hep and dd4hep, which only differ in capitalization is
   high sub-optimal - to say the least.
3) I investigated and *the* source of nearly all typedefs is the
   Geometry namespace - or better to say it's mis-use. In DDCore actually
   nearly everything is put into the Geometry namespace even if these
   classes are used everywhere and should be promoted.
   I also remember why: to minimize the amount of "using declarations"....
   If all stuff in
   LCDD.h (main entry point),
   Detector.h (DetElement, SensitiveDetector),
   Objects.h (Utility classes like LimitSet, Readout, etc)
   and very few other files is moved up to the DD4hep namespace miraculously
   ~80 % of all typedefs and "using" statements simply vanish.
   This is at least an indication that this is the right direction to go.
   This separation to me also looks logical.
4) We do NOT want to eliminate sub-namespace like Conditions/Alignments.
   We want to keep separated what needs to be separated.
   I also agree with Andre that namespaces are a good thing - if properly
   engeneered. Point 3) indicates that this was not really the case and
   something went out of control.
   Take it as a sporty excercise if you need to invoke the right-mouse button
   to follow some definition in doxygen.
5) Utility namespaces like XML should stay. Maybe the classes therein
   deserve some renaming to math the common typedefs used by everyone.
   Such a rename exercise would be highly intrusive and far from user-friendly.
   There is no transparent way to do such a name change.
6) DDSegmentation is ugly and should be removed. Agreed.
   Eventually ... if there is a volonteer.
   I will not touch it further.

One word concerning the typedefs:
typedefs are an essential part of the interface of a class. Like
virtuality or overloading. In particular templated programming uses typedefs
extensively - just have a look at STL: STL containers are only useful
to STL algorithms, because these can interface to them through the typedefs.
I spare myself the tip to consult text-books....

So, this leaves us with several solutions:

-- LCDD MUST move out of DD4hep::Geometry.
   Whatever else might be applied in addition. One could take the occasion
   to rename the class and keep a "backwards compatible header".
Then in addition:
a) Selectively move stuff from DD4hep::Geometry to DD4hep as indicated
   in 3) and only keep shapes and volumes and associates (VolumeManager)
   in Geometry.
b) Entirely eliminate DD4hep::Geometry and move all these to DD4hep.
   I am not really a friend of such brute force solution, but there
   may be a point, since shapes and volumes are effectively the only
   classes, which are also in use by reconstruction code.
c) Eliminate all DD4hep sub-namespaces, but
   -- keep utilities like XML and
   -- keep utility namespaces like "Interna", "DetectorTool" etc.
d) Moving everything into a flat DD4hep namespace is undesireable
   and cuts down flexibility in the tools.


I think working solutions are only a), b) or c).  Solution d) I veto.

jlingema:
Just felt like I need to respond to your point on typedefs: There is indeed good reason to use typedefs in many circumstances. But just to reduce the amount of typing that needs to be done is not a good example, imo.

As to the solution, I also agree that one flat namespace is not a good idea, as separation by namespaces is, in general, a good thing, as we agreed previously. Therefore, I'd vote for a) or b), where I'd say that b) may be a bit more user-friendly and less prone to cause confusion.

gaede:
@MarkusFrankATcernch  just as a quick response to your thinking and conclusions:
- my proposal to introduce one namespace **dd4hep** was targeted at having cleaner user code and not at resolving the "mess" with nested namespaces  that we currently have in DD4hep
- with this one could immediately tell if a class comes from dd4hep or not, say in reconstruction code, and that's already most of what I want from namespaces
- but as already pointed out this does not work due to name clashes and also you seem to have changed your mind on this, ( the last time we discussed this idea you seemed to have supported it)
- concerning your proposals: I think they all fall short of properly addressing the problem of a confusing user API and doxygen documentation
- why don't we take this as an opportunity to clean up some of the mess and restructure the namespaces  entirely ?
- this could for example be done along the line of the DD4hep components:
        * DDCore
        * DDG4
        * DDRec ( that is actually the only namespace that correctly maps to a component ;-) )
        * DDCond
        * DDAlign
- in these public namespaces we should then only put the *public* classes and *hide* all implementation stuff in the nested sub namespaces
- this would make it a lot easier for users to start browsing the API documentation ...
- lets discuss this tomorrow

gaede:
As we have now reached an agreement on how to proceed with namespaces (see: #166) this PR is obsolete.

PR #168, opened 2017-06-16
Additional environment helper scripts

### LINKED ISSUE #162: Why does thisdd4hep try to set up env for external libraries?
In my opinion, this should not happen.  On my system, it also fails to set up properly unless I remove the external library setup lines, making it even more annoying.

decibelCooper:
This request resolves #162 in the manner suggested in the thread.

BEGINRELEASENOTES

- Added environment helper scripts `thisdd4hep_only.(c)sh` that only set up variables for DD4hep and not for dependencies.

ENDRELEASENOTES


andresailer:
You also have to `configure` the scripts like here:
https://github.com/AIDASoft/DD4hep/blob/master/CMakeLists.txt#L151
And install them like a few lines below that line

decibelCooper:
@andresailer My apologies for the haste.  This has been updated and tested.

andresailer:
No worries, thanks for your contribution

cmake/thisdd4hep_only.sh
```diff
@@ -0,0 +1,91 @@
...
+# Author: Pere Mato. F. Gaede, M.Frank
+#-------------------------------------------------------------------------------
+#
+#echo " ### thisdd4hep.sh:   initialize the environment for DD4hep ! " 
+#
+#-----------------------------------------------------------------------------
+dd4hep_parse_this()   {
+    package=${2};
+    if [ "x${1}" = "x" ]; then
+	if [ ! -f bin/this${package}.sh ]; then
```
andresailer: This should be `this${package}_only.sh`

cmake/thisdd4hep_only.sh
```diff
@@ -0,0 +1,91 @@
...
+            echo ERROR: must "cd where/${package}/is" before calling ". bin/this${package}.sh" for this version of bash!;
```
andresailer: This should be `this${package}_only.sh`

PR #171, opened 2017-06-18
Changing the default install back to the (good) default.

whit2333:
Let's just leave the cmake defaults so the build behaves in a predictable way.

I'll never understand in-source installs... They should never be the default.

BEGINRELEASENOTES

- Changed the default CMAKE_INSTALL_PREFEX back to the standard default value instead of the source directory.

ENDRELEASENOTES


andresailer:
Let's just leave the defaults to what they have been for years, so the install behaves in a predictable way.
I'll never understand /usr installations, they should never be the default for packages that constantly change, so users have to hunt down header files that the compiler will find before the new ones.

whit2333:
For those people working in containers good defaults make a lot of sense. When almost every build in the word defaults to `CMAKE_INSTALL_PREFIX=/usr/local` it makes no sense to do anything else. I agree that the system installs don't make sense for unstable projects but at some point (hopefully) this argument will no longer be valid. So why not change it now?

Developers will have no problem manually changing this argument. (Actually it is probably a good idea for testing the various build options.)

petricm:
@whit2333 we had this discussion already once, and agreed that if you want to introduce different install behaviour, we are open to the possibility of introducing a flag called `SYSTEM_INSTALL` that changes the install behaviour. However the prerequisite that the developers agree to this is that the **current behavour does not change**.

PR #174, opened 2017-06-19
fix bwd compatibility for DDRec/DetectorData.h

gaede:
BEGINRELEASENOTES
- fix for the backward compatibility of classes in DDRec/DetectorData.h wrt. namespace change

ENDRELEASENOTES


gaede:
this is obsolete now.

PR #176, opened 2017-06-19
Examples in share directory

whit2333:
This a replacement for https://github.com/AIDASoft/DD4hep/pull/172. 

In an attempt to isolate the source/build/install the directory `CMAKE_INSTALL_PREFIX/share/DD4hep` is used to install all the example data.

The directory structure in this directory is
```
└── DD4hep
    ├── DDDetectors
    │   └── compact
    │       └── SiD
    └── examples
        ├── AlignDet
        │   ├── compact
        │   ├── drivers
        │   └── src
        ├── bin
        ├── CLICSiD
        │   ├── compact
        │   ├── eve
        │   ├── scripts
        │   └── sim
        ├── ClientTests
        │   ├── compact
        │   ├── scripts
        │   ├── src
        │   └── src_boost
        ├── Conditions
        │   ├── data
        │   ├── src
        │   └── xml
        ├── DDDB
        │   ├── data
        │   └── scripts
        ├── DDEve
        ├── DDG4
        │   ├── data
        │   └── examples
        ├── lib
        ├── Segmentation
        └── SimpleDetector
            ├── compact
            ├── examples
            └── src
```

There is a bit of overlap since the source directories `DDG4/examples` and `examples/DDG4` are both installing into the `share/DD4hep/examples/DDG4` directory. It seems these examples should be reorganized or removed.

BEGINRELEASENOTES
- All examples are installed to `CMAKE_INSTALL_PREFIX/share/DD4hep`

ENDRELEASENOTES


petricm:
This is again the same comment as one of the previous PRs, this can go under a `SYSTEM_INSTALL` option

whit2333:
Why are you opposed to it?

petricm:
- Because you are now installing a bunch of pointless stuff to `share` like `.cpp` files which is not needed in the installation destination
 - Because we don't want to mix different kinds of examples
 - Because there might be some hidden dependencies to files you moved

whit2333:
> Because you are now installing a bunch of pointless stuff to share like .cpp files which is not needed in the installation destination

Those are the examples. If the examples are pointless then that is another issue.

>  Because we don't want to mix different kinds of examples

This is a separate issue. The examples just need to be organized better

>  Because there might be some hidden dependencies to files you moved

I am not sure what you mean by this. "Hidden" dependencies just smells bad.

petricm:
The developers have decided on the last meeting to retain the current directory structure.

PR #199, opened 2017-07-04
set flag=true in VolumeManager_Populator::add_entry

gaede:
BEGINRELEASENOTES
- bug fix for VolumeManagerContext::toElement() and VolumeManagerContext::placement() 
      - set flag=true in VolumeManager_Populator::add_entry when 
         a ContextExtension is needed, i.e. sensitive volume is not DetElement's volume
- fixes problems in CellIDPositionConverter

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/VolumeManager.cpp
+++ b/DDCore/src/VolumeManager.cpp
@@ -268,7 +268,7 @@ namespace dd4hep {
             context->element    = e;
             if ( nodes.size() > 0 )  {
               ContextExtension* ext = new(_getExtension(context)) ContextExtension();
-	      context->flag=true ;
+              context->flag=1;
               ext->placement  = PlacedVolume(n);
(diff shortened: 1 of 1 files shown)


DDCore/src/VolumeManager.cpp
```diff
@@ -268,6 +268,7 @@ namespace dd4hep {
             context->element    = e;
             if ( nodes.size() > 0 )  {
               ContextExtension* ext = new(_getExtension(context)) ContextExtension();
+	      context->flag=true ;
```
andresailer: For consistency, please use spaces for indentation, and drop the space before the semicolon.

MarkusFrankATcernch:
context->flag = 1;
would probably be more flexible for future extensions, which could use the other bits.

andresailer:
This breaks 2 of the tests

PR #201, opened 2017-07-06
DDCore: Changes to the VolumeManager interface

MarkusFrankATcernch:
BEGINRELEASENOTES
  ## DDCore: Changes to the VolumeManager interface

  Recent descrepancies showed that the call to lookup a placement
  from the volume manager may have an ambiguous meaning:
  It may (as used until now) be the placement of thge closest
  detector element - a functionality used by various tests
  or be the placement of the sensitive volume itself.
  So far, since each sensitive volume in the DD4hep tests
  is represented by a DetElement structure, both
  approaches returned the same placed volume.

  Since there is the possibility to have sensitive volumes, which are not
  directly connected to a single DetElement structure, this call was
  split to resolve this ambiguity:

  /// Lookup a physical (placed) volume identified by its 64 bit hit ID
  PlacedVolume lookupVolumePlacement(VolumeID volume_id) const;
  /// Lookup a physical (placed) volume of the detector element
  /// containing a volume identified by its 64 bit hit ID
  PlacedVolume lookupDetElementPlacement(VolumeID volume_id) const;

ENDRELEASENOTES


MarkusFrankATcernch:
If tomorrow morning nobody has objections, I shall merge.
Travis and gitlab finished successfully.

gaede:
Thanks Markus. Why not merge right away ?

MarkusFrankATcernch:
OK. I'll merge.

PR #202, opened 2017-07-13
Implement ROOT persistency mechanism for detector descriptions

MarkusFrankATcernch:
BEGINRELEASENOTES

## Implement ROOT persistency mechanism for detector descriptions

The following commits allow to save and restore Detector objects (ie. the full detector description) to/from ROOT. Most changes affected the usage of void pointers. 

It is **NOT** possible to:
 * save object extensions. Another usage of void pointers and type-info objects, which both cannot be saved.
 * save callback objects. Callbacks store in memory pointers to member functions. These depend on the loading of libraries at run-time and hence may differ from application to application.
 * save conditions slices. This shall be a futute extension. There should be no fundamental problem doing so - it was simply not yet done.

In the example area a new slot call "Persistency" was created with various tests:
 * Save and restore simple conditions
 * Save and restore the MiniTel detector (`examples/ClientTests/compact/MiniTel.xml`)
 * Save and restore the CLICSiD detector (`examples/CLICSiD/compact/compact.xml`)

The detector examples also have a set of tests associated to check the validity of the restored information.

## Integrate `DDSegmentatation` into `DDCore`

As a first step to start the cleanup of `DDSegmentation` the code was integrated into `DDCore`. For backwards compatibility a dummy library `libDDSegmentation.so` is kept, so that client cmake files directly referring `DDSegmentation` still work. **This tough is only a temporary measure and users should remove references to the DDSegmentation library.** Include files can be accessed as before. These are part of the `DDCore` include directory.

ENDRELEASENOTES


petricm:
The tests have failed due to a race condition, since on gitlab we do `ctest` and on travis `ctest -j2`

andresailer:
One can add a `depends` property to the tests https://cmake.org/cmake/help/v3.3/prop_test/DEPENDS.html to ensure the order of execution

PR #204, opened 2017-07-14
MegatileLayerGridXY segmentation: add segmentation parameters

danieljeans:
BEGINRELEASENOTES
- add ncellsX/Y as a "parameter", allowing it to be set in compact description. This change is for easier use in the case of a MultiSegmentation. (Only a uniform segmentation can be defined in this way: for more complex cases, must set by driver.) 
- change from array to std::vector to store ncells information
ENDRELEASENOTES


danieljeans:
hmm seems to be some problem
I'll look into it

petricm:
@gaede if you review and merge this by 14:00 it will be in the next build so I can update the nightly at 14:45.

PR #205, opened 2017-07-14
Re-enable commented out code

MarkusFrankATcernch:
BEGINRELEASENOTES
 * Re-enable commented out code in `segmentations/Segmentation.cpp`

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/segmentations/Segmentation.cpp
+++ b/DDCore/src/segmentations/Segmentation.cpp
@@ -38,14 +38,12 @@ namespace dd4hep {
 
     /// Destructor
     Segmentation::~Segmentation() {
-#if 0
       if (_ownsDecoder and _decoder != 0) {
         delete _decoder;
       }
       for (auto& p : _parameters)  {
         if ( p.second ) delete p.second;
       }
-#endif
       _parameters.clear();
     }
   


gaede:
... and what does this re-enabled code do ?

andresailer:
Delete the parameters for segmentation and fix memory leaks

gaede:
.. then this should maybe added to the release notes - I could see that from the code diff myself ;-)

andresailer:
Actually as this is just a correction to the previous PR that will end up the same release there shouldn't be any release notes

MarkusFrankATcernch:
This was a temporary leftover from the persistency exercise. I had problems with segmentations and had to find a work-around. 
BTW a major source of problems were also the BitField64 structures. These also need some re-thinking, since in addition they are a prime example of anti-multi-threaded code.

gaede:
This is what get's added to the Release notes by the tagger script:
```
* 2017-07-14 Markus Frank ([PR#205](https://github.com/AIDASoft/DD4hep/pull/205))
  * Re-enable commented out code in `segmentations/Segmentation.cpp`
```

no big issue, but in general one should try and write reasonable PR comments ...

@MarkusFrankATcernch can you be a bit more specific about the BitField64 problem - or even better, create an issue ...

petricm:
@gaede until you do a tag you can edit PR message and the BEGINRELEASENOTES tag and it will be picked up by the tagger script, regardless if it is already merged.

gaede:
I know that Marko - and thought everyone else did as well - this is why I had written the first comment...

MarkusFrankATcernch:
Folks, I am not against discussions, but this is getting absurd.
Do you really have so much energy left to waste it for such an non-issue?

PR #216, opened 2017-08-14
Beampipe modifications

Voutsi:
BEGINRELEASENOTES
  * Adding variable "nocore" for beam pipe (with default value = false) In case the variable appears in the BP xml file as "true", the BP sections will have no core of beam material, in order that someone might add various BP walls made of different materials while avoiding G4 overlaps. Example of use:
```xml
<detector name="BeBeampipe" type="DD4hep_Beampipe_o1_v01" insideTrackingVolume="true" nocore="true" vis="BeamPipeVis">
```

ENDRELEASENOTES

### DIFF
--- a/DDDetectors/src/Beampipe_o1_v01_geo.cpp
+++ b/DDDetectors/src/Beampipe_o1_v01_geo.cpp
@@ -111,7 +111,7 @@ static dd4hep::Ref_t create_element(dd4hep::Detector& description,
   //Access to the XML File
   xml_det_t xmlBeampipe = element;
   const std::string name = xmlBeampipe.nameStr();
-
+  bool nocore  = xmlBeampipe.nocore(false);
 
   dd4hep::DetElement tube(  name, xmlBeampipe.id()  ) ;
 
@@ -210,10 +210,6 @@ static dd4hep::Ref_t create_element(dd4hep::Detector& description,
       Volume tubeLog( volName, tubeSolid, coreMaterial ) ;
       Volume tubeLog2( volName, tubeSolid, coreMaterial ) ;
 
-      // placement of the tube in the world, both at +z and -z
-      envelope.placeVolume( tubeLog,  transformer );
-      envelope.placeVolume( tubeLog2,  transmirror );
-
       // if inner and outer radii are equal, then omit the tube wall
       if (rInnerStart != rOuterStart || rInnerEnd != rOuterEnd) {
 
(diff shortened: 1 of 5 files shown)


DDDetectors/src/Beampipe_o1_v01_geo.cpp
```diff
@@ -266,9 +262,25 @@ static dd4hep::Ref_t create_element(dd4hep::Detector& description,
...
-	tubeLog2.placeVolume( wallLog2,  Transform3D() );
+	if (nocore){
+
+	  //placement of the wall only
+	  envelope.placeVolume( wallLog,  transformer );
+	  envelope.placeVolume( wallLog2,  transmirror );
+
+	}
+
+	else{
```
andresailer: Please put the else on the line with the closing bracket

andresailer:
Can you post or link to an example of how the beampipe looks with `nocore=true`? Thanks

DDDetectors/src/Beampipe_o1_v01_geo.cpp
```diff
@@ -111,7 +111,7 @@ static dd4hep::Ref_t create_element(dd4hep::Detector& description,
   //Access to the XML File
   xml_det_t xmlBeampipe = element;
   const std::string name = xmlBeampipe.nameStr();
-
+  bool nocore  = xmlBeampipe.nocore(false);
```
Voutsi: Yes, the idea is to put several BP detectors, the inner one (e.g. coating) with nocore=false and then several ones around it w/o core, like Be, water cooling tube, shields etc... I think that would be easier if I put it as a detector attribute, at least for the FCCee case. Your idea is to add an additional section attribute or to replace the detector attribute with a section one?

Voutsi:
You can find an example of an xml file with nocore = true at:
/afs/cern.ch/user/v/voutsina/public/Beampipe_o4_v02_test_new.xml

DDDetectors/src/Beampipe_o1_v01_geo.cpp
andresailer: Ok, I see. I hadn't thought about placing two beampipe detectors. I think then it is OK for me.

andresailer:
Can you now just squash your commits, please? At least getting rid of the merge commits, and the whitespace only commits.

andresailer:
Ok, thanks for cleaning up the commits. (The commit about "rootma[" will be removed by rebasing).
Is there anything else you want to add? Otherwise I will merge.

Voutsi:
Thanks, no I don't have anything else concerning this request

Issue #226, opened 2017-09-06
LCIOEventReader discards daughters of documentation particles


decibelCooper:
At https://github.com/AIDASoft/DD4hep/blob/b0793072867d0321603375afdf3448d749c3c230/DDG4/lcio/LCIOEventReader.cpp#L155, the LCIOEventReader discards daughters of documentation particles.  This is clearly based on the assumption that documentation particles cannot or should not have final state descendants.  However, this causes a problem with generators such as Pythia8.

I believe that this `else` line should be removed.  Its removal has functioned properly in my testing.

gaede:
This code indeed needs reviewing. We have seen similar problems with the latest version of Whizard and lcio files. However, I don't think that simply removing the else here is really correct in all cases and for all generators, as this would simply add all particles without parents (including documentation particles, e.g. the incoming particles) to the event...

decibelCooper:
I disagree that this "would simply add all particles without parents to the event."  By my understanding, the particles actually thrown in Geant4 are determined by https://github.com/AIDASoft/DD4hep/blob/987f752486c026985e25c9a0785952bfe6546021/DDG4/src/Geant4InputHandling.cpp#L358-L414.  Adding particles in `vtx->out` causes relevant descendants to be gathered from them at https://github.com/AIDASoft/DD4hep/blob/987f752486c026985e25c9a0785952bfe6546021/DDG4/src/Geant4InputHandling.cpp#L452.  Only then are the particles added to a G4PrimaryVertex at https://github.com/AIDASoft/DD4hep/blob/987f752486c026985e25c9a0785952bfe6546021/DDG4/src/Geant4InputHandling.cpp#L460.  I fail to see the drawback to adding primary documentation particles, as they are not thrown.

andresailer:
@decibelCooper : Could you link to the input file you used for your tests, please?

decibelCooper:
Hi @andresailer 

Here is a simple example that I have used in testing.  It is a simple single-particle generator with documentation parents...

https://drive.google.com/open?id=0B5xHMP_V-NlFSEpydTl5bTBCN0E

Also, here is a Pythia8 DIS file that exhibits the problem (also used in testing)...

https://drive.google.com/open?id=0B5xHMP_V-NlFcnFfbHlKbTI0djQ

andresailer:
Assigning @gaede per his comment in #229

PR #229, opened 2017-09-14
[WIP] Support genstatus=4 and 5 of MCParticles in lcio and stdhep inputs by Whizard2 

akiyamiyamoto:
BEGINRELEASENOTES
- New Whizard2 will be able to set generator particle status of 4 for beam particles and 5 for hard process partons.  Changes are to support these status code properly.
ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ option(CMAKE_MACOSX_RPATH "Build with rpath on macos" ON)
 # project version
 SET( DD4hep_VERSION_MAJOR 1  )
 SET( DD4hep_VERSION_MINOR 1 )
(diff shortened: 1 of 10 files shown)


petricm:
Please rebase your PR

andresailer:
I don't think this approach will scale. Pythia8 has 200 status codes http://home.thep.lu.se/~torbjorn/pythia81html/ParticleProperties.html. As there are use cases for using different generators #226 we need a different solution that works for other generators as well.

gaede:
I tend to agree with Andre, there is simply too many different conventions for the status codes other than 1 (stable) used by the different generators (and versions of the same generators).
I will make a PR that should address both the Pythia8  issue (#226) and the new Whizard2 lcio files...

PR #238, opened 2017-09-28
Add thread safe BitFieldCoder

### LINKED ISSUE #187: Review DDSegmentation API
Based on iLCSoft/lcgeo#113 where it is stated:

_Markus has changed the signature of Segmentation::decoder() from BitField64* decoder() const; to const BitField64* decoder() const; this is why this extra const_cast is needed now.
The other ones where there before.
Eventually, we should review both - might be a problem in the API ..._

gaede:
Resolves #187 
BEGINRELEASENOTES
- add new threadsafe class `BitFieldCoder` as replacement for `BitField64`
- use as `const` everywhere
- re-implement `BitField64` using `BitFieldCoder`
  - is thread safe if used locally 
  - can be instantiated from `const BitFieldCoder*`

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/BitField64.h
+++ b/DDCore/include/DD4hep/BitField64.h
@@ -23,7 +23,7 @@ namespace dd4hep {
   namespace detail {
 
     using DDSegmentation::BitField64;
-    using DDSegmentation::BitFieldValue;
+    using DDSegmentation::BitFieldElement;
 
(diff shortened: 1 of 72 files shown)


gaede:
@MarkusFrankATcernch should we now remove all the `//! not ROOT-persistent` statements for the encoder and related classes ?

PR #246, opened 2017-10-13
Improve DDCMS. Please see release notes and comments.

petricm:
Replaces #245 
BEGINRELEASENOTES
### DDCMS:
- Improve the CMS excercise. New examples etc.
- Support for simulation using DDG4 (at least partially - since not all subdetector volumes are accepted by Geant4).

### DDG4:
- Event reader returns `EVENT_READER_EOF` if `EOF` is detected rather than a generic IO error.
- Add generator status word to the `Geant4Particle` object. Remove the extension mechanism, which is very heavy to just add one integer.

### General:
 - We need to distinguish the plugins using some namespace mechanism. I started to introduce the namespace separator `"_".` Hence all DD4hep plugins start with `DD4hep_<plugin>`. I hope this does not break everything. If it does, please notify me and we can undo.

ENDRELEASENOTES


petricm:
Please do not merge, a HEAD integration build is running just now, I'll merge once it's over.

PR #248, opened 2017-10-17
Fix bug in VolumeManager returning bad top level subdetectors

MarkusFrankATcernch:
BEGINRELEASENOTES
# VolumeManager Implementation
A possibly important bug was fixed for the lookup of top level subdetectors in the VolumeManager by volume identifers of (sensitive) volumes. Due to a bug in the de-masking possible wrong top level subdetectors were returned. The default use cases typically do not use this call and hence should not be affected.

ENDRELEASENOTES


petricm:
Based on a discussion in the developers meeting, this bug has limited implications, since the function (`lookupDetector(VolumeID volume_id)`)  that has the bug is not called anywhere in DD4hep apart from one test.

PR #297, opened 2018-01-30
Geant4Output2LCIO: Run header order fix

rete:
BEGINRELEASENOTES
- Geant4Output2LCIO
   - Switch run header writing order at beginning of run instead of end of run

ENDRELEASENOTES


rete:
We noticed recently that the run header was written after all events at the end of the simulation

PR #301, opened 2018-02-05
Geant4Output2LCIO extension

rete:
BEGINRELEASENOTES
- Geant4Output2LCIO plugin
   - Added run number and event number offset properties. Enable steering of run and event counters from outside
   - Added 3 event parameters properties for int, float and string event parameters 

ENDRELEASENOTES


petricm:
rebase please

PR #317, opened 2018-02-20
[WIP] DDCore: added CartesianStrip segmentations

### LINKED ISSUE #303: Strip detector with multiple layers that constrain different directions
I cannot find built-in functionality to specify a strip detector with multiple sensitive layers with strips in different directions.  Seems like there are two reasonable ways to accomplish this.  Either 1) allow the specification of a readout coordinate transformation within the Volume, or 2) make custom detector constructor that rotates sensitive layers relative to each other while keeping required boundary after rotation.  Option 1 would allow much more flexibility.

IMO there is also an issue with the lack of one-dimensional segmentation.  Sure, 2-D segmentation could be used, but this then reports a strip length that is incorrect in general.  Take the SiTrackerEndcap2 for example, where the sensitive volumes are trapezoids.  In the case of the trapezoidal volume with strip readout, the reconstruction needs to probe the distance to volume boundary in order to calculate, e.g., a noise covariance matrix for the strip measurements.  I think the solution to this is to have Segmentations that only report 1 dimension, indicating to reconstruction that the other dimensions must be determined in another way.

I will be working on a PR to resolve this unless I get feedback that indicates that I should do otherwise.  Thanks in advance for any feedback!

decibelcooper:
This PR resolves #303.

The idea is to extend the collection of segmentations to include 1D segmentations in addition to the existing 2D and 3D segmentations.  All code was adapted from 2D segmentations.  This PR is a WIP because testing should be added.

Devs, please let me know what you think.  If this is something you want in the code, I can add tests.  I need something like this, so if not, I'll put it into my own code.

Cheers!

BEGINRELEASENOTES
- DDCore: added CartesianStrip segmentations

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/CartesianStripX.h
+++ b/DDCore/include/DD4hep/CartesianStripX.h
@@ -1,0 +1,95 @@
+//==========================================================================
+//  AIDA Detector description implementation
+//--------------------------------------------------------------------------
+// Copyright (C) Organisation europeenne pour la Recherche nucleaire (CERN)
+// All rights reserved.
+//
+// For the licensing terms see $DD4hepINSTALL/LICENSE.
+// For the list of contributors see $DD4hepINSTALL/doc/CREDITS.
+//
+//  \author   Markus Frank, David Blyth
+//  \date     2018-02-20
+//  \version  1.0
+//
+//==========================================================================
+#ifndef DD4HEP_DDCORE_CARTESIANSTRIPX_H
+#define DD4HEP_DDCORE_CARTESIANSTRIPX_H 1
+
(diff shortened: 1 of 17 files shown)


petricm:
Dear David,
thank you for your contribution. We are happy to include this, but as you said a few tests would be appreciated. And you can also drop the if:
```cpp
#if __cplusplus >= 201103L
    return {_stripSizeX};
#else
    std::vector<double> cellDims(1, 0.0);
    cellDims[0] = _stripSizeX;
    return cellDims;
#endif
```
since C++98 support wad dropped some time ago.

PR #319, opened 2018-02-22
Remove unnecessary import exceptions

epilicer:
BEGINRELEASENOTES
- Remove unnecessary `import exceptions` from python

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDG4.py
+++ b/DDG4/python/DDG4.py
@@ -121,7 +121,6 @@ def _registerGlobalFilter(self,filter):
   self.get().registerGlobalFilter(Interface.toAction(filter))
 #---------------------------------------------------------------------------
 def _getKernelProperty(self, name):
-  import exceptions
   #logging.info('_getKernelProperty: %s  %s',str(type(self)),name)
   ret = Interface.getPropertyKernel(self.get(),name)
   if ret.status > 0:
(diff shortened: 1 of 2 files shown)


DDG4/python/DDG4.py
```diff
@@ -264,16 +272,24 @@ def _get(self, name):
   #  return getattr(self,name)
   #traceback.print_stack()
   msg = 'Geant4Action::GetProperty [Unhandled]: Cannot access property '+a.name()+'.'+name
-  raise exceptions.KeyError(msg)
+  if sys.version_info[0] < 3:
+    import exceptions
```
andresailer: This check and import is not necessary for python2:
https://docs.python.org/2/library/exceptions.html
>>> This module never needs to be imported explicitly: the exceptions are provided in the built-in namespace as well as the exceptions module.
Just dropping `import exceptions` and `exceptions.` should be fine.

DDG4/python/DDG4.py
```diff
@@ -121,7 +121,7 @@ def _registerGlobalFilter(self,filter):
   self.get().registerGlobalFilter(Interface.toAction(filter))
 #---------------------------------------------------------------------------
 def _getKernelProperty(self, name):
-  import exceptions
+  import sys
```
andresailer: `sys` is not used

DDG4/python/DDG4.py
```diff
@@ -131,16 +131,16 @@ def _getKernelProperty(self, name):
   elif hasattr(self,name):
     return getattr(self,name)
   msg = 'Geant4Kernel::GetProperty [Unhandled]: Cannot access Kernel.'+name
-  raise exceptions.KeyError(msg)
+  raise KeyError(msg)
 
 #---------------------------------------------------------------------------
 def _setKernelProperty(self, name, value):
-  import exceptions
+  import sys
```
andresailer: `sys` is not used

DDG4/python/DDG4.py
```diff
@@ -250,7 +250,7 @@ def _adopt(self,action):  self.__adopt(action.get(),action.callback())
 
 #---------------------------------------------------------------------------
 def _get(self, name):
-  import exceptions, traceback
+  import sys, traceback
```
andresailer: `sys` is not used

DDG4/python/DDG4.py
```diff
@@ -264,16 +264,16 @@ def _get(self, name):
   #  return getattr(self,name)
   #traceback.print_stack()
   msg = 'Geant4Action::GetProperty [Unhandled]: Cannot access property '+a.name()+'.'+name
-  raise exceptions.KeyError(msg)
+  raise KeyError(msg)
 
 def _set(self, name, value):
-  import exceptions
+  import sys
```
andresailer: `sys` is not used

PR #321, opened 2018-02-27
Passing LCIO input parameters to output 

### LINKED ISSUE #309: DDG4 : pass LCEvent input parameters to LCEvent output
Hi all,

I'm looking for an (elegant) way to pass data from the LCIO event reader : 
https://github.com/AIDASoft/DD4hep/blob/master/DDG4/lcio/LCIOEventReader.h
to the LCIO event writer : 
https://github.com/AIDASoft/DD4hep/blob/master/DDG4/lcio/Geant4Output2LCIO.cpp
We would like to pass the following input LCEvent parameters : 
- event number
- run number
- event parameter maps

to the output LCEvent.

I've started to look a bit in the available DD4hep classes and I've found that maybe we can use the `Geant4Action::context()` method in the reader or in the writer to make a communication between the two, as it is used here to get the geometry description : 
- Example : https://github.com/AIDASoft/DD4hep/blob/master/DDG4/lcio/Geant4Output2LCIO.cpp#L229
- G4 Context class : https://github.com/AIDASoft/DD4hep/blob/master/DDG4/include/DDG4/Geant4Context.h#L237

I'm not sure what to look for then.
Pinging @gaede for eventual additional information.

rete:
Resolves #309 
BEGINRELEASENOTES
- Geant4InputAction and Geant4EventReader : 
    - Register input action pointer to event reader, enabling the event reader to access the Geant4 context
- Added new class LCIOEventParameters to handle LCIO input event parameters
- LCIOFileReader : 
    - Add extension with LCIO input event parameters to Geant4event
- Geant4Output2LCIO : 
    - Get LCIO event parameters from event extension (if any) and write them to LCIO output event

ENDRELEASENOTES

### DIFF
--- a/DDG4/src/Geant4InputAction.cpp
+++ b/DDG4/src/Geant4InputAction.cpp
@@ -40,7 +40,7 @@ Geant4EventReader::~Geant4EventReader()   {
 /// Get the context (from the input action)
 Geant4Context* Geant4EventReader::context() const {
   if( 0 == m_inputAction ) {
-    printout(FATAL,"Geant4EventReader: %s", "No input action registered!");
+    printout(FATAL,"Geant4EventReader", "No input action registered!");
     throw std::runtime_error("Geant4EventReader: No input action registered!");
   }
   return m_inputAction->context();
--- a/DDG4/src/Geant4InputAction.cpp
+++ b/DDG4/src/Geant4InputAction.cpp
@@ -29,12 +29,26 @@ typedef Geant4InputAction::Vertices Vertices ;
 
 /// Initializing constructor
 Geant4EventReader::Geant4EventReader(const std::string& nam)
-  : m_name(nam), m_directAccess(false), m_currEvent(0)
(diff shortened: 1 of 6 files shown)


DDG4/src/Geant4InputAction.cpp
```diff
@@ -29,14 +29,28 @@ typedef Geant4InputAction::Vertices Vertices ;
...
 }
 
 /// Default destructor
 Geant4EventReader::~Geant4EventReader()   {
 }
 
+/// Get the context (from the input action)
+Geant4Context* Geant4EventReader::context() const {
+  if( 0 == m_inputAction ) {
+    printout(FATAL,"Geant4EventReader: %s", "No input action registered!");
```
andresailer: the "%s" has to be removed, second argument is the "name for the printout
From the failed test (https://travis-ci.org/AIDASoft/DD4hep/jobs/346807978#L1623)
```
Geant4EventReader: ��҂ FATAL No input action registered!
```

rete:
Can be merged on my side, if approved.

PR #322, opened 2018-02-28
Fix multipole fields

danieljeans:
I believe there are bugs in the definition of the multipole fields (both in the code and the documentation).
I've compared to the definitions given in
https://dd4hep.web.cern.ch/dd4hep/reference/classdd4hep_1_1MultipoleField.html

BEGINRELEASENOTES
- correct coded expressions for quadrupole, sextupole and octopole fields.
- correct expression for octopole field in documentation
ENDRELEASENOTES

### DIFF
--- a/DDCore/src/FieldTypes.cpp
+++ b/DDCore/src/FieldTypes.cpp
@@ -99,12 +99,12 @@ void MultipoleField::fieldComponents(const double* pos, double* field) {
     double y2 = y*y;
     switch(coefficents.size())  {
     case 4:      // Ocupole momentum
-      by +=  coefficents[3] * (x2*x - 3.0*x*y2) + skews[3]*(y2*y - 3.0*x2*y);
-      bx +=  coefficents[3] * (3.0*x2*y - y2*y) + skews[3]*(x2*x - 3.0*x*y2);
+      by += (1./6.) * ( coefficents[3] * (x2*x - 3.0*x*y2) + skews[3]*(y2*y - 3.0*x2*y) );
+      bx += (1./6.) * ( coefficents[3] * (3.0*x2*y - y2*y) + skews[3]*(x2*x - 3.0*x*y2) );
       [[fallthrough]];
     case 3:      // Sextupole momentum:
-      by +=  coefficents[2] * (x2 - y2) - skews[2] * 2.0 * xy;
-      bx +=  coefficents[2] * 2.0 * xy + skews[2] * (x2 - y2);
+      by +=  (1./2.) * ( coefficents[2] * (x2 - y2) - skews[2] * 2.0 * xy );
+      bx +=  (1./2.) * ( coefficents[2] * 2.0 * xy + skews[2] * (x2 - y2) );
       [[fallthrough]];
     case 2:      // Quadrupole momentum:
       bx += coefficents[1] * y + skews[1]*x;
--- a/DDCore/src/FieldTypes.cpp
+++ b/DDCore/src/FieldTypes.cpp
@@ -99,16 +99,16 @@ void MultipoleField::fieldComponents(const double* pos, double* field) {
     double y2 = y*y;
     switch(coefficents.size())  {
     case 4:      // Ocupole momentum
-      by +=  coefficents[3] * (x2*x - 3.0*x*y2) + skews[3]*(y2*y - 3.0*x*y2);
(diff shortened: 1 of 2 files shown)


DDCore/src/FieldTypes.cpp
```diff
@@ -99,16 +99,16 @@ void MultipoleField::fieldComponents(const double* pos, double* field) {
...
     switch(coefficents.size())  {
     case 4:      // Ocupole momentum
-      by +=  coefficents[3] * (x2*x - 3.0*x*y2) + skews[3]*(y2*y - 3.0*x*y2);
+      by +=  coefficents[3] * (x2*x - 3.0*x*y2) + skews[3]*(y2*y - 3.0*x2*y);
       bx +=  coefficents[3] * (3.0*x2*y - y2*y) + skews[3]*(x2*x - 3.0*x*y2);
       [[fallthrough]];
     case 3:      // Sextupole momentum:
-      by += -coefficents[2] * (x2 - y2) + skews[2] * 2.0 * xy;
+      by +=  coefficents[2] * (x2 - y2) - skews[2] * 2.0 * xy;
       bx +=  coefficents[2] * 2.0 * xy + skews[2] * (x2 - y2);
```
petricm: Shouldn't this be
```cpp
by +=  0.5 * coefficents[2] * (x2 - y2) - skews[2] * xy;
bx +=  coefficents[2] * xy + 0.5 * skews[2] * (x2 - y2);
```
to have the coefficient[1] vs. coefficient[2] properly normalised?

DDCore/src/FieldTypes.cpp
```diff
@@ -99,16 +99,16 @@ void MultipoleField::fieldComponents(const double* pos, double* field) {
     double y2 = y*y;
     switch(coefficents.size())  {
     case 4:      // Ocupole momentum
-      by +=  coefficents[3] * (x2*x - 3.0*x*y2) + skews[3]*(y2*y - 3.0*x*y2);
```
petricm: I think this fix is correct, but the same comment as below, shouldn't there be a 1/6 in front for normalization?

danieljeans:
I believe that what I implemented in the code is consistent with the definition in the documentation in the header file, which is in turn consistent with the references to which it links: for the nth multipole, 
By + i Bx = ( b + i a ) ( x + i y ) ^ (n-1). [b is the "coefficient", a is the "skew".]
(eg see p10 of http://cas.web.cern.ch/sites/cas.web.cern.ch/files/lectures/bruges-2009/wolski-1.pdf)
I'm sure that definitions with different normalisations are possible, but this one seems standard at least in the accelerator community.

petricm:
I cannot agree with the your statement 

> this one seems standard at least in the accelerator community.

If you look at [this lecture](http://uspas.fnal.gov/materials/09VU/Lecture2.pdf) from another accelerator school you find an explicit definition for the terms (page 23) which give you different normalization (the form of the expression is however the same).

danieljeans:
@petricm thanks for pointing this out to me.
I have no preference as to which convention to use, as long as it's consistent and well documented.

petricm:
I'll ask someone in the magnets department for clarification and let's see what they say...

petricm:
I still did not receive a reply but for book reference you can look at [this book](https://books.google.ch/books?id=oR0_CgAAQBAJ&pg=PA130&lpg=PA130#v=onepage&q&f=false) with the definition the same as the USPAS slides.

petricm:
I have received a reply that we should consider _Field Computation for Accelerator Magnets_ by Stephan Russenschuck as the reference for all matters accelerator magnets. If you with you can download the book inside the CERN network [here](http://onlinelibrary.wiley.com/book/10.1002/9783527635467).

The conclusion is that a dipole is sextupole is parameterised as 
```
Bx(x,y)= 1/r_0^2 ( A3(r_0) (x^2-y^2) + 2B3(r_0)xy)
```
where r_0 is an arbitrary reference radius, so the boundary conditions defines you normalization. 

If you however do a multipole expansion as our code does, so add a dipole field to a quadropole, then there is a **1/(n-1)!** term.

So I would like to kindly ask you if you can fix the code to include this.

danieljeans:
thanks for the info. It's easy to include the (n-1)! scaling, I'll do it right away.

if you want the ability to explicitly include the r_0 normalisation (ie to be able to specify the coefficients and skews at any particular reference radius), we would need an extra parameter. 
In my opinion, we can explicitly say that r_0 = 1 and avoid this complication: do you agree?

petricm:
I went through the equations again and they are correct.

PR #325, opened 2018-03-01
Add dictionaries and python interface  to DDRec

### LINKED ISSUE #306: DDRec in python
Hi DD4hep devs !

I looked in the available tools in the DD4hep package and I haven't found a python interface for DDRec as it is available for DDG4 (and thus ddsim). 
I have no idea how difficult it is and how much effort I would take to make this interface.
I want to use it for the calibration software, LCCalibration, mainly to to compute the total interaction length in  sub-detectors, in particular in the calorimeters. 
As an ugly work-around, I'm converting the compact file to a Gear file and then parse it with the lxml python module. It currently works, but as we would like to get rid of Gear in the ILD reconstruction chain and also like to remove this (ugly) conversion from the calibration software.

MarkusFrankATcernch:
Resolves #306 
Closes #312 
BEGINRELEASENOTES
- Add dictionaries to `DDRec`. 
  - Consequently add the python interface to access the exported classes.

ENDRELEASENOTES

### DIFF
--- a/DDCond/CMakeLists.txt
+++ b/DDCond/CMakeLists.txt
@@ -17,7 +17,8 @@ dd4hep_package(    DDCond
 dd4hep_add_dictionary( G__DDCond
   SOURCES include/ROOT/Warnings.h
   src/ConditionsDictionary.h
-  LINKDEF include/ROOT/LinkDef.h )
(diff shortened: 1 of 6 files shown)


petricm:
Why are there only deprecated warnings in the dictionary creation?

PR #333, opened 2018-03-08
graphicalMaterialScan: add electric and magnetic fields scans

danieljeans:
BEGINRELEASENOTES
- changed utility name from graphicalMaterialScan -> graphicalScan
- add visualisation of electric and magnetic fields (switched by parameter)
- change order of x/y/z range parameters for ease of use
- added parameter for name of output root file
ENDRELEASENOTES


petricm:
Thank you for this new addition, I am just wondering if putting it into `graphicalMaterialScan` is the right place, given that it visualizes the field. Or maybe we should rename this tool to `graphicalScan` to indicate it is doing more?

danieljeans:
I think it's rather natural to have it in the same package. It's just a few added lines of code, whereas if split into a separate utility, a lot of code duplication would be needed.
I am not against renaming this utility to graphicalScan (which I agree is more descriptive), but this may be disruptive to users. Is there some dd4hep policy in this regard?

petricm:
I would then suggest that you rename the tool to `graphicalScan`.
If you would be willing it would be appreciated if you could an option to do just the material scan or just the field scan if one does not want to 'waste' time with the other and default can be do both.

danieljeans:
renamed, and added requested feature to select type of scan (material or field).

PR #344, opened 2018-03-27
Move DDCMS into examples

petricm:
BEGINRELEASENOTES
- Move `DDCMS` into examples as it is not core funcionality

ENDRELEASENOTES


gaede:
@MarkusFrankATcernch are you happy with this ? I would like to make a new release tag and this could be included ...

PR #351, opened 2018-04-05
Add possibility to reference pools in used slices

MarkusFrankATcernch:
BEGINRELEASENOTES
- To avoid unwanted disappearing conditions sub pools, a conditions slice may be instructed to collect shared references to the used pools in the slice.
- For python:
  - Move DDG4/SystemOfUnits.py to DDG4/g4units.py
    Keep SystemOfUnits.py with deprecation warning
  - move DD4hep.py to dd4hep.py. 
    import dd4hep also imports all TGeoUnit units into its namespace.
    Hence: import dd4hep; print dd4hep.m  gives: "100.0"
  - DD4hep.py is kept for backwards compatibility issuing a deprecation warning
  - DDG4.py: imports g4units    as G4Units:         DDG4.G4Units.m etc.
  - DDG4.py: imports TGeoUnit as TGeo4Units:  DDG4.TGeoUnits.m etc.
  - Some problem with replacing DD4hepUnits.h with TGeoSystemOfUnits.h
    Surface test complains. To be investigated. Keep old DD4hepUnits for the time being.
ENDRELEASENOTES


MarkusFrankATcernch:
@petricm  TGeoPhysicalConstants.h is not necessary to be included in the python interface.
These values are grouped together with TGeoSystemOfUnits in the namespace TGeoUnits.

Issue #356, opened 2018-04-06
Detector field is valid while InitDD4hep not set


ebrondol:
- OS version: SLC6
 - Compiler version: GCC 6.2
 - Reproduced by: 
```
source /cvmfs/clicdp.cern.ch/iLCSoft/builds/2018-03-02/x86_64-slc6-gcc62-opt/init_ilcsoft.sh
git clone https://github.com/iLCSoft/CLICPerformance.git
cd CLICPerformance
mkdir build
cd build
cmake -C $ILCSOFT/ILCSoft.cmake ..
make install
cd ..
```
copying in the local folder the following input files:
-- http://ebrondol.web.cern.ch/ebrondol/CLIC/test_PfoSelector.xml
-- http://ebrondol.web.cern.ch/ebrondol/CLIC/example_rec.slcio

and run 
```
Marlin test_PfoSelector.xml
```
 - Output: the full log obtained can be found in 
-- http://ebrondol.web.cern.ch/ebrondol/CLIC/test_PfoSelector_output.log
 - Goal: 
InitDD4hep is commented out from the xml file.
The uncorrected initialization of the geometry should be spotted in 
https://github.com/iLCSoft/MarlinUtil/blob/master/source/src/GeometryUtil.cc#L17
but no exception is thrown because `theDetector.field().isValid()` is considered `true`.
The result is that the file is executed using wrong values of the geometry.

ebrondol:
The problem seems to be related to the setting of `m_field` in `DetectorData`:
https://github.com/AIDASoft/DD4hep/blob/master/DDCore/src/DetectorData.cpp#L155

petricm:
@MarkusFrankATcernch: "The return value of `det.field().isValid()` is always `true` due to inheritance and this is a design feature."

PR #358, opened 2018-04-09
Add field test

ebrondol:
BEGINRELEASENOTES

- Add test to check the validity of the field depending on the geometry status.

ENDRELEASENOTES

In this moment the test fails.

Test described problem in #356


petricm:
The return value of `det.field().isValid()` is always `true`

PR #381, opened 2018-05-14
WIP: use tau- for CLICSiDXML examples/tests

gaede:
DO NOT MERGE !

Used to demonstrate the problem reported in #383 ( this is why the CI  fails).

### DIFF
--- a/examples/CLICSiD/sim/sequences.xml
+++ b/examples/CLICSiD/sim/sequences.xml
@@ -12,7 +12,7 @@
     
        To run batch: set UI=""
   -->
-  <kernel UI="" NumEvents="10"/>  
+  <kernel UI="" NumEvents="100"/>
 
   <!-- List of globally known actions. These actions will be registered with the Geant4Kernel 
        object and may be retrieved by there using the action name.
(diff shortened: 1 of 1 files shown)


petricm:
It might be easier just to use this
`/cvmfs/clicdp.cern.ch/iLCSoft/builds/nightly/x86_64-slc6-gcc7-opt/init_ilcsoft.sh`
this is identical to the CI

gaede:
@petricm I was not able to build the examples with `/cvmfs/clicdp.cern.ch/iLCSoft/builds/nightly/x86_64-slc6-gcc7-opt/init_ilcsoft.sh` - see #382.
However I did run this nightly with single taus in the ILD and CLIC detector (o3_v14) and could reproduce the crash. Surprising that you did not see this in your physics simulation so far...

andresailer:
@gaede We didn't simulate any Taus in the central production so far...
@webermat have you seen this exception in your Tau samples?

gaede:
Ok thanks. Please let me know if you have a way of running real tau samples with pre-assigned decays through ddsim. I suspect that something is wrong in our simulation setup, as the bug seems to happen with the built in decays but we really are using pre-assigned ones (at least we believe that this is what we do) ...

PR #389, opened 2018-05-22
fix bug in input handling

### LINKED ISSUE #387: Incorrect treatment of documentation lines in Geant4InputHandling
We observed a sever bug in events with charged lepton pairs - in particular w/ taus.
In many of these events one of the two taus is actually simulated twice - once with the correct pre-assigned decay and once w/ a decay chosen from Geant4.  This is indirectly related to #383, however the actual geant4 crash is an independent bug in geant4 tau decays. These decays should never happen in generator samples w/ pre-assigned decays.
I strongly suspect this fix: https://github.com/AIDASoft/DD4hep/pull/315/commits/23bd42f9a0671e3880dcb909decf268ef285d368 committed to fix #307.
The problem occurs in events, of the form
```
tau+, tau- > 94 object
94 -> tau+, tau-
tau - -> tau-, gamma
```
here only the last tau+, tau-, gamma should have been given to geant4, but we include also the previous tau parents ( that have zero lifetime in the input file). This also causes the tau lifetime to be applied n-times by geant4, as geant4 treats this as a decay chain.
@andresailer any idea how to revert your fix and still get the correct treatment for real decay chains w/ very short lived particles ?
I guess it would need a more correct check for `float value 0` ...
Issue can be reproduced by running: `/afs/desy.de/user/g/gaede/public/4f_ZZ_semileptonic_G4crash.slcio ` through any detector w/ DDG4/ddsim.
In the output file one of the taus will have daughters added in simulation not present in the input file.

petricm:
cherry pick @gaede commit from branch rel-v01-07 to fix input handling 
BEGINRELEASENOTES
 - fix bug in input handling, for details see discussion #387 
   - exclude leptons with zero lifetime from Geant4

ENDRELEASENOTES

Resolves #387 
Closes #381

### DIFF
--- a/DDG4/src/Geant4InputHandling.cpp
+++ b/DDG4/src/Geant4InputHandling.cpp
@@ -384,10 +384,12 @@ getRelevant(set<int>& visited,
     double proper_time_Precision = pow(10.,-DBL_DIG)*me*fmax(fabs(p->time),fabs(dp->time));
     bool isProperTimeZero = (proper_time <= proper_time_Precision);
 
+    const std::set<int> leptonPDGs{11,13,15,17};
     // -- remove original if ---
     bool rejectParticle = not p.definition() // completely unknown to geant4
       or (rejectPDGs.count(abs(p->pdgID)) != 0) // quarks, gluon, "strings", W, Z etc.
-      or (isProperTimeZero and p.definition()->GetPDGStable()); // initial state electrons, etc.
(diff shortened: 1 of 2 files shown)


gaede:
... but in a sense the bug fix is applied twice, once in the branch release v01-07-01 and once in the master. This way v01-07-01 will never be documented in future releases....

PR #412, opened 2018-06-26
 Fix the sparse combined hits from two volumes in this silicon tracker SD accumulation algorithm. 

shaojunlu:
BEGINRELEASENOTES
- Added one more if statement: If the track went into new Volume, 
  - then extracted the hit in previous Volume,
  - and start a new hit in this current Volume,
    - in this current process, also allow the same following treatments for the new hit.

ENDRELEASENOTES


MarkusFrankATcernch:
I think the debugging is fine. I did not check the default. Default should be without printing.

PR #437, opened 2018-09-11
Support all CMake booleans

HadrienG2:
CMake supports other Boolean truth values than "ON" and "OFF", including for example "TRUE"/"FALSE", "0"/"1", and "YES"/"NO". Therefore, hardcoding boolean comparisons to "ON"/"OFF" will cause subtle problems when CMake is used in an unintended way and is best avoided.

BEGINRELEASENOTES
 - Remove if string equals ON/OFF in cmake IF statements and check default CMake truth values

ENDRELEASENOTES


petricm:
thank you for your contribution and I see your point, but this in it's current state breaks everything. Please fix the PR.

HadrienG2:
You're right, I missed some use of undefined variables. Adding a check of defined-ness...

HadrienG2:
Ok, this version looks good.

petricm:
thank you for fixing it, please just squash the commits.

HadrienG2:
Done!

MarkusFrankATcernch:
Marko,
if you agree with this solution,please merge.

PR #447, opened 2018-10-18
Fix plugin manager cmake file 

MarkusFrankATcernch:
BEGINRELEASENOTES
-  Fix plugin manager cmake file -- remove explicit dependency on c++ standard.
This caused inconsistencies when building for a specified standard.

ENDRELEASENOTES


petricm:
replaced by #448

petricm:
Sorry did not see you fixed it

PR #457, opened 2018-11-13
Provide support for volume/shape divisions 

MarkusFrankATcernch:
BEGINRELEASENOTES
## Provide support for Volume divisions.
Since DD4hep requires Volumes (aka `TGeoVolume`) and PlacedVolumes (aka `TGeoNode`) to be enhanced with the user extension mechanism, therefore shape divisions **must** be done using the division mechanism of the DD4hep shape or the volume wrapper. Otherwise the enhancements are not added and you will get an exception when DD4hep is closing the geometry. The same argument holds when a division is made from a `Volume`. Unfortunately there is no reasonable way to intercept this call to the `TGeo` objects - except to the sub-class each of them, which is not really acceptable either.

Hence: **If you use DD4hep: Never call the raw TGeo routines.**
   
For any further documentation please see the following ROOT documentation on [TGeo](http://root.cern.ch/root/html/TGeoVolume.html)

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/Shapes.h
+++ b/DDCore/include/DD4hep/Shapes.h
@@ -50,6 +50,7 @@
 /// Namespace for the AIDA detector description toolkit
 namespace dd4hep {
 
+  class Volume;
 
   /** Utitlity functions   */
   /// Pretty print of solid attributes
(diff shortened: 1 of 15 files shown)


petricm:
Maybe it would be nice to have the option to call the check separately, without the need to close the volume?

MarkusFrankATcernch:
@petricm I do not understand your comment. What do you mean exactly ?

petricm:
you write
> Otherwise the enhancements are not added and you will get an exception when DD4hep is closing the geometry.

Can I call "test" and check for the exception without calling close geometry

MarkusFrankATcernch:
In principle you get an exception whenever you do something with the volume, which is served by the user extension. This is what Yana experienced and what triggered this development.
When closing the geometry in dd4hep this is only the first occurrence where this happens.

petricm:
OK, I understand thank you for the clarification.

PR #467, opened 2018-12-05
Harmonize argument names in Shapes.h with their actual functionality 

MarkusFrankATcernch:
BEGINRELEASENOTES
# Harmonize argument names in Shapes.h with their actual functionality 
For many shapes in DD4hep/Shapes.h the argument names were misleading:
very often deltaPhi was mentioned, whereas the code actually used instead
of (phi, deltaphi) the input arguments to the ROOT constructors (startPhi,endPhi)
or (phi1, phi2).  Wherever ROOT uses (startPhi,endPhi) the argument names were changed accordingly.
Please note:
a bug was found in the legacy constructor:
    /// Legacy: Constructor to create a new identifiable tube object with attribute initialization
    Tube(const std::string& nam, double rmin, double rmax, double dz, double startPhi, double endPhi)
(which is probably not used by anyone....).
Here opposite to all other constructors delta_phi was used as such - in contradiction to other constructors of the same class. This was rectified.

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/Conditions.h
+++ b/DDCore/include/DD4hep/Conditions.h
@@ -332,7 +332,9 @@ namespace dd4hep {
     /// Operator less (for map insertions) using hash value
     bool operator<(const Condition::key_type compare)  const;
     /// Automatic conversion to the hashed representation of the key object
-    operator Condition::key_type () const             {  return hash;     }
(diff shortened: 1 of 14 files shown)


petricm:
Please add a deprecated pragma to the legacy constructor

PR #489, opened 2019-03-05
Birks Law bug fix

ebrianne:
Due to a complete change in geant4 versions over 10.03, the class [G4EmSaturation](https://github.com/Geant4/geant4/blob/v10.3.2/source/processes/electromagnetic/utils/include/G4EmSaturation.hh) needs to be initialized via `InitialiseG4Saturation()`. This sets the default birks coefficients that are in geant4. If this method is not called, the birks law does not work as the coefficient is set to 0.

BEGINRELEASENOTES
- Added Initialization of G4EmSaturation to initialize birks coefficients - for g4 version > 10.03

ENDRELEASENOTES

### DIFF
--- a/DDG4/src/Geant4StepHandler.cpp
+++ b/DDG4/src/Geant4StepHandler.cpp
@@ -121,7 +121,11 @@ double Geant4StepHandler::birkAttenuation() const    {
 #endif
 
 #if G4VERSION_NUMBER >= 1003
-  s_emSaturation.InitialiseG4Saturation();
+  static bool s_initialised = false;
+  if(not s_initialised) {
+      s_emSaturation.InitialiseG4Saturation();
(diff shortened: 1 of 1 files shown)


andresailer:
The `birkAttenuation` function is called for every step, so now for every step the initialise function would be called. While the G4EMSaturation object is static, and therefore only created once. I think the initialisation should be protected through a static boolean flag.
```c++
#if G4VERSION_NUMBER >= 1003
static bool s_initialised = false;
if(not s_initialised) {
   s_emSaturation.InitialiseG4Saturation();
   s_initialised = true;
}
#endif
```
Alternatively a wrapper object could be created

ebrianne:
The protection for the initialization with a static boolean flag has been added

PR #503, opened 2019-04-16
DDG4: add possibility to set leptonPDGs i.e. ignoring particles when …

### LINKED ISSUE #390: Make rejected list configurable in Geant4InputHandling
The current solution for the bug (#387) in Geant4InputHandling for documentation particles is to introduce a hard-coded list of rejected PGD
```cpp
const std::set<int> leptonPDGs{11,13,15,17};
...
(isProperTimeZero and leptonPDGs.count(abs(p->pdgID)) != 0 ) 
```
I would be good to generalize this list to be steerable, with current hard-coded values being the defaults.

andresailer:
…properTime is Zero

refactor the way rejectPDGs and leptonPDGs are passed through the reading of primary particles

BEGINRELEASENOTES
- DDG4: DDSim add option Physics.zeroTimePDG to configure ignoring particles of given PDG when their properTime is ZERO, e.g. charged leptons undergoing FSR, fixes #390 
- DDG4: DDSim: fix parsing of rejectPDGs values from the command line

ENDRELEASENOTES

### DIFF
--- a/DDG4/include/DDG4/Geant4PrimaryHandler.h
+++ b/DDG4/include/DDG4/Geant4PrimaryHandler.h
@@ -21,6 +21,23 @@ namespace dd4hep {
 
   /// Namespace for the Geant4 based simulation part of the AIDA detector description toolkit
   namespace sim {
+    /// Geant4PrimaryConfig to hold configuration for PrimaryHandlers
+    class Geant4PrimaryConfig {
+    public:
+      /// particles with these PDG IDs are not passed to geant for simulation
+      std::set<int> m_rejectPDGs = {1, 2, 3, 4, 5, 6, 21, 23, 24, 25};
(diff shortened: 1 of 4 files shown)


DDG4/include/DDG4/Geant4PrimaryHandler.h
```diff
@@ -41,8 +58,7 @@ namespace dd4hep {
       virtual void operator()(G4Event* event);
 
     public:
-      /// particles with these PDG IDs are not passed to geant for simulation
-      std::set<int> m_rejectPDGs={1,2,3,4,5,6,21,23,24};
+      Geant4PrimaryConfig m_primaryConfig;
```
andresailer:
```suggestion
      Geant4PrimaryConfig m_primaryConfig{};
```

PR #508, opened 2019-04-29
Add properties to Geant4GDMLWriteAction

### LINKED ISSUE #507: Convert geometry with Geant4GDMLWriteAction in DDG4
Hi! I'm glad to find that you have add an example about writing Geant4 GDML file.
When I try this new plugin, I find that some information like energy cut or region is lost in the generated gdml file.
It seems that you create the gdml file by calling G4GDMLParser, which supports writing auxiliary and region information into gdml file. So why these information are lost?
Looking forward to your reply!

MarkusFrankATcernch:
Add properties to Geant4GDMLWriteAction to steer writing of regions, cuts and sensitive detectors. Resolves #507.

BEGINRELEASENOTES
* Geant4GDMLWriteAction:
  * Add properties to Geant4GDMLWriteAction to steer writing of regions, cuts and sensitive detectors. See github issue #507
    * Property: Export region information to the GDML: ExportRegions, default: True
    * Property: Export energy cut information to the GDML: ExportEnergyCuts, default: True
    * Property: Export sensitive detector information to the GDML: ExportSensitiveDetectors, default: True
  * **Note: The Geant4 physics list must be initialized BEFORE invoking the writer with options. Otherwise the particle definitions are missing! If you ONLY want to dump the geometry to GDML you must call**
    ```
       /run/beamOn 0
    ```
    before writing the GDML file!
    You also need to setup a minimal generation action like:
     ```py
     sid.geant4.setupGun('Gun','pi-',10*GeV,Standalone=True)
     ```
ENDRELEASENOTES

### DIFF
--- a/DDG4/include/DDG4/Geant4UIManager.h
+++ b/DDG4/include/DDG4/Geant4UIManager.h
@@ -60,6 +60,8 @@ namespace dd4hep {
       std::string m_visSetup;
       /// Property: Array of macro files to be chained
       std::vector<std::string> m_commands;
+      /// Property: Array of macro files to be chained and executed AFTER running
+      std::vector<std::string> m_postRunCommands;
       /// Property: Array of commands to be chained
       std::vector<std::string> m_macros;
       /// Property: New prompt if the user wants to change it. (Default is do nothing)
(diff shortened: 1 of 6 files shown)


MarkusFrankATcernch:
See https://github.com/AIDASoft/DD4hep/pull/508
Please read the comments in: Geant4GDMLWriteAction

PR #548, opened 2019-08-09
Shapes: Sphere fix conversion for startTheta

andresailer:
BEGINRELEASENOTES
- Shapes: fix conversion of `startTheta` for Sphere::setDimensions
- Shapes::get_shape_dimension: add return value conversion for angles to internal unit (radians)

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/Shapes.cpp
+++ b/DDCore/src/Shapes.cpp
@@ -864,7 +864,7 @@ void Sphere::make(const std::string& nam, double rmin, double rmax, double start
 
 /// Set the Sphere dimensions
 Sphere& Sphere::setDimensions(double rmin, double rmax, double startTheta, double endTheta, double startPhi, double endPhi) {
-  double params[] = { rmin, rmax, startTheta, endTheta/units::deg, startPhi/units::deg, endPhi/units::deg };
+  double params[] = { rmin, rmax, startTheta/units::deg, endTheta/units::deg, startPhi/units::deg, endPhi/units::deg };
   _setDimensions(params);
   return *this;
 }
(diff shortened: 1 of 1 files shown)


andresailer:
Shouldn't these angles (`GetTheta`, etc.) have `*units::deg` ?

https://github.com/AIDASoft/DD4hep/blob/9f9459d008972c8944ae8185b5ee4da16010b3e5/DDCore/src/Shapes.cpp#L117

MarkusFrankATcernch:
I do not know exactly. 
This piece of code was triggered by CMS. I do not know exactly what they expect here to be returned.
Otherwise you are right -- and there are several other inconsistencies then.

andresailer:
So maybe @ianna can tell us.
I would think that chains of `get_shape_dimensions` and `setDimensions` should work without manual unit transformations.

ianna:
we assume that the returned angles are in **deg** and convert them to **rad** ourselves

MarkusFrankATcernch:
@ianna : This approach looks very inconsistent. 
The dd4hep shape constructors take radians, but setDimension/getDimension use degrees. 
This does not sound logical to me.

ianna:
@MarkusFrankATcernch - agree, if you switch setDimension/getDimension to rad for consistency, we adapt our code.

andresailer:
Then I will add the unit conversion in the getDimensions (get_shape_dimensions) function.

ianna:
@cvuosalo - FYI, we'll need to keep an eye on it and remove the deg to rad conversion from the code when this is merged to the master and the dd4hep external is updated.

andresailer:
Added the conversion, someone please double check

PR #549, opened 2019-08-12
Move to new Gaudi plugin manager

MarkusFrankATcernch:
BEGINRELEASENOTES
- Adopted new Gaudi plugin manager V2. V1 can be enabled using compile switch in DD4hep/config.h.
  Removed the traced from the ROOT5 Reflex based plugin service.
  The new plugin service depends on Boost::file_system and Boost::system.
  This must be fixed for MAC by a MAC expert. Until then MAC is broken.
- Improve GDML saving from ROOT. (requires ROOT >= 6.20)
- Fix ROOT persistency for the volume manager.
- Fix Geant4FieldTrackingSetup: Issue warning if the G4MagIntegratorStepper cannot be created.
  Do we need such warnings also for the G4EquationOfMotion ?
- BTW: The factory to create an instance of G4ClassicalRK4 is called "ClassicalRK4".
  By default in ddsim never a G4MagIntegratorStepper was created...
  Changed all occurrences of G4ClassicalRK4 to ClassicalRK4.
- Examples: based the CLICSiD example on the XML sources of DDDetectors.
  This ensures XML sources match C++ sources....
ENDRELEASENOTES


petricm:
As for 

>  Do we need such warnings also for the G4EquationOfMotion ?

YES. And basically this should not be a WARNING but in both cases this should be FATAL

And I don't get this comment
>By default in ddsim never a G4MagIntegratorStepper was created...

What happened if you specified "G4ClassicalRK4" in ddsim?

petricm:
And from which version of Gaudi did you take the plugin manager?

MarkusFrankATcernch:
@petricm 

- Gaudi plugin manager from Gaudi v32r1
- What happened if you specified "G4ClassicalRK4" in ddsim?
  The factory was not called, hence some Geant4 default kicked in.
I will make it FATAL on the next commit.

Please have a look why the listcomponents does not work on MAC. I spent 2 full days and gave up- finally.

petricm:
You forgot to mention, that with this version of the GaudiPluginManger we are dropping C++11, because you need C++14 to compile it.

MarkusFrankATcernch:
Oh. This I was not even aware of.
Will C++14 cause problems with ILCSoft ?

gaede:
No problem. iLCSoft works w/ c++14 and also almost completely w/ c++17.

PR #550, opened 2019-08-13
Fix plugin mac

gaede:
BEGINRELEASENOTES
- make compatible with MacOS (10.14.6)
     - address latest developments w/ new Gaudi Plugin Manager

ENDRELEASENOTES


examples/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ dd4hep_enable_tests (
   DDG4_MySensDet
   DDCodex
   )
-if ( "${ROOT_VERSION}" VERSION_GREATER "6.17.00" )
+if ( "${ROOT_VERSION}" VERSION_GREATER "6.20.00" )
```
andresailer:
```suggestion
if ( "${ROOT_VERSION}" VERSION_GREATER "6.17.99" )
```

petricm:
I assume you turned SIP off or how did you get the rest to work?

andresailer:
Can you run
```
otool -D .../libboost_filesystem.dylib
```
on you mac please?

examples/CMakeLists.txt
gaede: Changed to `if ( "${ROOT_VERSION}" VERSION_GREATER_EQUAL "6.20.00" )` which is
what is required in `OpticalSurfaces/CMakeLists.txt`

gaede:
```
13:36:25-gaede/data/ilcsoft/HEAD/DD4hep/HEAD/build (fix_plugin_mac)$ otool -D /usr/local/lib/libboost_filesystem.dylib
/usr/local/lib/libboost_filesystem.dylib:
/usr/local/opt/boost/lib/libboost_filesystem.dylib
```
@petricm yes SIP is always turned off on my Mac - would not know how to develop software otherwise ...

petricm:
This is an rpathed version of Boost, which is good. It's not the normal/default way boost behaves on a Mac, but it is probably the only way to solve this.

gaede:
Ok - then in this case homebrew seems to have done the right thing...
Nevertheless I find it increasingly difficult to get things working on Mac and constantly have to
fiddle around with '$PATH` and `$DYLD_LIBRARY_PATH` in order to get a consistent environment....

peterkostka:
On macOS 10.14.6 (18G87) - without(!) applying the fix :
using fink (Python 2.7.16),  root v6.18.00,  geant4.10.05.p01,  boost v1.70.0

> otool -D /opt/boost/lib/libboost_filesystem.dylib
/opt/boost/lib/libboost_filesystem.dylib:
@rpath/libboost_filesystem.dylib

The following tests FAILED:
	 12 - test_ddsim (Failed)
 *** Break *** segmentation violation
INFO: |  Failed to load DDG4 library:                                                                          |
INFO: |   int TUnixSystem::Load(const char* module, const char* entry = "", bool system = kFALSE) =>
    problem in C++; program state has been reset

and:
100% tests passed, 0 tests failed out of 236 - examples 

> csrutil status
System Integrity Protection status: enabled
I did NOT disabled it in the past!

petricm:
@peterkostka where did you get your boost from?

peterkostka:
I downloaded boost from   https://www.boost.org 
Did an install standalone.

petricm:
Can you please post the exact install commands, because I don't understand where did you get the `@rpath/libboost_filesystem.dylib`

peterkostka:
./bootstrap.sh
./b2 -a --cxxstd=17 --reconfigure --prefix=$SWDIR/boost --build-dir=$SWDIR/boost/build install

examples/CMakeLists.txt
andresailer: There is just one test in OpticalSurfaces which requires root 6.20, the others require 6.18.

MarkusFrankATcernch: What it needs is ROOT 6.18.
The GDML example anyhow explicitly requires 6.20.

gaede: Yes, but if you then have ROOT_VERS<6.20 the cmake crashes - so that's not what you want. I guess the 'OpticalSurfaces/CMakeLists.txt' should be changed accordingly (try to build the GDML stuff only if ROOT_VERS>=6.20).

gaede: Ok, I realize that Markus had fixed it in the meantime - so reverted back to 
` "${ROOT_VERSION}" VERSION_GREATER_EQUAL "6.18.00"`.

PR #566, opened 2019-09-30
Dictionary and cmake fixes

### LINKED ISSUE #565: Python bindings work only with dd4hep sources present
Running 

```
python -c "import dd4hep"
```

fails with a [lengthy error message](https://gist.github.com/vvolkl/26b364f09176634278cdcec03b367b28) if the sources used to build dd4hep are no longer in place (so affecting most packaged versions of dd4hep) Can be reproduced by temporarily moving the dd4hep source folder.

This affects v1-10 on centos7 using LCG release 96b for example (`source /cvmfs/sft.cern.ch/lcg/views/LCG_96b/x86_64-centos7-gcc62-opt/setup.sh` on lxplus )

I first thought this was an issue with the lcg release, after some discussion (https://sft.its.cern.ch/jira/browse/SPI-1461) we determined that it can be reproduced locally.

andresailer:
BEGINRELEASENOTES
- PythonBindings: fix issue when source files were not available, fixes #565 
- CMake: drop DDCores dependency on DD4hepGaudiPluginMgr
- CMake: DD4hepConfig: use find_dependency instead of find_package

ENDRELEASENOTES


andresailer:
Reading https://sft.its.cern.ch/jira/browse/ROOT-10040
lead me to believe that this is the only fix

PR #579, opened 2019-10-16
CMake: add possibility to compile static libraries instead of shared libraries

### LINKED ISSUE #493: Build DD4Hep libraries static - cmake option
- OS version: e.g. CC7
 - Compiler version: e.g GCC 7.3.1 (7.4.1, 8.3)
 - DD4hep version: latest commits on master
 - Goal: Build static DD4Hep libraries

Hi,
we are using the DD4HEP_USE_GEANT4=ON option which builds  libs linked against Geant4 libs. What we are trying to do is to build this libs statically, and link them against 
statically built Geant4 libs (Geant have an option that builds their libs statically with .a extension)

Can we have an option for building the DD4Hep libs statically, and provided libs are built how to link them against the static Geant4 libs ? 

Thanks in advance,
Cheers,
Mircho

@ianna @smuzaffar

andresailer:
BEGINRELEASENOTES
- CMake: Add possibility to build only shared libraries, fixes #493: 
  - Usage `cmake ... -D BUILD_SHARED_LIBS=OFF ...` 

ENDRELEASENOTES


petricm:
@mrodozov does this fulfill the CMS requirements?

mrodozov:
It works. I took this branch and build it with out setup. Adding the suggested flag builds the libs static.
It fulfills the CMS requirements in short. 
One thing that didn't work was adding -D DD4HEP_BUILD_PACKAGES="DDG4" flag
resulted in 

```
CMake Error at DDTest/CMakeLists.txt:27 (add_executable):                                                                                                                                                                                 
  Target "test_PolarGridRPhi2" links to target "DD4hep::DDRec" but the target                                                                                                                                                             
  was not found.  Perhaps a find_package() call is missing for an IMPORTED                                                                                                                                                                
  target, or an ALIAS target is missing?  
```

but that could go as separate issue as its unrelated. I'm just reporting it anyway
Thank you Andre, this will save us the patching from now on.

andresailer:
@mrodozov thanks for the confirmation. 
To fix the problem with test_* I think the sensible solution would be `cmake ... -DBUILD_TESTING=OFF ...`

mrodozov:
I confirm with this option it configures.

.dd4hep-ci.d/compile_and_test.sh
```diff
@@ -13,8 +13,15 @@ cmake -GNinja -DDD4HEP_USE_GEANT4=ON \
...
     -DCMAKE_CXX_FLAGS="-fdiagnostics-color=always -Werror"  \
-    -DCMAKE_CXX_STANDARD=${STANDARD} ${CMAKE_ARGS} .. && \
-ninja && \
+    -DBUILD_SHARED_LIBS=${SHARED} \
+    -DCMAKE_CXX_STANDARD=${STANDARD} .. && \
+    ninja
+
+if [[ "${SHARED}" == "OFF" ]];
+then
+    exit 0;
```
petricm: Like this it will always pass since there is no and and the last return code will be given to travis, I think maybe something like this would be better
```suggestion
    exit $?;
```

petricm: Sorry, I was premature, the evaluation of the if statement overrides `$?` so you have to save the return value from the build command in a temp variable...

andresailer: Done, I think.

petricm: In principle, I think you still have to `and` the build result with the rest of the CI chain after the if
```
BUILD_RESULT && \
ninja install && \ 
. ../bin/thisdd4hep.sh && \ 
...
```
otherwise the non static chains will not work properly. But it's probably not possible that `BUILD_RESULT` would be 1 and the subsequent CI commands would results in a positive result. So it's more a question of taste.

andresailer: Since `ninja install` will rerun compilation or linking, it might be that things are working in the second pass of the compilation attempt, which is not what we want. So I added a check for BUILD_RESULT

PR #586, opened 2019-10-22
remove deprecated rootcling flags

petricm:
BEGINRELEASENOTES
- Remove deprecated rootcling flags (`-cint`, `-c`, `-p`, `-std=c++`) from dictionary creation script

ENDRELEASENOTES

### DIFF
--- a/cmake/DD4hepBuild.cmake
+++ b/cmake/DD4hepBuild.cmake
@@ -559,7 +559,7 @@ function(dd4hep_add_dictionary dictionary )
 
   SET(COMP_DEFS )
   file(GENERATE OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${dictionary}_arguments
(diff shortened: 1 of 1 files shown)


cmake/DD4hepBuild.cmake
```diff
@@ -559,7 +559,7 @@ function(dd4hep_add_dictionary dictionary )
 
   SET(COMP_DEFS )
   file(GENERATE OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${dictionary}_arguments
-    CONTENT "${ROOT_rootcling_CMD} -cint -f ${dictionary}.cxx -s ${output_dir}/${dictionary} -inlineInputHeader -c -p ${ARG_OPTIONS} -std=c++${CMAKE_CXX_STANDARD}  \
```
andresailer:
```
-std=c++${CMAKE_CXX_STANDARD}
```
the warning was just because CMAKE_CXX_STANDARD wasn't set when you ran, this should be kept I think

petricm: I tried to add to the examples CMakeList the `CMAKE_CXX_STANDARD`, but I still get
```
warning: Please remove the deprecated flag -std=c++14
```
I also don't see the `-std` flag in the rootcling help

andresailer: Indeed, I stand corrected!

PR #592, opened 2019-10-28
Fix AClick Tests (fixes #591)

### LINKED ISSUE #591: Moving to minor version 11 broke some of the tests
For example this test `t_CLICSiD_DDG4_CLICSiDXML_as_AClick_LONGTEST` now times out, instead of finishing successfully after 25 seconds. Trying things locally the test passes when I move the version back to 1.10, but fails when I move to 1.11.
I am not sure if the test somehow picks up DD4hep libraries installed on CVMFS or somewhere else and the test passes because of this, or if something else is broken.

There also some configurations that work:
https://gitlab.cern.ch/CLICdp/DetectorSoftware/DD4hep/pipelines/1179026

andresailer:
As far as I can tell rootcling (or something) does something weird with the library flags passed to it in the AClick scripts. Instead of taking the libraries from the "-L" location before it takes them from wherever it finds them first (or maybe that is correct behaviour that just doesn't work here), which means if the libraries exist already in the LCG CVMFS installation we use for some dependencies then things go wrong.

So now the DD4hep library locations are explicitly specified.

Solves #591 
Also running a pipeline on gitlab: https://gitlab.cern.ch/sailer/DD4hep/pipelines/1183456
Could be missing some other place where this is necessary

### DIFF
--- a/DDG4/examples/initAClick.C
+++ b/DDG4/examples/initAClick.C
@@ -62,7 +62,7 @@ int initAClick(const char* command=0)  {
   std::string clhep   = make_str(gSystem->Getenv("CLHEP_ROOT_DIR"));
   std::string defs    = "";
   std::string libs    = " -L"+rootsys+"/lib";
-  std::string inc     = " -I"+dd4hep+"/examples/DDG4/examples -I"+dd4hep + " -I"+dd4hep+"/include";
+  std::string inc     = " -I"+dd4hep+"/examples/DDG4/examples -I"+dd4hep + " -I"+dd4hep+"/include ";
   std::string ext = "so";
   if ( !geant4.empty() )  {
     inc  += " -I"+geant4+"/include/Geant4";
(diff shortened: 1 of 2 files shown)


andresailer:
MacOS Fix pipeline: https://gitlab.cern.ch/sailer/DD4hep/pipelines/1183512

andresailer:
https://gitlab.cern.ch/sailer/DD4hep/pipelines/1184254

andresailer:
Is it progress if the tests now fail for different reasons?
Only for clang on centos7  and with gcc62 and c++14...

andresailer:
https://gitlab.cern.ch/sailer/DD4hep/pipelines/1185673

andresailer:
https://gitlab.cern.ch/sailer/DD4hep/pipelines/1185829

andresailer:
Everything is passing now, except for a timeout in the test against dev3

Issue #595, opened 2019-11-14
G4 Materials vs ROOT Materials (NTP vs STP)


ianna:
XML definition:
```xml
  <ElementaryMaterial name="Aluminium" density="2.7*g/cm3" symbol=" " atomicWeight="26.98*g/mole" atomicNumber="13"/>

```
DD4hep converter of the **atomicWeight** attribute:
```c++
 double atomicWeight = xmat.attr<double>(DD_CMU(atomicWeight));
```
gives a very large number:
```
DD4CMS                 +++ Converting material "materials:Aluminium"                             Atomic weight 16839.593, Atomic number 13, Density: 1685.208.
DD4CMS                     ROOT definition of AL                                                 Atomic weight 26.982, Atomic number 13, Number of nucleons 27

```

ianna:
Looking at the above XML definition and executing:
```c++
for(auto vat : element.attributes())
    if(dd4hep::xml::_toString(element.attr_name(vat)).compare(std::string{"density"}) == 0)
        std::cout << dd4hep::xml::_toString(element.attr_name(vat)) << " " << dd4hep::xml::_toString(vat) << " (" << dd4hep::_toDouble(dd4hep::xml::_toString(vat)) << ")\n";
```
The XML **density** attribute is read correctly, but a ```dd4hep::_toDouble``` converted it to different from the default density units which are **mg/cm^3** in ROOT:
```
density 2.7*g/cm3 (1685.21)
```

ianna:
Normalisation after parsing does help. The densities and atomic weights are fine. The temperature is different:
```
 Material: materials:Air    density:  1.214 mg/cm3  RadL: 301.522 m    Nucl.Int.Length: 704.082 m  
                       Imean:  85.538 eV   temperature: 273.15 K  pressure:   1.00 atm

   --->  Element: CMS element (Nitrogen)   Z =  7.0   N =    14   A = 14.007 g/mole
         --->  Isotope: Nitrogen14   Z =  7   N =  14   A =  14.00 g/mole   abundance: 99.632 %
         --->  Isotope: Nitrogen15   Z =  7   N =  15   A =  15.00 g/mole   abundance:  0.368 %
          ElmMassFraction:  74.94 %  ElmAbundance  77.06 % 

   --->  Element: CMS element (Oxygen)   Z =  8.0   N =    16   A = 15.999 g/mole
[...]

PR #596, opened 2019-11-15
Disable default ROOT element table

MarkusFrankATcernch:
BEGINRELEASENOTES
- As agreed last meeting the default ROOT element table is disabled. All materials must now be entered explicitly.
- Fix some examples, which relied on the default ROOT elements being present
- Fix issue  https://github.com/AIDASoft/DD4hep/issues/595 for DDCMS:
  -- The density was not re-normalized to the units TGeo was expecting. Same for atomic weights.
  -- Fill missing elements to DDCMS/data/materials.xml (Values to be cross-checked by CMS)
ENDRELEASENOTES

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -38,12 +38,12 @@ dd4hep_configure_output()
 
 #==========================================================================
 
-SET(DD4HEP_BUILD_EXAMPLES "AlignDet CLICSiD ClientTests Conditions DDCMS DDCodex DDDB DDDigi DDG4 DDG4_MySensDet LHeD OpticalSurfaces Persistency SimpleDetector"
+SET(DD4HEP_EXAMPLES "AlignDet CLICSiD ClientTests Conditions DDCMS DDCodex DDDB DDDigi DDG4 DDG4_MySensDet LHeD OpticalSurfaces Persistency SimpleDetector"
   CACHE STRING "List of DD4hep Examples to build")
-SEPARATE_ARGUMENTS(DD4HEP_BUILD_EXAMPLES)
+SEPARATE_ARGUMENTS(DD4HEP_EXAMPLES)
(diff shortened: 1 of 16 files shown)


MarkusFrankATcernch:
I am quite sure it has to be g/cm3, because in the compact format translation we perform the same translation exercise and these values were all largely cross-checked by Frank, Andre & Co.

examples/CMakeLists.txt
```diff
@@ -38,12 +38,12 @@ dd4hep_configure_output()
 
 #==========================================================================
 
-SET(DD4HEP_BUILD_EXAMPLES "AlignDet CLICSiD ClientTests Conditions DDCMS DDCodex DDDB DDDigi DDG4 DDG4_MySensDet LHeD OpticalSurfaces Persistency SimpleDetector"
+SET(DD4HEP_EXAMPLES "AlignDet CLICSiD ClientTests Conditions DDCMS DDCodex DDDB DDDigi DDG4 DDG4_MySensDet LHeD OpticalSurfaces Persistency SimpleDetector"
   CACHE STRING "List of DD4hep Examples to build")
-SEPARATE_ARGUMENTS(DD4HEP_BUILD_EXAMPLES)
+SEPARATE_ARGUMENTS(DD4HEP_EXAMPLES)
 MESSAGE(STATUS "Will be building these examples: ${DD4HEP_BUILD_EXAMPLES}")
```
petricm: You have to change also this
```suggestion
MESSAGE(STATUS "Will be building these examples: ${DD4HEP_EXAMPLES}")
```

examples/DDCMS/CMakeLists.txt
```diff
@@ -23,13 +23,25 @@ ENDIF()
 
 dd4hep_set_compiler_flags()
 #==========================================================================
+set (Geant4_builtin_clhep_FOUND ON)
```
petricm: Why are you setting this to permanently ON?

PR #611, opened 2020-02-05
Fixed Grammar declaration location and defined a default grammar

sponce:
BEGINRELEASENOTES
- Fixed Grammar declaration location and defined a default grammar for all objects.
 The change makes the code simpler, more readable and much more usable from user point of view. Main points being :
o    the implementation of the member functions of Grammar are now in the right file, so you find them when you look for them, and you get all you need when including Grammar.h, without having to know that BasicGrammar_inl.h is actually somehow the implementation of Grammar.h
 o   BasicGrammar now has a generic implementation for its instance method, that was missing before, forcing every end user to use an ugly macro (DD4HEP_DEFINE_PARSER_GRAMMAR_TYPE) to define a specialization for its own type. That specialization was actually not a specialization, but the generic implementation itself. So that macro is now gone.
o  Likewise, there is now a generic implementation of the parse function, allowing to remove the part creating a specialization of it in the DD4HEP_DEFINE_PARSER_GRAMMAR_DUMMY_SERIAL macro. Here again the specialization was fake and was a generic implementation, so better to put it straight in the template.

ENDRELEASENOTES


petricm:
Can you elaborate on this a bit, why is this necessary or what is the reason behind this?

MarkusFrankATcernch:
Please check: https://travis-ci.org/AIDASoft/DD4hep/builds/646508140?utm_source=github_status&utm_medium=notification

sponce:
> Can you elaborate on this a bit, why is this necessary or what is the reason behind this?

It is not necessary, but makes the code simpler, more readable and much more usable from user point of view. Main points being : 
- the implementation of the member functions of Grammar are now in the right file, so you find them when you look for them, and you get all you need when including Grammar.h, without having to know that BasicGrammar_inl.h is actually somehow the implementation of Grammar.h
- BasicGrammar now has a generic implementation for its instance method, that was missing before, forcing every end user to use an ugly macro (DD4HEP_DEFINE_PARSER_GRAMMAR_TYPE) to define a specialization for its own type. That specialization was actually not a specialization, but the generic implementation itself. So that macro is now gone.
- likewise, there is now a generic implementation of the parse function, allowing to remove the part creating a specialization of it in the DD4HEP_DEFINE_PARSER_GRAMMAR_DUMMY_SERIAL macro. Here again the specialization was fake and was a generic implementation, so better to put it straight in the template.

I did not go forward further, as this allowed me already to drop all macro calls from my user code, but I expect that one could drop the whole set of macros in BasicGrammar_inl.h

sponce:
> Please check: https://travis-ci.org/AIDASoft/DD4hep/builds/646508140?utm_source=github_status&utm_medium=notification

Damn, forgot one commit when I pushed. Should be fixed now. Sorry about that

sponce:
Just fixed backward compatibility with C++14

petricm:
I have no idea how much work this would be, but would you be willing to go a bit further so that one could drop the whole set of macros in BasicGrammar_inl.h?

sponce:
> I have no idea how much work this would be, but would you be willing to go a bit further so that one could drop the whole set of macros in BasicGrammar_inl.h?

:-) Was my intention ! I've started locally and so far have dropped all the ones in ConditionsInterna.h to start with. Whenever I've checked everything works, I'll open a new merge request

andresailer:
This doesn't work on macOS, e.g.: https://gitlab.cern.ch/CLICdp/DetectorSoftware/DD4hep/-/jobs/7081435

sponce:
> This doesn't work on macOS, e.g.: https://gitlab.cern.ch/CLICdp/DetectorSoftware/DD4hep/-/jobs/7081435

I could test it on macOS now (thanks @petricm !) and I kind of understand the problem. It's very well explained in https://stackoverflow.com/questions/45269457/c-template-symbol-visibility-differences-between-macos-and-linux. I'm on it.

PR #619, opened 2020-02-11
WIP: add plugin for EDM4hep output from DDG4

gaede:
BEGINRELEASENOTES
- add plugin for EDM4hep output from DDG4
    - writes MCParticles, SimTrackerHits and SimCalorimeterHits
    - similar to LCIO output
- can be chosen in ddsim with "*_edm4hep.root" as outputfile
- requires EDM4hep and PODIO to be available
- only build if `DD4HEP_USE_EDM4hep=TRUE` set
- status: *experimental*

ENDRELEASENOTES


DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -346,6 +346,12 @@ def run(self):
       lcOut.EventParametersString, lcOut.EventParametersInt, lcOut.EventParametersFloat = eventPars
       lcOut.RunNumberOffset = self.meta.runNumberOffset if self.meta.runNumberOffset > 0 else 0
       lcOut.EventNumberOffset = self.meta.eventNumberOffset if self.meta.eventNumberOffset > 0 else 0
+    elif self.outputFile.endswith("edm4hep.root"):
+      e4Out = simple.setupEDM4hepOutput('EDM4hepOutput', self.outputFile)
+#      e4Out.RunHeader = self.meta.addParametersToRunHeader(self)
```
petricm: why this comment line?

petricm:
can you please add a test e.g. ddsim gun 1 particle output to edm4hep

DDG4/python/DDSim/DD4hepSimulation.py
gaede: This is still WIP. There are many lines commented out that deal with event and run headers and meta data parameters as these still don't exist in EDM4hep...

vvolkl:
Hi Frank @gaede,

this looks great already! Maybe it is simpler in terms of dependencies to move this plugin to the EDM4hep repository? Here we'd have to provide a preliminary installation somehow, and it looks like we at least have to add the Run/EventHeaders for a first version.

petricm:
what about adding EDM4hep/master to dev3?

petricm:
Has been agreed that this will be moved to the [EDM4hep Repo](https://github.com/HSF/EDM4hep)

vvolkl:
Supersede by: https://github.com/HSF/EDM4hep/pull/43

PR #621, opened 2020-02-14
Improvements: Plugin service, dd4hep_add_dictionary

### LINKED ISSUE #600: Conflicting thispackage.sh scripts for multiple packages
- OS version: ubuntu 18.4
 - Compiler version: GCC 8.3.0
 - DD4hep version: HEAD (982b65499b76600ede1e0e6f79c40fb832967c35)

## What I'm trying to do
I'm developing an event display package based on DD4hep and ROOT Eve7 experimental web event display.

See https://github.com/rete/LCEve

To make it modular I use the plugin mechanism of DD4hep to load LCIO collection converters.
I have successfully:
- Compiled the general library (libLCEve)
- Compiled a DD4hep plugin library (libLCEvePlugins) using the CMake function `add_dd4hep_plugin()`
- Generated my `thislceve.sh` using the CMake function `dd4hep_instantiate_package( lceve )`
- Create an executable `LCEventDisplay` that is looking for a set of plugins using `dd4hep::PluginService::Create()`

Important point: I'm using lcgeo to load our standard ILD geometry for the display (ILD_l5_v02).

## The issue

To run my event display I execute the following:

```shell
source /path/to/lcgeo/bin/thislcgeo.sh
source /path/to/lceve/bin/thislceve.sh
/path/to/lceve/bin/LCEventDisplay [args]
```

The program fails while loading the geometry:

```
Compact          INFO  ++ Converted subdetector:Tube of type DD4hep_Beampipe_o1_v01 
Compact          ERROR FAILED     subdetector:QD0_cryostat of type TubeSupport_o1_v01 
PluginService    ERROR Factory requested: TubeSupport_o1_v01 (N10__cxxabiv120__function_type_infoE) :bad any_cast
PluginService    ERROR Stub is invalid!
Compact          ERROR ++ FAILED    to convert subdetector: QD0_cryostat: Failed to execute subdetector creation plugin. 		No factory with name Create(TubeSupport_o1_v01) for type TubeSupport_o1_v01 found.
		Please check library load path and/or plugin factory name.

**************************************************** 
*  A runtime error has occured :                     
*    Failed to execute subdetector creation plugin. 		No factory with name Create(TubeSupport_o1_v01) for type TubeSupport_o1_v01 found.
		Please check library load path and/or plugin factory name.
*  the program will have to be terminated - sorry.   
**************************************************** 
```
but my plugins from LCEve are found correctly.

It seems that my `thislceve.sh` hides the environement of lcgeo. As a confirmation, from a new shell, if I switch the file source commands, the geometry is correctly loaded but my plugins are not found.

I looked my environment and found that LD_LIBRARY_PATH in the first case (lcgeo then lceve source) is set to the following:
```shell
$ printenv
...
LD_LIBRARY_PATH=/home/ilc/dev/install/lib:/home/ilc/DD4hep/lib::/home/ilc/root/install/lib:/home/ilc/lcgeo/lib:/home/ilc/DD4hep/lib:/home/ilc/python/lib
...
```

DD4hep library path is set twice. `/home/ilc/dev/install/` is the install prefix for LCEve.
Now comes the strange fact: if I remove the last DD4hep library path, i.e:

```shell
export LD_LIBRARY_PATH=/home/ilc/dev/install/lib:/home/ilc/root/install/lib:/home/ilc/lcgeo/lib:/home/ilc/DD4hep/lib:/home/ilc/python/lib
```
then everything works perfectly.

## Summary:

- I'm not sure why it is not working so far... Having twice DD4hep library path should not be a problem, but it is for plugins loading.
- Sourcing multiple this{package}.sh scripts from different packages in the same shell somehow leads to conflicts.

andresailer:
BEGINRELEASENOTES
- Cmake: dd4hep_add_dictionary: directly use command, no longer created bash script to be called. See details of call with `make VERBOSE=1`
- PluginServiceV2: use `boost::split` instead of walking of char arrays, fix bug when two colons are in the environment variable. Fixes #600 
ENDRELEASENOTES

### DIFF
--- a/GaudiPluginService/src/PluginServiceV2.cpp
+++ b/GaudiPluginService/src/PluginServiceV2.cpp
@@ -14,6 +14,8 @@
 #define GAUDI_PLUGIN_SERVICE_V2
 #include <Gaudi/PluginService.h>
 
+#include <boost/algorithm/string.hpp>
+
 #include <dirent.h>
 #include <dlfcn.h>
 
@@ -22,6 +24,7 @@
 #include <iostream>
 #include <memory>
 #include <regex>
+#include <vector>
 
 #include <cxxabi.h>
 #include <sys/stat.h>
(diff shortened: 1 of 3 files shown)


GaudiPluginService/src/PluginServiceV2.cpp
```diff
@@ -140,80 +143,70 @@ namespace Gaudi {
...
-              }
-              fs::path dirName =
-#ifdef USE_BOOST_FILESYSTEM
-                  std::string{search_path.substr( start_pos, end_pos - start_pos )};
-#else
-                  search_path.substr( start_pos, end_pos - start_pos );
-#endif
-              start_pos = end_pos;
+          std::vector<std::string> directories;
+          boost::split(directories, search_path, boost::is_any_of(sep));
```
petricm: Wouldn't it be maybe better to use `strtok` then if you build with C++17 it's independent of boost.

andresailer: But boost is anyway needed for other things.
And it is definitely more readable, and doesn't need one to deal with `char*`
```diff
-          std::vector<std::string> directories;
-          boost::split(directories, search_path, boost::is_any_of(sep));
-
-          for(fs::path dirName: directories) {
+          std::vector<char> cSP(search_path.c_str(), search_path.c_str() + search_path.size() + 1);
+          char* c_dirName = std::strtok(&cSP[0], sep.c_str());
+          while(c_dirName != nullptr) {
+            fs::path dirName(std::string{c_dirName});
+            c_dirName = std::strtok(nullptr, sep.c_str());

```

PR #627, opened 2020-02-20
MakeGaudiMap: use target properties to get the necessary command and …

andresailer:
…library locations

BEGINRELEASENOTES
- MakeGaudiMap: better inference of listcomponents and library location by using target properties
- MakeGauiMap: Detailed information about the command is not only printed when VERBOSE=1

ENDRELEASENOTES


andresailer:
Still have to run on macos, but gitlab is currently down

@MarkusFrankATcernch Can you try with this as well?

andresailer:
Works: https://gitlab.cern.ch/sailer/DD4hep/pipelines/1430325

cmake/MakeGaudiMap.cmake
```diff
@@ -1,22 +1,19 @@
...
   #  Running listcomponents from DD4hep
-  #   
-  FIND_PROGRAM( Gaudi_listcomponents_CMD listcomponents PATHS ${genmap_install_dir}/../bin ${DD4hep_DIR}/bin
-    NO_DEFAULT_PATH NO_CMAKE_PATH)
-  GET_FILENAME_COMPONENT(GAUDI_LISTCOMP_INSTALL ${Gaudi_listcomponents_CMD} DIRECTORY)
-  ###MESSAGE( STATUS " *** MakeRootMap ${genmap_install_dir}/../bin $ENV{DD4hep_DIR}/bin ${DD4hep_DIR}" )
-  MESSAGE( STATUS " *** MakeGaudiMap.cmake run command : ${Gaudi_listcomponents_CMD} -o ${rootmapfile} ${libname}
-                 WORKING_DIRECTORY ${genmap_install_dir} "
+  #
+  MESSAGE( STATUS " *** MakeGaudiMap.cmake run command : ${DD4HEP_LISTCOMPONENTS_CMD} -o ${rootmapfile} ${libname}
```
petricm: It might be better to convert these two status messages to `dd4hep_debug`?

andresailer: This runs in an independent cmake run.
Thus there is no dd4hep_debug

Issue #625, opened 2020-02-19
Build failure with ROOT 6.22 new python structure


lastCent:
DD4hep fails to build in our nightly builds, probably because of the way in which multiple Python versions now are organized in the lib/ directory, and the splitting of PyROOT from TPython. ROOT 6.20 changelog: https://root.cern/doc/v620/release-notes.html
 - OS version: e.g. SLC6
   - SLC6, Centos7 
 - Compiler version: e.g GCC 6.2
   - gcc7, gcc8, gcc9, clang8
 - DD4hep version: tag or commit ID, or GitHub branch
   - Master, commit: 84cf31e748ee44b8453818823189377d92a22a7a
 - Reproduced by: exact steps to reproduce the problem: checkout, setup environment (Geant, ROOT versions, iLCSoft release), cmake, build, run...
    - The build was run with [LCGCmake](https://gitlab.cern.ch/sft/lcgcmake/) as part of the LCG stacks. 
    - Geant4 version: 10.6
    - ROOT version: Master (1796471234a85b7d6272e371d586b58a7e1287e7)
 - Input: link to input files if applicable
 - Output: full build and run log output
   - Current build logs are available in the dev3 and dev3python3 section [here](https://cdash.cern.ch/index.php?project=LCGSoft). 
   - [logs.tar.gz](https://github.com/AIDASoft/DD4hep/files/4225196/logs.tar.gz)

 - Goal: A short description of what you are trying to achieve
   - Build DD4hep as part of our nightly builds.

petricm:
Interesting, we are testing in our CI against dev3 and dev4 and we did not observe this. How can we test this particular ROOT configuration of yours?

petricm:
But just for clarification dev3 has ROOT built with both python2 and 3 or is there be dev3(py3) and dev3python3(py3)?

andresailer:
@lastCent I couldn't find the cmake command and output in the logs or the cdash website

petricm:
So this is related to this

> Dear All,
> 
> As discussed in the ROOT planning meeting of last week (https://indico.cern.ch/event/884189/), we are going to do the switch of experimental PyROOT to default in master today. Therefore, nightly builds with ROOT master will start running with experimental PyROOT tonight.
> Please communicate to us any issues you may observe and we will work on them right away.
> 
> Best,
> 
> Enric

petricm:
@lastCent but this affects only root master and not the 6-20 branch, so this change is for ROOT 6.22?

[...]

PR #631, opened 2020-02-21
DDG4Python: only build things if python and pyroot are found

andresailer:
Fixes (at least part of) #625 

While the add_dictionary command was called, the rootcling command behind it was never executed because the library that used the cxx was not created when PyROOT is not found, still the install command for the pcm file was there leading to the error described in #625 

@MarkusFrankATcernch I am not sure if having the pcms without the libraries makes sense so I put it all in one if, alternatively we could force the creation of the cxx and pcm file even if they are not needed.

BEGINRELEASENOTES
- Only build DDG4 Python dependent libraries and pcms if Python and PyROOT are found

ENDRELEASENOTES

### DIFF
--- a/DDG4/CMakeLists.txt
+++ b/DDG4/CMakeLists.txt
@@ -50,7 +50,7 @@ dd4hep_add_plugin(DDG4Plugins
 #---------------------------  Plugin library for the simulation framework  ---------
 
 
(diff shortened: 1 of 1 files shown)


andresailer:
Merging now so maybe the lcg nightly builds can run over the weekend

PR #633, opened 2020-02-21
Add CAD interface to load shapes from CAD files 

MarkusFrankATcernch:
BEGINRELEASENOTES
- Add CAD interface to load shapes from CAD files
 The basic stuff is done. It still will have to be figured out how to best embed the volume/shape loading
 from CAD files into the existing infrastructure. See for details:
 https://indico.cern.ch/event/885083/contributions/3758180/attachments/1990793/3318984/2020-02-20-DD4hep-Tessellated-Shapes.pdf
In any case tessellated shapes shall only be available with ROOT 6.21 onwards. 
- DDCAD used the assimp librray from github to interprete the CAD files.
- Simplify grammar instantiation. 
-- For grammars to be parsed with boost::spirit simply include
#include "DD4hep/detail/Grammar_parsed.h"
--For Grammars which should not be parsed with boost::spirit include
#include "DD4hep/detail/Grammar_unparsed.h"
To instantiate the code and to register the instantiation call:
template <> dd4hep::Grammar<my-class>;
If the Grammar is supposed to be registered call instead/in addition:
static auto s_registry = GrammarRegistry::pre_note<my-class>();
All other macros are gone.
ENDRELEASENOTES


DDCAD/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+if(${ROOT_VERSION} VERSION_LESS 6.21.01)
+  return()
+endif()
+#
+# Open Asset Importer Library
+#
+find_package(assimp QUIET)
+if(TARGET assimp::assimp)
+    MESSAGE(STATUS "+++> Use existing assimp implementation")    
+else()
```
andresailer: I think we should add a flag to enable or disable the automatic building of assimp, and disable this by default.
If neither assimp is found, nor the flag is enabled we just don't build DDCAD.

petricm:
There are a lot of binary files in the examples added are these really necessary?

petricm:
It might be best to put a submodule to the [assimp](https://github.com/assimp/assimp/tree/master/test/models) repo subfolder `test/models`

andresailer:
To really get rid of the cad example files from the history, we have to squash the commits.

PR #636, opened 2020-02-25
HepMC3Reader: initial prototype

andresailer:
BEGINRELEASENOTES
- DDG4: add reader for HepMC3 files
- DDG4: add generic EventParameters class to pass event level parameters from different input sources to different output sources, supercedes the LCIOEventParameters class

ENDRELEASENOTES


andresailer:
This hepmc3 reader should also be able to read the old hepmc2 ascii files (http://hepmc.web.cern.ch/hepmc/classHepMC3_1_1ReaderAsciiHepMC2.html) so we could drop that now

PR #641, opened 2020-03-02
DDSim: bring back the visible printout of dumpSteeringFile

### LINKED ISSUE #643: Segmentation fault in DDSim when using position
When running `ddsim` with a `position` or `direction` option in gun it generated a `Segmentation fault`

```
ddsim --compactFile DDDetectors/compact/SiD.xml --runType=batch -G -N=10  --gun.position "0.0 0.0 0.0" --outputFile=testCLIC_o3_v14.slcio
```

Checking back with old releases this problem was **not** present in `2019-09-04` but was present already in `2020-02-07`

andresailer:
BEGINRELEASENOTES
- DDSim: make the output of `ddsim --dumpSteeringFile` visible again
- DDSim: fix exception in ConfigHelper.printOptions, called by `--dumpParameters` option
- DDSim: fix parsing of vector command line parameters, gun.position, gun.direction etc.
- DDSim: better testing of command line parameters

ENDRELEASENOTES

When moving to logger, the logger was only configured after the parsing or parameters, but actually just `print`ing the steering file is better, because then we don't get the logging header, which would invalidate the steering file."
Resolves #643

### DIFF
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -6,7 +6,7 @@ Based on M. Frank and F. Gaede runSim.py
    @version 0.1
 
 """
-from __future__ import absolute_import, unicode_literals, division
+from __future__ import absolute_import, unicode_literals, division, print_function
 __RCSID__ = "$Id$"
 import sys
 import os
@@ -628,7 +628,8 @@ SIM = DD4hepSimulation()
         else:
           steeringFileBase += "SIM.%s = %s" % (parName, str(parameter))
         steeringFileBase += "\n"
-    logger.info("%s", steeringFileBase)
+    for line in steeringFileBase.splitlines():
+      print(line)
 
   def _consistencyChecks(self):
     """Check if the requested setup makes sense, or if there is something preventing it from working correctly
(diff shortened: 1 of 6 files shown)


DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -628,7 +628,9 @@ def __printSteeringFile(self, parser):
         else:
           steeringFileBase += "SIM.%s = %s" % (parName, str(parameter))
         steeringFileBase += "\n"
-    logger.info("%s", steeringFileBase)
+    for line in steeringFileBase.splitlines():
+      print(line)
```
petricm: You have to add `from __future__ import print_function` since there is no other usage of `print` in this file, the import is not added at the beginning.

DDG4/python/DDSim/Helper/ConfigHelper.py
```diff
@@ -53,10 +53,10 @@ def __repr__(self):
     return self.printOptions()
 
   def printOptions(self):
-    """print all paramters"""
+    """print all parameters"""
     options = []
     for opt, val in six.iteritems(self.getOptions()):
-      options.append("\n\t'%s': '%s'" % (opt, val[0]))
+      options.append("\n\t'%s': '%s'" % (opt, val['default']))
```
petricm: This is actually a bug fix, since before you probably got `KeyError`

andresailer: mentioned this in the release notes after creation of PR

DDG4/python/DDSim/DD4hepSimulation.py
andresailer: added the import

PR #644, opened 2020-03-02
Make DDSim test more realistic with more options

petricm:
Make DDSim test more realistic with more options, this also shown now better the bug from #643


andresailer:
Cherry picked the commit into #641

PR #654, opened 2020-03-25
Update RPATH settings

petricm:
BEGINRELEASENOTES
- Rewrite rpath usage on macOS and make it relocatable with help of `@rpath`
  - itroduce new option `DD4HEP_SET_RPATH` default `ON`
- adopt `thisdd4hep.sh` and other `.sh` scripts to work with zsh (new default for macOS 10.15)
- Drop `MakeGaudiMap.cmake` and call directly `listcomponents`
  - fix env settings for macOS
- Drop deprecated `-std=` flag in dictionary creation

ENDRELEASENOTES


petricm:
the settings follow the root settings [here](https://github.com/root-project/root/blob/master/cmake/modules/RootBuildOptions.cmake#L401-L422)

cmake/thisdd4hep.sh
```diff
@@ -61,7 +65,12 @@ dd4hep_add_library_path()    {
 }
 #-----------------------------------------------------------------------------
 #
-dd4hep_parse_this ${BASH_ARGV[0]} DD4hep;
+SOURCE=${BASH_ARGV[0]}
+if [ "x$SOURCE" = "x" ]; then
+    SOURCE=${(%):-%N} # for zsh
+fi
+
+dd4hep_parse_this $SOURCE  @PackageName@;
```
andresailer, line 73:
```suggestion
dd4hep_parse_this $SOURCE DD4hep;
```
As it was before

petricm: but that was probably wrong, it should at best be not capitalized just `dd4hep`

PR #659, opened 2020-04-09
Support for conditions with non-default ctor

MarkusFrankATcernch:
BEGINRELEASENOTES
 - Allow condition objects, which do not provide a default constructor. If such objects should be managed, these cannot be saved with ROOT - these objects may only live in memory. ROOT requires a default constructor.

ENDRELEASENOTES


MarkusFrankATcernch:
I have added a JIRA issue, but relaxed the importance, since it is the ROOT HEAD
https://sft.its.cern.ch/jira/browse/ROOT-10669

Issue #671, opened 2020-05-08
ROOT 6.22 compatibility issue


petricm:
Currently there is this blocking issue upstream

https://sft.its.cern.ch/jira/browse/ROOT-10669

petricm:
One issue with ROOT 6.22 solved but now another was uncovered. The multi collection tests fail

```
        127 - t_ClientTests_g4material_scan_MultiCollections_LONGTEST (Failed)
	158 - t_ClientTests_sim_MultiCollections (Failed)
	160 - t_ClientTests_sim_MultiSegmentCollections (Failed)
``` 

with the trace

```py
Traceback (most recent call last):
  File "/DD4hep/examples/examples/ClientTests/scripts/MultiCollections.py", line 71, in <module>
    run()
  File "/DD4hep/examples/examples/ClientTests/scripts/MultiCollections.py", line 53, in run
    seq, act = geant4.setupCalorimeter('TestCal')
  File "/DD4hep/lib/python2.7/site-packages/DDG4.py", line 630, in setupCalorimeter
    return self.setupDetector(name, type, collections)
  File "/DD4hep/lib/python2.7/site-packages/DDG4.py", line 612, in setupDetector
    act = SensitiveAction(self.kernel(), sensitive_type + '/' + coll_nam + 'Handler', name)
NotImplementedError
```

petricm:
So the simple reproducer is
```py
import DDG4

kernel = DDG4.Kernel()
geometry = "file:" + "__PATH__/DD4hep/examples/examples/ClientTests/compact/MultiCollections.xml"
kernel.loadGeometry(str(geometry))
geant4 = DDG4.Geant4(kernel)

seq, act = geant4.setupCalorimeter('TestCal')
```
and the last line breaks.
But if you expand the `setupCalorimeter` call it is
```py
description = kernel.detectorDescription()
sd = description.sensitiveDetector(str('TestCal'))
type = 'Geant4CalorimeterAction'
```
followed by a call to `return self.setupDetector(name, type, collections)` which is
```py
seq = DDG4.SensitiveSequence(kernel, 'Geant4SensDetActionSequence/' + 'TestCal')
sd = description.sensitiveDetector(str('TestCal'))
coll_nam = 'TestCallInnerLayerHits' # or TestCallMiddleLayerHits or TestCallOuterLayerHits
act = DDG4.SensitiveAction(kernel, sensitive_type + '/' + coll_nam + 'Handler', 'TestCal')
```
[...]

PR #672, opened 2020-05-29
Dictionaries: use unique include guards

andresailer:
As advised by https://sft.its.cern.ch/jira/browse/ROOT-10669

BEGINRELEASENOTES

- Use unique include guards

ENDRELEASENOTES

Partially addresses #671


petricm:
Did you check all include guards?

andresailer:
Yes, there are a few more suspect cases and some that might be intentional

PR #673, opened 2020-06-05
Fix guard and consequences and casting to string

### LINKED ISSUE #671: ROOT 6.22 compatibility issue

petricm:
`DDCore/include/DD4hep/BitField64.h` has an include
```cpp
#include "DDSegmentation/BitField64.h"
```
which has an include
```cpp
#include "BitFieldCoder.h"
```
which probably should be the one from `DDSegmentation`?

Because of the re-usage of the include guard (and of course `using namespace`) an ambiguity in `BitField64` was shadowed. 

BEGINRELEASENOTES
- Use unique include guard in `BitFieldCoder.h`
  - fix ambiguity in test between `BitField64` in LCIO and DD4hep
- Cast to string `coll_nam` in `DDG4.py` before using further, otherwise string concatenation does not work

ENDRELEASENOTES
Resolves #671

### DIFF
--- a/DDCore/include/DD4hep/BitFieldCoder.h
+++ b/DDCore/include/DD4hep/BitFieldCoder.h
@@ -10,8 +10,8 @@
 // Author     : M.Frank
 //
 //==========================================================================
-#ifndef DD4HEP_DDCORE_BITFIELD64_H
-#define DD4HEP_DDCORE_BITFIELD64_H
+#ifndef DD4HEP_DDCORE_BITFIELDCODER_H
+#define DD4HEP_DDCORE_BITFIELDCODER_H
 
 // Framework include files
 #include "DDSegmentation/BitFieldCoder.h"
@@ -27,4 +27,4 @@ namespace dd4hep {
 
   }       /* End namespace detail           */
 }         /* End namespace dd4hep             */
-#endif    /* DD4HEP_DDCORE_BITFIELD64_H     */
+#endif    /* DD4HEP_DDCORE_BITFIELDCODER_H     */
--- a/DDCore/include/DDSegmentation/BitField64.h
+++ b/DDCore/include/DDSegmentation/BitField64.h
@@ -19,7 +19,7 @@
 #include <map>
 #include <sstream>
 
-#include "BitFieldCoder.h"
+#include "DDSegmentation/BitFieldCoder.h"
 
 namespace dd4hep {
   
--- a/DDG4/python/DDG4.py
+++ b/DDG4/python/DDG4.py
@@ -599,16 +599,16 @@ class Geant4:
         params = {}
         if isinstance(coll, tuple) or isinstance(coll, list):
           if len(coll) > 2:
-            coll_nam = coll[0]
+            coll_nam = str(coll[0])
             sensitive_type = coll[1]
-            params = coll[2]
+            params = str(coll[2])
           elif len(coll) > 1:
-            coll_nam = coll[0]
+            coll_nam = str(coll[0])
             sensitive_type = coll[1]
           else:
-            coll_nam = coll[0]
+            coll_nam = str(coll[0])
         else:
-          coll_nam = coll
+          coll_nam = str(coll)
         act = SensitiveAction(self.kernel(), sensitive_type + '/' + coll_nam + 'Handler', name)
         act.CollectionName = coll_nam
         for parameter, value in six.iteritems(params):
--- a/UtilityApps/src/test_surfaces.cpp
+++ b/UtilityApps/src/test_surfaces.cpp
@@ -19,7 +19,6 @@
 #include "DD4hep/DDTest.h"
 
 #include "DD4hep/DD4hepUnits.h"
-#include "DD4hep/BitField64.h"
 
 #include "lcio.h"
 #include "IO/LCReader.h"
@@ -35,7 +34,6 @@ using namespace std ;
 using namespace dd4hep ;
 using namespace dd4hep::detail;
 using namespace dd4hep::rec ;
-using namespace lcio;
 
 
 static DDTest test( "surfaces" ) ; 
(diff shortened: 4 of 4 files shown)


andresailer:
What was the type before casting to string that caused the exception?

petricm:
Then it fails `type` returns `<class cppyy.gbl.std.string at 0xafdc2c0>`

petricm:
And in LCG_97 with the old version the same variable is `<type 'str'>`

petricm:
In the end it's a pure `cppyy` "problem"
```py
>>> import cppyy
>>> s1 = cppyy.gbl.std.string("test")
>>> s2 = "test"
>>> s+s2
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NotImplementedError
```
this is a bit annoying since it behave differently then other builtins. From the `cppyy` manual:
> For example, a C++ float is returned as a Python float, which is in fact a C++ double. If sizes allow, conversions are automatic. For example, a C++ unsigned int becomes a Python2 long or Python3 int, but unsigned-ness is still honored:

andresailer:
Yes, and the manual says string comparison can be done, but the + operator just gives NotImplementedError exception.

MarkusFrankATcernch:
So whenever a C++ object returns std::string one has to do:
ret_str = str(cxx_object.function())

This sort of reminds of something couple a months ago....

andresailer:
Do you mean the conversion from unicode to str before passing to c++ ? Or something else?

MarkusFrankATcernch:
Oh. Yes. At the end this is very similar. Replace unicode by std::string....

andresailer:
Same solution, different cause though.

The status from travis did not get reported, but it is all green (including dev3) for the last commit
https://travis-ci.org/github/AIDASoft/DD4hep/builds/695102829

MarkusFrankATcernch:
The actual cause is that we have to solve here a conversion matrix of C++ strings (char*, std::string, (less: std::wstring)) and python strings (str, unicode). This matrix is less strictly defined due to the rise of unicode in python.

andresailer:
and now also the cppyy.gbl.std.string type in python!

MarkusFrankATcernch:
Yes. 
There are no rules, which could be programmed into pyROOT which automatic type conversions to apply between the types. 
I sort of think before the default was: if there was no direct conversion it used __str__ and __repr__.

petricm:
So I had a small chat with Enric and they were not aware of this behaviour that you cannot anymore concatenate `std::string` with `str` in PyROOT and they have to see if there is a reason for this or it is just a missing functionality. Personally, I doubt that this is missing functionality, most likely cppyy does not overload the + operator  because the `type` of the result is ambiguous (`str` vs `cppyy.gbl.std.string`). Anyway we concluded to open an [issue](https://sft.its.cern.ch/jira/browse/ROOT-10830) and they will decide if this is a bug or just new not backwards compatible behaviour.

wlav:
@MarkusFrankATcernch: I don't know about PyROOT, but you can write custom executors for cppyy as well as type contractions for the wrapper generators (that last one highly unlikely in PyROOT, as they don't keep up with patches or bug reports for ROOT/meta).

Anyway, this particular case of `operator+` for `std::string`/`str` has nothing to do with that and is trivial to solve in the operator lookup, so already fixed in (the cppyy) repo ...

petricm:
Thank you @wlav. You refer to [this commit](https://bitbucket.org/wlav/cppyy-backend/commits/ea594f920a9db9f535b6ecfcbd0403210b19aaee)? So with cppyy 1.12.1 this should be solved.

wlav:
Yes. I'm writing tests atm. for the other conversions (unicode, wstring) to fill out the combinations.

petricm:
@wlav is this behaviour below intentional
```py
>>> import cppyy
>>> cppyy.cppdef('std::vector<std::string> foo() { return std::vector<std::string>{"foo","bar"};}')
True
>>> colls = cppyy.gbl.foo()
>>> type(colls)
<class cppyy.gbl.std.vector<string> at 0x4ad8220>
>>> type(colls[0])
<type 'str'>
>>> for coll in colls:
...   print type(coll)
... 
<class cppyy.gbl.std.string at 0x4cd41b0>
<class cppyy.gbl.std.string at 0x4cd41b0>
```
The type when I loop over the verctor or when I access an element is different.

wlav:
History... Returning `std::string` has historically been converted to Python `str`, but never made much sense (e.g. `str` is immutable, but `std::string` is not) and is clearly wrong in Python3 where `str` is `unicode`. It just hasn't led to problems (other than inefficiencies) so far, so I left it. But any new features (such as the fast `vector` iteration in this case) working with `std::string` should just keep it as such and `std::string` should be able to drop in everywhere `str` can. Just that at the time when the choice was made (what, back in 2003?) to do a conversion, `std::string` was far from a drop-in, CINT didn't properly support return-by-value, CINT rules for `const&` returns were murky, etc., etc. These are no longer a problem.

wlav:
I got the whole set of valid combinations to work (`str`, `unicode`, `std::string`, `std::wstring`, and both ways). The ones that do _not_ work are only `std::string`+`std::wstring` and vice versa, b/c C++ does not support them.

MarkusFrankATcernch:
@wlav This is fantastic! Thanks a lot.
This shall solve many compatibility issues.

petricm:
Thank you, this is great news. Just being curious what is a type of the result if you do `str` + `std::string` or other combinations?

wlav:
`str` or `unicode`, depending on the types used, this b/c returning functions still convert to Python types (this hasn't changed).

The details of how it works are as follows: `operator+` in C++ is a globally overloaded template. It thus needs to be found and possibly instantiated before use. Global operators are not searched for on proxy class creation b/c it's expensive and anyway, global operators can be loaded/defined after declaring and binding the class. So, they're looked for only on first use.

At that point, `cppyy` needs to figure out the types (and full signature) to use for instantiation. What went wrong in the above example, is that the types used were `std::string` and `str`, so the template instantiation fails, as C++ does not know `str`. The fix is to check for `str` beforehand and changing the Python string type to match the C++ one if that is a string type, or fall back to `std::string` otherwise. The normal type converters can handle both `str` and `unicode` (the latter may encode/decode if need be). The return type is then the return type of `operator+`, which is subsequently converted to a Python type. (This is a good example of why that conversion may well be a bad idea, but as said above: history.)

For example `std::string` + `unicode` would return `str`, b/c the templated `operator+` is triggered by the `std::string` part. In theory, `unicode` return might have been better, but that would require an `operator+` of `std::string` and `std::wstring` returning `std::wstring` and C++ does not support this. May be a future change to have the `std::wstring` converter accept `std::string`, but if you really care about `unicode`, it's not my task to handle your `std::string`s. :)

PR #675, opened 2020-06-11
[WIP] [Test] Migrating code from CMSSW - Filter and SpecParRegistry

ianna:
BEGINRELEASENOTES
- Filter and SpecParRegistry classes from CMSSW port test

Note: tbb is used

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/Filter.h
+++ b/DDCore/include/DD4hep/Filter.h
@@ -54,13 +54,13 @@ namespace dd4hep {
       /* }); */
     }
 
-    void print(const std::unique_ptr<Filter>& filter) const {
+    //    void print(const std::unique_ptr<Filter>& filter) const {
       /* edm::LogVerbatim("Geometry").log([&](auto& log) { */
       /*   for (const auto& it : filter->skeys) { */
       /*     log << it << ", "; */
       /*   } */
       /* }); */
-    }
+    //}
 
     std::vector<std::string_view> skeys;
     std::vector<std::regex> keys;
--- a/DDCore/include/DD4hep/Filter.h
+++ b/DDCore/include/DD4hep/Filter.h
@@ -1,0 +1,84 @@
+//==========================================================================
+//  AIDA Detector description implementation 
+//--------------------------------------------------------------------------
+// Copyright (C) Organisation europeenne pour la Recherche nucleaire (CERN)
+// All rights reserved.
+//
+// For the licensing terms see $DD4hepINSTALL/LICENSE.
+// For the list of contributors see $DD4hepINSTALL/doc/CREDITS.
+//
+// Author     : Ianna Osborne
+//
+//==========================================================================
+#ifndef DD4HEP_DDCORE_FILTER_H
+#define DD4HEP_DDCORE_FILTER_H
+
+// -*- C++ -*-
+//
+// Class:      Filter
+//
+/**\class Filter
+
+ Description: Filter list
+
+ Implementation:
+     Filter criteria is defined in XML
+*/
+//
+// Original Author:  Ianna Osborne
+//         Created:  Tue, 12 Mar 2019 09:51:33 CET
+//
+//
+
+#include <memory>
+#include <vector>
+#include <regex>
+
+namespace dd4hep {
+  struct SpecPar;
+
+  struct Filter {
+    void print() const {
+      /* edm::LogVerbatim("Geometry").log([&](auto& log) { */
+      /*   for (const auto& it : skeys) { */
+      /*     log << it << ", "; */
+      /*   } */
+      /*   if (next) { */
+      /*     log << "Next:\n"; */
+      /*     print(next); */
+      /*   } */
+      /*   if (up) { */
+      /*     log << "Up:\n"; */
+      /*     up->print(); */
+      /*   } */
+      /* }); */
+    }
+
+    void print(const std::unique_ptr<Filter>& filter) const {
+      /* edm::LogVerbatim("Geometry").log([&](auto& log) { */
+      /*   for (const auto& it : filter->skeys) { */
+      /*     log << it << ", "; */
+      /*   } */
+      /* }); */
+    }
+
+    std::vector<std::string_view> skeys;
+    std::vector<std::regex> keys;
+    std::unique_ptr<Filter> next;
+    struct Filter* up;
+    const SpecPar* spec = nullptr;
+  };
+
+  namespace dd {
+    bool accepted(std::vector<std::regex> const&, std::string_view);
+    bool isRegex(std::string_view);
+    bool isMatch(std::string_view, std::string_view);
+    bool compareEqual(std::string_view, std::string_view);
+    bool compareEqual(std::string_view, std::regex);
+    std::string_view realTopName(std::string_view);
+    std::vector<std::string_view> split(std::string_view, const char*);
+    std::string_view noNamespace(std::string_view);
+  }  // namespace dd
+}  // namespace dd4hep
+
+#endif
--- a/DDCore/src/Filter.cpp
+++ b/DDCore/src/Filter.cpp
@@ -9,7 +9,7 @@ namespace cms {
   namespace dd {
 
     bool isMatch(string_view node, string_view name) {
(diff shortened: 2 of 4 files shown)


ianna:
@MarkusFrankATcernch - the build failures are due to `tbb`

MarkusFrankATcernch:
Hi Yana,
I had a look at your changes.
I expected somewhere in the DDCore CMakeLists.txt a reference to TBB to resolve
compilation and linker issues. I did not find any. I somehow suppose it compiles because the 
TBB headers are either in the system includes (or the LCG view), but then it comes to linking you need a real link dependency. What you see are all linker errors.
Is this analysis sort of correct ?

There are then 2 solutions:
1) We need a way to find tbb (either from LCG or the system) and if there is none,
we have to perform a dependent compilation.
2) We always require tbb and refuse to build if not present. tbb would then be at the same level as ROOT.

What I do not exactly know is how to check in cmake if tbb is present on the building system.

petricm:
You can use
```cmake
find_package(CMakeTools)
UseCMakeTools()
...
find_package(TBB)
```
but this is LCG specific.

MarkusFrankATcernch:
Would something like this work?

```
find_package(TBB)

if(TARGET TBB::TBB)
  file(GLOB DDCore_TBB_SOURCES src/Filter/*.cpp)
endif()
```
One also would somehow have to do a similar trick to the dictionary generation.

petricm:
The problem is that there is no FindTBB in cmake and the existence of a `TBBConfig.cmake` is system dependent.

andresailer:
Not sure if this will work like this and some details and typos to clean up, some logging to add, probably some tricky cmake issues I didn't consider, and then one also needs to propagate the TBB dependency to users of DDCore, maybe add a flag to require TBB and fail if it isn't found instead of moving on and not compiling the relevant files.

```cmake
# find TBB either by TBBConfig or FindTBB file from CMakeTools
# Populate TBB_TARGET argument with the TBB Library target
function find_tbb(TBB_TARGET)
  find_package(TBB QUIET MODULE) # should require TBBConfig.cmake
  if(TARGET TBB::TBB)
    set(${TBB_TARGET} TBB::TBB PARENT_SCOPE)
    return()
  endif()
  find_package(CMakeTools)
  if(NOT CMakeTools_DIR) # we didn't find cmake tools
    return()
  endif()
  UseCMakeTools() # should not leak the CMAKE_MODULE_PATH outside this function scope
  find_package(TBB QUIET) # should use FindTBB from CMakeTools
  if(NOT TBB_FOUND)
    return()
  endif()
  ADD_LIBRARY(TBB_INTERFACE INTERFACE)
  ... # set the library and include directories
  SET(${TBB_TARGET} TBB_INTERFACE PARENT_SCOPE)
endfunction()
```

DDCore/include/DD4hep/Filter.h
```diff
@@ -0,0 +1,84 @@
...
+
+    //    void print(const std::unique_ptr<Filter>& filter) const {
+      /* edm::LogVerbatim("Geometry").log([&](auto& log) { */
+      /*   for (const auto& it : filter->skeys) { */
+      /*     log << it << ", "; */
+      /*   } */
+      /* }); */
+    //}
+
+    std::vector<std::string_view> skeys;
```
andresailer, line 65: this is c++17 only

MarkusFrankATcernch:
@andresailer We could try it out for DDDigi.
If it works there, we could see how to do all the conditional stuff in DDCore....

petricm:
@mrodozov what do you use to find TBB in CMS?

mrodozov:
We use SCRAM which setups xml files that are exporting the paths to any externals (including TBB)
CMSSW is not using cmake modules as in here:
https://github.com/cms-sw/cmsdist/pull/5743/files#diff-e983be576e1f6ccc56bc969745dfb70f
where we had to switch to cmake for protobuf as onnx started using cmake modules to find it

What we do in CMS is that for a package that is using TBB we put it in a BuildFile like:
https://github.com/cms-sw/cmssw/blob/afe34f626e8e6fd78f2eb387bf22a28898a24dc3/DetectorDescription/Core/BuildFile.xml#L7

and SCRAM will setup the paths to it. I have to see what is missing and if it's some piece of TBB that is not setup we have to add it from this file:
https://github.com/cms-sw/cmsdist/blob/8a09a8ae19af9e4409d390f6a6c4058fb9aa8605/tbb-toolfile.spec

thats in case some cmssw piece doesn't compile
----
in case DD4Hep doesn't compile (this PR)

We don't use LCG releases but build our own TBB and setup it.
If this concerns building this PR, I can see that DD4hep in our specs doesn't require 
TBB directly https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_2_X/master/dd4hep.spec#L10
, although ROOT does, and since it does it indirectly should bring it . 

I'll build this PR and in the meanwhile could you post here what is wrong with the build ?

andresailer:
@mrodozov Do you have a TBBConfig.cmake file as part of your TBB installation? Or do you include the CMakeTools (https://gitlab.cern.ch/sft/cmaketools) ?
DD4hep is build with cmake also in cmssw (https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_2_X/master/dd4hep.spec), right?

mrodozov:
No I see that we don't have the TBBConfig.cmake in the installation:
/cvmfs/cms.cern.ch/slc7_amd64_gcc820/external/tbb/2019_U9-bcolbf/
and we don't use this cmaketools.
DD4Hep is using cmake to build, yes, but not TBB which may explain why the file is missing.

andresailer:
I think you could create the TBBConfig.cmake file after the fact with
https://github.com/oneapi-src/oneTBB/blob/tbb_2020/cmake/tbb_config_generator.cmake#L16

mrodozov:
We are not using cmake to build TBB but we can start doing it
https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_2_X/master/tbb.spec
as we did for protobuf (for the same reason)
Ok I'll make it build with cmake. Hope it wont take long

andresailer:
As far as I understand you don't have to build with cmake, just call the tbb_config_generator.cmake with the arguments set.

mrodozov:
in the installation there is a cmake folder now containing
TBBConfig.cmake  TBBConfigVersion.cmake

andresailer:
That's great! I think in the LCG installation this will also come (SPI-1600 was re-opened), so we can just assume that TBBConfig.cmake is provided when TBB should be used and not bother creating something ourselves.

smuzaffar:
DD4Hep deppends on root ... right? So root has FindTBB.cmake ( https://github.com/root-project/root/blob/6fdd0c55f457ad795ba9a40def9c251afdc1981d/cmake/modules/FindTBB.cmake ) , can not you make use of that?

smuzaffar:
Currently we are not building TBB using cmake ( I think there were some issues with cmake build) but of course we can keep the TBBConfig.cmake in our distribution if that helps

andresailer:
Just for the record: As far as I can tell we cannot use FindTBB from root, because it doesn't get installed.
e.g., `find /cvmfs/cms.cern.ch/slc7_amd64_gcc820/lcg/root/6.20.01/ -name FindTBB.cmake` yields no results.
The FindTBB in the lcg installation comes from Gaudi, Geant4, or CMakeTools.

But regardless, using the TBBConfig is the most useful and standard solution because of the imported targets we can use then, and it doesn't cause a lot of overhead.

petricm:
The `TBBConfig.cmake` is already deployed in the LCG nightlies (dev3 and dev4) which corresponds to out last two pipelines in the CI.

DDCore/src/Filter.cpp
```diff
@@ -0,0 +1,66 @@
+#include "DD4hep/Filter.h"
+#include <functional>
+#include <regex>
+#include <iostream>
+
+using namespace std;
+
+namespace cms {
```
andresailer, line 8:
```suggestion
namespace dd4hep {
```

petricm:
Superseded by #682

PR #677, opened 2020-06-12
Fixed the CLHEP::mm to dd4hep::mm convert problem.

mirguest:
BEGINRELEASENOTES
- Fixed the CLHEP::mm to dd4hep::mm conversion problem in DDG4/src/Geant4SensitiveDetector.cpp

ENDRELEASENOTES

### DIFF
--- a/DDG4/src/Geant4SensitiveDetector.cpp
+++ b/DDG4/src/Geant4SensitiveDetector.cpp
@@ -30,7 +30,7 @@
 #include "DD4hep/DD4hepUnits.h"
 #include "CLHEP/Units/SystemOfUnits.h"
 
-static const double MM_2_CM = (CLHEP::millimeter/dd4hep::millimeter);
+static const double MM_2_CM = (dd4hep::millimeter/CLHEP::millimeter);
 
 #define DEBUG 0
 


DDG4/src/Geant4SensitiveDetector.cpp
```diff
@@ -30,7 +30,7 @@
 #include "DD4hep/DD4hepUnits.h"
 #include "CLHEP/Units/SystemOfUnits.h"
 
-static const double MM_2_CM = (CLHEP::millimeter/dd4hep::millimeter);
+static const double MM_2_CM = (dd4hep::millimeter/CLHEP::millimeter);
```
petricm, line 33: Shouldn't this be inside
```cpp
#ifdef HAVE_GEANT4_UNITS
#define MM_2_CM 1.0
#else
#define MM_2_CM 0.1
#endif
```

mirguest: Hi @petricm ,
I am Tao Lin from IHEP, working for CEPC. The current SD used in CEPC is still the legacy one. 

The problem we found is that in following code
```c++
    G4ThreeVector global = 0.5 * ( h.prePosG4()+h.postPosG4());
    G4ThreeVector local  = h.preTouchable()->GetHistory()->GetTopTransform().TransformPoint(global);
    Position loc(local.x()*MM_2_CM, local.y()*MM_2_CM, local.z()*MM_2_CM);
    Position glob(global.x()*MM_2_CM, global.y()*MM_2_CM, global.z()*MM_2_CM);
    VolumeID cID = seg.cellID(loc,glob,volID);
```
The units of variables `global` and `local` are `CLHEP::mm`. In order to convert to `dd4hep::mm`, we need to divide the `CLHEP::mm` first, then multiply `dd4hep::mm`.

Thank you!

Tao Lin

MarkusFrankATcernch:
Hi Tao,

Firstly, thanks for spotting this! Secondly, please move away from legacy sensitive detectors soon. They are only there because we thought some FCC guys are still using them. Otherwise this is entirely unmaintained code. As soon as you do no longer use them, please make a pull request deleting this file and the corresponding header or alternatively notify me (or Marko).

Thanks you!

mirguest:
> Hi Tao,
> 
> Firstly, thanks for spotting this! Secondly, please move away from legacy sensitive detectors soon. They are only there because we thought some FCC guys are still using them. Otherwise this is entirely unmaintained code. As soon as you do no longer use them, please make a pull request deleting this file and the corresponding header or alternatively notify me (or Marko).
> 
> Thanks you!

Hi @MarkusFrankATcernch ,

Thank you! I will try to remove the dependencies on the legacy code.

Tao

zaborowska:
Hi @MarkusFrankATcernch , as far as I know FCCSW never used `Geant4SensitiveDetector`, I think @vvolkl can confirm if it is still true.

MarkusFrankATcernch:
@zaborowska I know you do not use these since quite some time now. 
Unfortunately we never cleaned up the code.

PR #678, opened 2020-06-17
Iov improvements

pikacic:
Trying to use `dd4hep::IOV` we found a things that could be improved:
- use modern C++ instead of legacy C
- `IOV::Key` was allowing for different types for the two bounds of the interval, but the feature was not used and could lead to weird situations
- the value type in the `IOV::Key` was `long`, but that's not portable across architectures, so I changed it to `std::int64_t` 

BEGINRELEASENOTES

- Modernization and clean up of `dd4hep::IOV` (`using` instead of `typedef`, `constexpr` instead of `enum`)
- Make `dd4hep::IOV::Key` uniform wrt first and second type
- Use `std::int64_t`instead of `long` for `dd4hep::IOV::Key` elements

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/IOV.h
+++ b/DDCore/include/DD4hep/IOV.h
@@ -17,6 +17,7 @@
 #include <string>
 #include <limits>
 #include <algorithm>
+#include <utility>
 
 /// Namespace for the AIDA detector description toolkit
 namespace dd4hep {
--- a/DDCore/include/DD4hep/IOV.h
+++ b/DDCore/include/DD4hep/IOV.h
@@ -34,8 +34,8 @@ namespace dd4hep {
    */
   class IOVType   {
   public:
-    enum _IOVTypes { UNKNOWN_IOV = ~0x0 };
-    /// integer identifier ised internally
+    static constexpr unsigned int UNKNOWN_IOV = ~0x0;
+    /// integer identifier used internally
     unsigned int type = UNKNOWN_IOV;
     /// String name
     std::string  name;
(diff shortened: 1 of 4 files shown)


andresailer:
There are a number of tests failing:
```
The following tests FAILED:
	  4 - t_AlignDet_Telescope_dump_xml (Failed)
	  5 - t_AlignDet_Telescope_write_xml (Failed)
	  6 - t_AlignDet_Telescope_stress (Failed)
	  9 - t_AlignDet_Telescope_readback_xml (Failed)
	 10 - t_AlignDet_CLICSiD_stress_LONGTEST (Failed)
	168 - t_Conditions_Telescope_root_load_usr (Failed)
        175 - t_Conditions_CLICSiD_root_load_usr_LONGTEST (Failed)
```

DDCore/include/DD4hep/IOV.h
```diff
@@ -68,20 +70,16 @@ namespace dd4hep {
     explicit IOV() = delete;
   public:
     /// Key definition
-    typedef long Key_first_type;
-    typedef long Key_second_type;
-    typedef std::pair<Key_first_type,Key_second_type> Key;
+    using Key_value_type = std::uint64_t;
```
MarkusFrankATcernch, line 73: I am not sure if this change from "long" to "unsigned long" actually causes the problems in the tests.

pikacic: I changed it to `std::int64_t`, and I'll prepare another PR to remove the sign.

PR #679, opened 2020-06-18
Add CPack Configuration

vvolkl:
Allows to package the binaries with `make package`.

BEGINRELEASENOTES
- Add CPack Configuration

ENDRELEASENOTES


cmake/DD4hepCPack.cmake
```diff
@@ -0,0 +1,46 @@
+
+set(CPACK_PACKAGE_NAME "dd4hep")
+set(CPACK_SET_DESTDIR TRUE)
+set(CPACK_PACKAGE_HOMEPAGE_URL "https://github.com/AIDASoft/dd4hep")
+set(CPACK_PACKAGE_DESCRIPTION "Detector Description Toolkit for HEP")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY ${CPACK_PACKAGE_DESCRIPTION})
+set(CPACK_PACKAGE_VENDOR "AIDASoft")
+set(CPACK_PACKAGE_VERSION ${dd4hep_VERSION})
```
andresailer, line 8:
```suggestion
set(CPACK_PACKAGE_VERSION ${DD4hep_VERSION})
```
?

vvolkl: Ah, actually there is no variable `${DD4hep_VERSION}` either. but because of the following lines this one isn't needed, so I removed it.

andresailer: This
https://github.com/AIDASoft/DD4hep/blob/fbb34839feea6f0462a1c55fad9ba1c21584b6ee/CMakeLists.txt#L98-L102
with this
https://github.com/AIDASoft/DD4hep/blob/fbb34839feea6f0462a1c55fad9ba1c21584b6ee/cmake/DD4hepBuild.cmake#L208
doesn't create DD4hep_VERSION?

cmake/DD4hepCPack.cmake
```diff
@@ -0,0 +1,35 @@
+
+set(CPACK_PACKAGE_NAME "dd4hep")
+set(CPACK_SET_DESTDIR TRUE)
+set(CPACK_PACKAGE_HOMEPAGE_URL "https://github.com/AIDASoft/dd4hep")
```
petricm, line 4: Is this meant as homepage or repo page? Otherwise
```suggestion
set(CPACK_PACKAGE_HOMEPAGE_URL "https://github.com/AIDASoft/dd4hep")
set(CPACK_PACKAGE_HOMEPAGE_URL "https://dd4hep.cern.ch")
```

vvolkl: I don't think it matters too much, however that particular URL gives me a `SSL_ERROR_BAD_CERT_DOMAIN `. I'll change it to `https://cern.ch/dd4hep`

andresailer: I still have to create the https openshift for it

PR #681, opened 2020-07-01
Tbb dependency

andresailer:
This only addresses the issues for DDDigi in dev3 and dev4 at the moment. Changes for #675 to come.
But I think we need to add this flag in any case so people can chose if they need TBB or not and not have things silently fail

BEGINRELEASENOTES
- CMake: add option DD4HEP_USE_TBB to require TBB or not

ENDRELEASENOTES


petricm:
Superseded #682

PR #686, opened 2020-07-09
Fix bug in VolumeBuilder

MarkusFrankATcernch:
BEGINRELEASENOTES
- Fix bug in `VolumeBuilder` where `VolIDs` were not properly set to physical volumes if the `xml` supported an `id` tag.

ENDRELEASENOTES


petricm:
Can you add please a bit more of explanation to the release notes, what the bug was

PR #697, opened 2020-08-20
Add basic geometrical accessors to shapes.

MarkusFrankATcernch:
BEGINRELEASENOTES
As discussed in the thread https://github.com/cms-sw/cmssw/pull/30931 it was felt that read-only access to the basic construction parameters of a shape would be highly useful. In this pull request these data member accessors are implemented in Shapes.h

ENDRELEASENOTES


DDCore/include/DD4hep/Shapes.h
```diff
@@ -829,6 +945,27 @@ namespace dd4hep {
     Trap& setDimensions(double z, double theta, double phi,
                         double h1, double bl1, double tl1, double alpha1,
                         double h2, double bl2, double tl2, double alpha2);
+
+    /// Accessor: phi value
+    double phi() const                     { return access()->GetPhi()*dd4hep::deg;    }
```
ianna, line 950: @MarkusFrankATcernch  - IMHO we need to avoid double conversion, e.g. converting everything to radians (e.g. G4 definition) and then back to degrees (e.g. ROOT definition). We will definitely see some precision loss.

DDCore/include/DD4hep/Shapes.h
```diff
@@ -1271,6 +1471,27 @@ namespace dd4hep {
     PolyhedraRegular& operator=(PolyhedraRegular&& copy) = default;
     /// Copy Assignment operator
     PolyhedraRegular& operator=(const PolyhedraRegular& copy) = default;
+
+    /// Accessor: Number of edges
+    size_t numEdge()  const                { return access()->GetNedges();           }
```
ianna, line 1476: I think you are converting signed to an unsigned here. I'd add an `s`:
```c++
int numEdges() const
```

DDCore/include/DD4hep/Shapes.h
```diff
@@ -1271,6 +1471,27 @@ namespace dd4hep {
...
+    /// Accessor: start-phi value
+    double startPhi() const                { return access()->GetPhi1()*dd4hep::deg; }
+    /// Accessor: delta-phi value
+    double deltaPhi() const                { return access()->GetDphi()*dd4hep::deg; }
+
+    /// Accessor: r-min value
+    double z(size_t which) const           { return access()->GetZ(which);           }
```
ianna, line 1483: I think, it should be `int`

ianna (changes requested):
@MarkusFrankATcernch - thanks for taking care of it! Just a few points:
* I think, we should avoid the unit conversions. We ought to leave it to a user IMHO.
* I do not care much if the API is close to G4 or ROOT as long as it is well documented.
* Also, I cannot tell if CMS IB checks will be happy with using `size_t` in place of `int` which is the most commonly used type. Perhaps, you could try to run `clang-tidy` with a narrowing conversion check? 
Thanks!

MarkusFrankATcernch:
@ianna 
- size_t --> int OK taken into account.
- degree-> radians. We sort of have to stay consistent. The constructors also take radians. I would prefer to do the conversion.

More importantly:
Should we check if the object is valid ? This probably really costs. Using ptr()->  instead of access()-> is much cheaper, but also causes SEGV if the object is invalid. The optimizer actually directly leaves code as if the raw pointer is used.

ianna:
> @ianna
> 
> * size_t --> int OK taken into account.
> * degree-> radians. We sort of have to stay consistent. The constructors also take radians. I would prefer to do the conversion.
> 
> More importantly:
> Should we check if the object is valid ? This probably really costs. Using ptr()-> instead of access()-> is much cheaper, but also causes SEGV if the object is invalid. The optimizer actually directly leaves code as if the raw pointer is used.

I'd leave the optimisation for later. Most of the CMSSW code uses a dimensions vector any way...

cvuosalo:
The standard has been that DD4hep returns angles as radians. Let's keep that convention.

DDCore/include/DD4hep/Shapes.h
```diff
@@ -1420,6 +1669,19 @@ namespace dd4hep {
...
     EightPointSolid& operator=(const EightPointSolid& copy) = default;
+
+    /// Accessor: delta-z value
+    double dZ() const                      { return access()->GetDz();               }
+    /// Accessor: vertices
+    std::vector<double> vertices() const   {
+      const double* values = access()->GetVertices();
+      return detail::_make_vector(values, 8*2);
+    }
+    /// Accessor: singlke vertex
```
andresailer, line 1680:
```suggestion
    /// Accessor: single vertex
```

cvuosalo:
Concerning `TruncatedTube`, "The construction parameters of the TruncatedTube are stored as a string in the object's title" Since the parameters are stored, could you please provide named accessor methods that return each of the construction parameters?

DDCore/include/DD4hep/Shapes.h
```diff
@@ -829,6 +945,27 @@ namespace dd4hep {
     Trap& setDimensions(double z, double theta, double phi,
                         double h1, double bl1, double tl1, double alpha1,
                         double h2, double bl2, double tl2, double alpha2);
+
+    /// Accessor: phi value
+    double phi() const                     { return access()->GetPhi()*dd4hep::deg;    }
+    /// Accessor: theta value
+    double theta() const                   { return access()->GetTheta()*dd4hep::deg;  }
+    /// Angle between centers of x edges an y axis at low z
```
cvuosalo, line 953:
```suggestion
    /// Angle between centers of x edges and y axis at low z
```

DDCore/include/DD4hep/Shapes.h
```diff
@@ -829,6 +945,27 @@ namespace dd4hep {
...
+    double alpha1()  const                 { return access()->GetAlpha1()*dd4hep::deg; }
+    /// Angle between centers of x edges an y axis at low z
```
cvuosalo, line 955:
```suggestion
    /// Angle between centers of x edges and y axis at high z
```

ianna:
@MarkusFrankATcernch - I think, you could introduce a nested namespace for units, for example `dd4hep::units::m`
```bash
[119/607] Building CXX object DDCore/CMakeFiles/DDCore.dir/src/Conditions.cpp.o
1232FAILED: DDCore/CMakeFiles/DDCore.dir/src/Conditions.cpp.o 
1233/cvmfs/sft.cern.ch/lcg/releases/clang/8.0.0-ed577/x86_64-centos7/bin/clang++  -DBOOST_SPIRIT_USE_PHOENIX_V3 -DDD4HEP_USE_TINYXML -DDDCore_EXPORTS -I../DDCore/include -I../DDParsers/include -ftls-model=global-dynamic -Wheader-hygiene -Wno-c++1z-extensions -Winconsistent-missing-override -fdiagnostics-color=auto -Wshadow -Wdeprecated -Wno-long-long -Wformat-security -Wshadow -pedantic -Wextra -Wall -fdiagnostics-color=always -Werror              -pthread -O3 -DNDEBUG -fPIC   -std=c++17 -MD -MT DDCore/CMakeFiles/DDCore.dir/src/Conditions.cpp.o -MF DDCore/CMakeFiles/DDCore.dir/src/Conditions.cpp.o.d -o DDCore/CMakeFiles/DDCore.dir/src/Conditions.cpp.o -c ../DDCore/src/Conditions.cpp
1234../DDCore/src/Conditions.cpp:177:23: error: declaration shadows a variable in namespace 'dd4hep' [-Werror,-Wshadow]
1235  const BasicGrammar* g = access()->data.grammar;
1236                      ^
1237../DDParsers/include/Evaluator/DD4hepUnits.h:191:29: note: previous declaration is here
1238    static constexpr double g = gram;
1239                            ^
1240../DDCore/src/Conditions.cpp:193:12: error: declaration shadows a variable in namespace 'dd4hep' [-Werror,-Wshadow]
1241  KeyMaker m(detector.key(), detail::hash32(value));
1242           ^
1243../DDParsers/include/Evaluator/DD4hepUnits.h:98:29: note: previous declaration is here
1244    static constexpr double m = meter;
1245                            ^
1246../DDCore/src/Conditions.cpp:199:12: error: declaration shadows a variable in namespace 'dd4hep' [-Werror,-Wshadow]
1247  KeyMaker m(detector.key(), item_key);
1248           ^
1249../DDParsers/include/Evaluator/DD4hepUnits.h:98:29: note: previous declaration is here
1250    static constexpr double m = meter;
1251                            ^
1252../DDCore/src/Conditions.cpp:205:12: error: declaration shadows a variable in namespace 'dd4hep' [-Werror,-Wshadow]
1253  KeyMaker m(det, detail::hash32(value));
1254           ^
1255../DDParsers/include/Evaluator/DD4hepUnits.h:98:29: note: previous declaration is here
1256    static constexpr double m = meter;
1257                            ^
12584 errors generated.
1259[121/607] Building CXX object DDCore/C...s/DDCore.dir/src/ConditionsDebug.cpp.o
1260ninja: build stopped: subcommand failed.
1261The command "if [[ "$PYTHON_CHECK" == "yes" ]]; then docker exec -ti CI_container /bin/bash -c "./DD4hep/.dd4hep-ci.d/runPythonChecks.sh"; else docker exec -ti CI_container /bin/bash -c "./DD4hep/.dd4hep-ci.d/compile_and_test.sh"; fi" exited with 1.
1262
1263
1264Done. Your build exited with 1.
```

cvuosalo:
@MarkusFrankATcernch Thanks for the additions. This pull request will be very helpful for us in CMS.

MarkusFrankATcernch:
@cvuosalo Yes, but we first have to discuss in the dd4hep meeting the issue of putting the units into a separate namespace - ideally in a user-transparent way.

cmake/DD4hepBuild.cmake
```diff
@@ -36,7 +36,7 @@ endmacro(dd4hep_to_parent_scope)
 macro(dd4hep_set_compiler_flags)
   include(CheckCXXCompilerFlag)
 
-  SET(COMPILER_FLAGS -Wall -Wextra -pedantic -Wshadow -Wformat-security -Wno-long-long -Wdeprecated -fdiagnostics-color=auto)
+  SET(COMPILER_FLAGS -Wshadow -Wformat-security -Wno-long-long -Wdeprecated -fdiagnostics-color=auto -Wall -Wextra -pedantic)
```
petricm, line 39: What does this re-shuffling change?

MarkusFrankATcernch: Apparently nothing. I had the suspicion that -Wall or -Wextra after -Wshadow disables -Wshadow.
Note that the real order when compiling is inverted.
Net result is that -Wshadow does not work at all for gcc. Only for clang.

PR #718, opened 2020-09-28
Replace deprecated PyOS_AfterFork for python 3.7 and above

petricm:
BEGINRELEASENOTES
- Replace deprecated `PyOS_AfterFork` with `PyOS_AfterFork_Child` for python 3.7 and above

ENDRELEASENOTES

### DIFF
--- a/DDG4/tpython/DDPython.cpp
+++ b/DDG4/tpython/DDPython.cpp
@@ -305,7 +305,11 @@ void DDPython::prompt()  const   {
 void DDPython::afterFork()  const  {
   if ( ::Py_IsInitialized() ) {
     cout << "[INFO] Re-init python after forking....." << endl;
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
This is a clash between C++
```
#include <functional>
```
and boost due to the placeholders:
``` namespace placeholders
  {
  /* Define a large number of placeholders. There is no way to
   * simplify this with variadic templates, because we're introducing
   * unique names for each.
   */
    extern const _Placeholder<1> _1;
    extern const _Placeholder<2> _2;
    extern const _Placeholder<3> _3;
    extern const _Placeholder<4> _4;
    extern const _Placeholder<5> _5;
    extern const _Placeholder<6> _6;
    extern const _Placeholder<7> _7;
    extern const _Placeholder<8> _8;
 ....
```

PR #725, opened 2020-10-19
[CMAKE]: fail if dd4hep and root python do not match

andresailer:
BEGINRELEASENOTES
-  CMake: fail if ROOT and DD4hep python disagree

ENDRELEASENOTES

PR #728, opened 2020-10-27
CI: add builds against dev3, dev4

andresailer:
This brings back the builds against dev3 and dev4, both are currently failing some tests
(dropping the others for debugging at the moment)

reminder (basically):
dev3: against root master
dev4: against root 6.22 patches
(http://lcginfo.cern.ch/compare_releases/x86_64-centos7-gcc8-opt/dev3/dev4)

@MarkusFrankATcernch  @petricm Any idea why the failing tests?

Sadly github actions don't have an as nice allow failure as travis, so this needs to get figured out as well.

BEGINRELEASENOTES

* CMake: fix python version compatibility check between python and root used for python. Add exception for CMake 3.17.1
* Tests: fix some test configuration to run all tests after `-D DD4HEP_BUILD_EXAMPLES=ON`
* Tests: fix tests for root 6.22.04 and master

ENDRELEASENOTES


petricm:
Is `G__DDG4UserDict.cxx` not present for the test? The dictionary has pragma for `std::pair<double, std::vector<double> >`

MarkusFrankATcernch:
The first test for sure failed due to the missing dictionary.
The crash is then due to an exception, which is unhandled (as it should be), a call to std::terminate, 
which makes G4 crash on static destructors.
The original problem is the missing dictionary.

The second test does not really say why it fails, but it looks it is on the finalize
of the job without more specific details.

andresailer:
OK. I can fix the useranalysis test by creating the rootmap file for DDG4UserLib, so that library is loaded automatically.

.github/workflows/linux.yml
```diff
@@ -139,3 +139,38 @@ jobs:
     - name: Compile and test
       run: |
         docker exec CI_container /bin/bash -c "./Package/.github/scripts/compile_and_test.sh"
+
+
+  dev:
+    runs-on: ubuntu-18.04
```
petricm, line 145: I think this should get `continue-on-error:` but sadly [not fully correctly implemented](https://github.community/t/continue-on-error-allow-failure-ui-indication/16773/12)

cmake/DD4hepBuild.cmake
```diff
@@ -652,11 +652,7 @@ macro(DD4HEP_SETUP_ROOT_TARGETS)
...
     # cmake 3.17.1 doesn't include python patch level in PYTHON_VERSION
-    IF(CMAKE_VERSION VERSION_EQUAL 3.17.1)
-      string(REGEX MATCH [23]\.[0-9]+ REQUIRE_PYTHON_VERSION ${ROOT_PYTHON_VERSION})
-    ELSE()
-      SET(REQUIRE_PYTHON_VERSION ${ROOT_PYTHON_VERSION})
-    ENDIF()
+    string(REGEX MATCH [23]\.[0-9]+ REQUIRE_PYTHON_VERSION ${ROOT_PYTHON_VERSION})
     dd4hep_debug("D++> Python version used for building ROOT ${ROOT_PYTHON_VERSION}" )
     dd4hep_debug("D++> Required python version ${REQUIRE_PYTHON_VERSION}")
     FIND_PACKAGE(Python ${ROOT_PYTHON_VERSION} EXACT REQUIRED COMPONENTS Development)
```
petricm, line 658: Shouldn't it be here `REQUIRE_PYTHON_VERSION`

andresailer: yes

.github/workflows/linux.yml
andresailer: let's just not make the dev builds required. then we see that they fail and either ignore or not. Otherwise we will only notice too late.

petricm (approved):

petricm:
For the record why the exact python version check is necessary:
 - Python gives no guarantee for ABI compatibility of patch versions.
 - ROOT was bitten by Python-patch-release-not-binary-compatible a couple of times already

andresailer:
The only failing test is `t_LHeD_DDG4_LHeDACLick_as_ACLick_LONGTEST `. So when Markus confirms this fixes his problem as well, we can merge

andresailer:
Markus actually has 3.17.2 (https://github.com/AIDASoft/DD4hep/pull/725#issuecomment-720986707), added that to the check.

PR #738, opened 2020-11-10
fix units in LCIO conversions

gaede:
BEGINRELEASENOTES
- fix units in DDG4/LCIO conversions  
     - prepend namespace CLHEP where missing

ENDRELEASENOTES


gaede:
Test is failing w/ an unrelated python problem - will merge this.

PR #740, opened 2020-11-12
add back the  Geant4Output2EDM4hep plugin

gaede:
BEGINRELEASENOTES
- add back the  Geant4Output2EDM4hep plugin from EDM4hep
     - use optional flag D4HEP_USE_EDM4HEP to build this
     - requires podio and EDM4hep to be present
- fix treatment of units in Geant4Output2EDM4hep
- DDTest: add optional ddsim tests with edm4hep and lcio output (A.Sailer)
- enable D4HEP_USE_EDM4HEP in CI for dev (A.Sailer)

ENDRELEASENOTES


petricm:
This also needs some changes in `ddsim` to enable this output

andresailer:
> This also needs some changes in `ddsim` to enable this output

That was already done

gaede:
How should we deal with this optional plugin in CI and testing ?

petricm:
I would do for both LCIO and EDM4hep a silent ON e.g. `find_package([LCIO|EDM] QUITE)` so that if it is in the env it picks it up automatically, but does not require it. Then add a flag, to steer if you really require/disable/enable.

gaede:
Actually for the standard cmake I would have a preference for leaving the flags for optional dependencies as they are, i.e. OFF by default and if turned ON the package will be required to be there. I think this is cleaner for large software installations, where you want to be in control of what gets build and what not.  
For the CI testing, I think it is a good idea to build against anything that is there on the corresponding system. But I would not know how to do this....

petricm:
> Actually for the standard cmake I would have a preference for leaving the flags for optional dependencies as they are, i.e. OFF by default and if turned ON the package will be required to be there. 

There is no well defined state "as they are", as on one side LCIO behaves as you said but `assimp` the other way around
https://github.com/AIDASoft/DD4hep/blob/f890a2caffde05392ebddcde926c84f4b254f03f/DDCAD/CMakeLists.txt#L16-L25

gaede:
But your example is different: it builds w/ ASSIMP in any case - just silently uses a findable one with a fall back to explicit download. For LCIO and EDM4hep we could of course do the same - but I am not sure that this is what we want...

petricm:
ok, then do it as LCIO, and we can discus some other time how to handle`assimp`

gaede:
Should be ready to be merged now...

petricm:
Let's wait with merging, until we have the new tags propagated to `dev3`

gaede:
But the tests w/ EDM4hep on dev3 and dev4 have passed successfully  - so what exactly do we need to wait for now ?

petricm:
In dev3/4 there were the old tags, the new tags were [just merged](https://gitlab.cern.ch/sft/lcgcmake/-/merge_requests/567), so a new nightly needs to be created. I would just like to be sure that it works with the new tags.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,588 @@
...
+#include "DD4hep/VolumeManager.h"
+#include "DDG4/Geant4HitCollection.h"
+#include "DDG4/Geant4OutputAction.h"
+#include "DDG4/Geant4SensDetAction.h"
+#include "DDG4/EventParameters.h"
+
+// Geant4 headers
+#include "G4Threading.hh"
+#include "G4AutoLock.hh"
+#include <G4Version.hh>
```
petricm, line 28:
```suggestion
#include "G4Version.hh"
```
I am a bit perplexed where do the CLHEP units come from, there is no header?

andresailer:
This PR needs a rebase, there is a conflict with the added test for hepmc3
the dev builds are failing because of a problem with SIO on the lcg views, the SIOConfig.cmake are only in the base folder, not e.g., lib/cmake/SIO

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
petricm: The Units from `G4ParticleDefinition.hh` -> `PhysicalConstants.h` -> `SystemOfUnits.h` it would still good to add the header explicitly.

andresailer:
Everything is working now.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,589 @@
...
+#include "DDG4/Geant4HitCollection.h"
+#include "DDG4/Geant4OutputAction.h"
+#include "DDG4/Geant4SensDetAction.h"
+#include "DDG4/EventParameters.h"
+
+// Geant4 headers
+#include "G4Threading.hh"
+#include "G4AutoLock.hh"
+#include "G4Version.hh"
+#include "G4SystemOfUnits.hh"
```
petricm, line 29:
```suggestion
#include "CLHEP/Units/SystemOfUnits.h"
```
Sorry I was unclear, I meant CLHEP. `G4SystemOfUnits` should never be used in conjunction with DD4hep.

gaede: Why is that ? We are dealing with Geant4 here explicitly and not CLHEP - of course it is the same in the end, as
`G4SystemOfUnits.hh` includes the CLHEP header:

```c++
#include <CLHEP/Units/SystemOfUnits.h>
using CLHEP::millimeter;
using CLHEP::millimeter2;
...
```
so I find it more logical to document that we are using Geant4 units here ....

gaede: .... ah, it is the namespace issue  - forget my last comment ;-)

gaede:
Ok, I think we can now merge this finally...

PR #741, opened 2020-11-13
Miscellaneous build system & script tweaks

hansenjo:
BEGINRELEASENOTES
- Various small tweaks to CMake scripts, implementing fixes to glitches observed during testing on macOS 10.15.
- Prevent system directories like /usr/local/lib to sneak into DYLD_LIBRARY_PATH in the test environment on macOS, which can cause issues with Homebrew.
- Fix link error in DDCMS examples when using Geant4's builtin CLEHP.
- Fix broken t_CLICSiD_DDG4_g4geometry_scan_LONGTEST test condition which always matched.
ENDRELEASENOTES


cmake/DD4hepBuild.cmake
```diff
@@ -782,7 +782,7 @@ MACRO(DD4HEP_SETUP_GEANT4_TARGETS)
         PROPERTIES
         INTERFACE_INCLUDE_DIRECTORIES "${Geant4_INCLUDE_DIRS}"
       )
-      TARGET_LINK_LIBRARIES(CLHEP::CLHEP INTERFACE G4clhep)
+      TARGET_LINK_LIBRARIES(CLHEP::CLHEP INTERFACE Geant4::G4clhep)
```
andresailer, line 796: This only applies for Geant10.06, doesn't it?

hansenjo: Good point. That's very well possible. I'll try it with a 10.5 installation and add a version check if necessary.

petricm: It has to work down to Geant 10.2.2

hansenjo: Right right. I saw. Will do.

petricm:
A few suggestions as this PR addresses several things, I would split this up:

1. PR:
 - Fix compilation warnings from Apple Clang 12
 - Fix a bunch of typos
 - Fix t_CLICSiD_DDG4_g4geometry_scan_LONGTEST regex (always matched)

2. PR 
 - Correctly configure tests when building with -DDD4HEP_BUILD_EXAMPLES=ON 

and one for CLHEP+Xerces changes and one for your proposed changes to `thisdd4hep.sh`. As there might be different timescales to all these.

PR #744, opened 2020-11-13
Make example tests work when compiled together with main project

hansenjo:
This allows the examples and example tests to be built and run together with the main project in one go. Usage example:
```
mkdir build && cd build
cmake -DCMAKE_INSTALL_PREFIX=<install-dir> \
      -DDD4HEP_USE_GEANT4=ON \
      -DDD4HEP_USE_XERCESC=ON \
      -DDD4HEP_USE_TBB=ON \
      -DBoost_NO_BOOST_CMAKE=ON \
      -DDD4HEP_BUILD_EXAMPLES=ON \
      -DBUILD_TESTING=ON \
      ..
cmake --build . -j8 --target install
ctest -j8 --output-on-failure
```
This reverts commit e2f826e9c91a070d3a60f02ea497ce24634901d6.

The build system outside of "examples" assumes that DD4hep_DIR points to the source tree. Switching this meaning on the fly makes things rather fragile. For instance, `dd4hep_configure_scripts()` is broken after the `if(DD4HEP_BUILD_EXAMPLES)` block if the test is true, otherwise not. This commit consistently points DD4hep_ROOT to the installation tree and DD4hep_DIR to the source tree.

BEGINRELEASENOTES
- CMake: Fix regression that examples and example tests can now be built and run together with the main project

ENDRELEASENOTES

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -35,6 +35,16 @@ IF(NOT TARGET DD4hep::DDCore)
 ENDIF()
 include(DD4hepMacros) 
 
+IF( "$ENV{DD4hepINSTALL}" STREQUAL "" )
+  IF( DD4hep_FOUND )
+    MESSAGE( FATAL_ERROR "Must source <DD4hep-install-dir>/bin/thisdd4hep.sh first" )
+    EXIT(1)
+  ENDIF()
+  # When building the examples together with DD4hep (-DDD4HEP_BUILD_EXAMPLES=ON),
+  # thisdd4hep.sh hasn't run. Manually set up the required variables for the tests.
+  SET( ENV{DD4hepINSTALL} ${CMAKE_INSTALL_PREFIX} )
+ENDIF()
+
 dd4hep_set_compiler_flags()
 dd4hep_configure_output()
 
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -35,16 +35,6 @@ IF(NOT TARGET DD4hep::DDCore)
 ENDIF()
 include(DD4hepMacros) 
 
-IF( "$ENV{DD4hepINSTALL}" STREQUAL "" )
-  IF( DD4hep_FOUND )
-    MESSAGE( FATAL_ERROR "Must source <DD4hep-install-dir>/bin/thisdd4hep.sh first" )
-    EXIT(1)
-  ENDIF()
-  # When building the examples together with DD4hep (-DDD4HEP_BUILD_EXAMPLES=ON),
-  # thisdd4hep.sh hasn't run. Manually set up the required variables for the tests.
-  SET( ENV{DD4hepINSTALL} ${CMAKE_INSTALL_PREFIX} )
-ENDIF()
-
 dd4hep_set_compiler_flags()
 dd4hep_configure_output()
 
--- a/examples/ClientTests/CMakeLists.txt
+++ b/examples/ClientTests/CMakeLists.txt
@@ -300,7 +300,7 @@ foreach (test Assemblies BoxTrafos CaloEndcapReflection LheD_tracker MagnetField
   if (DD4HEP_USE_GEANT4)
     dd4hep_add_test_reg( ClientTests_g4material_scan_${test}_LONGTEST
       COMMAND    "${CMAKE_INSTALL_PREFIX}/bin/run_test_ClientTests.sh"
-      EXEC_ARGS  python ${DD4hep_DIR}/bin/g4MaterialScan --compact=file:${ClientTestsEx_INSTALL}/compact/${test}.xml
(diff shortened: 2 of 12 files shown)


examples/ClientTests/CMakeLists.txt
```diff
@@ -342,21 +342,22 @@ if (DD4HEP_USE_GEANT4)
     COMMAND    "${CMAKE_INSTALL_PREFIX}/bin/run_test_ClientTests.sh"
     EXEC_ARGS  python ${ClientTestsEx_INSTALL}/scripts/Check_Air.py
     -geometry  ${ClientTestsEx_INSTALL}/compact/Check_Air.xml batch
+    REQUIRES   DDG4 Geant4
```
andresailer, line 345: This argument doesn't exist for dd4hep_add_test_reg
```suggestion
```

andresailer:
Instead of using DD4hep_ROOT, please directly use CMAKE_INSTALL_PREFIX.

examples/CMakeLists.txt
```diff
@@ -35,6 +35,16 @@ IF(NOT TARGET DD4hep::DDCore)
...
 include(DD4hepMacros) 
 
+IF( "$ENV{DD4hepINSTALL}" STREQUAL "" )
+  IF( DD4hep_FOUND )
+    MESSAGE( FATAL_ERROR "Must source <DD4hep-install-dir>/bin/thisdd4hep.sh first" )
+    EXIT(1)
+  ENDIF()
+  # When building the examples together with DD4hep (-DDD4HEP_BUILD_EXAMPLES=ON),
+  # thisdd4hep.sh hasn't run. Manually set up the required variables for the tests.
+  SET( ENV{DD4hepINSTALL} ${CMAKE_INSTALL_PREFIX} )
```
andresailer, line 45: Does this actually do anything for the tests? It sets the environment variable for the cmake run, but when running ctest this isn't set any longer, is it?

andresailer: I see that some test definitions use ENV{DD4hepInstall}, can't this be DD4hep_ROOT?

examples/ClientTests/CMakeLists.txt
hansenjo: That's odd. I don't recall writing that line. I suspect it slipped in due to something I did wrong with git. So yeah, it can go.

andresailer: Probably from the `git revert`

examples/CMakeLists.txt
hansenjo: The environment variable is used in some places in the CMake code. For example, in examples/CLICSiD/CMakeLists.txt:

```cmake
  COMMAND    "${CMAKE_INSTALL_PREFIX}/bin/run_test_CLICSiD.sh"
  EXEC_ARGS  python ${DD4hep_ROOT}/bin/checkGeometry --compact=file:$ENV{DD4hepINSTALL}/DDDetectors/compact/SiD.xml
```

When cmake runs, `$ENV{DD4hepInstall}` will be evaluated, and the references to it disappear from the Makefiles, so it doesn't have to be set later on when running ctest. The tests set up the required environment again from scratch in their respective setup scripts. I think they only care about `$DD4hep_DIR` at runtime.

I'm not sure `$ENV{DD4hepInstall}` can be replaced with `${DD4hep_ROOT}` since, at the moment, you also want to be able to build the examples separately in a second build step (and possibly install them somewhere else). In that scenario, `${DD4hep_ROOT}` is not set and cmake picks up the location of the main DD4hep installation from `$ENV{DD4hepInstall}`. In fact, that's how the example tests are currently built in the CI workflow.

andresailer: `${DD4hep_ROOT} is set in DD4hepConfig.cmake
https://github.com/AIDASoft/DD4hep/blob/1b66f00c4793677f621bf21a812afb7af2a7b5e1/cmake/DD4hepConfig.cmake.in#L25
So if the examples are based on an existing installation, DD4hep_ROOT will also be available.

hansenjo: I see. Well, then the detour through the environment can be avoided. Much cleaner indeed. Let me make the change and try it out.

andresailer: To be clear: I propose these changes

```diff
-  EXEC_ARGS  python ${DD4hep_ROOT}/bin/checkGeometry --compact=file:$ENV{DD4hepINSTALL}/DDDetectors/compact/SiD.xml
+  EXEC_ARGS  python ${DD4hep_ROOT}/bin/checkGeometry --compact=file:${DD4hep_ROOT}/DDDetectors/compact/SiD.xml
```

hansenjo: Yup, exactly what I had in mind.

andresailer: Thanks!

andresailer (approved):
Thanks again!

PR #745, opened 2020-11-13
Improve thisdd4hep.sh setup scripts

hansenjo:
BEGINRELEASENOTES
- Improve path handling in setup scripts, by avoiding duplicates in paths. Also support whitespaces in paths and don't add system paths to library paths on macOS

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,12 @@ IF(${CMAKE_CXX_STANDARD} LESS 14)
   MESSAGE(FATAL_ERROR "DD4hep requires at least CXX Standard 14 to compile")
 ENDIF()
 
+# ----for APPLE scripts have to set the DYLD_LIBRARY_PATH
+if(APPLE)
+  set(USE_DYLD 1)
+  set(CMAKE_MACOSX_RPATH 1)
+endif()
+
 ###############################
 # Define DD4hep build options #
 ###############################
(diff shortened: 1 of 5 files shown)


hansenjo:
I'm checking why the CI checks are failing with this. The script works fine on my private machines.

hansenjo:
This seems to be working now in the CI environment as well. The PR is ready to be reviewed.

CMakeLists.txt
```diff
@@ -58,6 +58,12 @@ IF(${CMAKE_CXX_STANDARD} LESS 14)
   MESSAGE(FATAL_ERROR "DD4hep requires at least CXX Standard 14 to compile")
 ENDIF()
 
+# ----for APPLE scripts have to set the DYLD_LIBRARY_PATH
+if(APPLE)
+  set(USE_DYLD 1)
+  set(CMAKE_MACOSX_RPATH 1)
```
petricm, line 64: `set(CMAKE_MACOSX_RPATH 1)` is set in `cmake/DD4hepBuild.cmake` along with all the rpath treatment. And `USE_DYLD` could be drooped entirely in favor of using the variable `APPLE`

petricm (changes requested):

CMakeLists.txt
hansenjo: Sure, that works. Requested changes made in commit d926c9692b5d9aacb96585ba756b7b42794c6115

PR #746, opened 2020-11-13
Fix fill_dd4hep_library_path, set XercesC CMake variables 

hansenjo:
BEGINRELEASENOTES
- On macOS, make sure that paths to external Xerces-C and CLHEP libraries are correctly added to DD4HEP_LIBRARY_PATH in the environment while CMake runs.
- On macOS, keep `/usr/local/lib` out of library paths as this may cause issues with ROOT.

ENDRELEASENOTES


petricm:
Can you write a bit better release notes in BEGIN/ENDRELEASENOTES this does not really explain much

hansenjo:
I've updated the release notes to be more descriptive.

PR #753, opened 2020-11-20
Clarify units inconsistencies 

petricm:
BEGINRELEASENOTES
- Always use `CLHEP/Units/SystemOfUnits.h` and never `G4SystemOfUnits.hh`
  -  in `G4SystemOfUnits.hh`units are directly in the top level namespace
  - in `CLHEP/Units/SystemOfUnits.h` units are in the `CLHEP::` namespace
- Remove all usage `using namespace CLHEP` and `using CLHEP::__some_unit__` to clarify which unit is used

ENDRELEASENOTES

Please everybody review these files very carefully not just the diff but the whole file if I have missed something that should be `CLHEP::`

It was also checked that `G4SystemOfUnits.hh` is not included in any Geant4 header file


andresailer (approved):
Looks good to me!

PR #758, opened 2020-11-24
Add edm4hep

andresailer:
Like #740  but adding tests to CI

BEGINRELEASENOTES

- 

ENDRELEASENOTES


DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,589 @@
...
+#include "DDG4/Geant4HitCollection.h"
+#include "DDG4/Geant4OutputAction.h"
+#include "DDG4/Geant4SensDetAction.h"
+#include "DDG4/EventParameters.h"
+
+// Geant4 headers
+#include "G4Threading.hh"
+#include "G4AutoLock.hh"
+#include <G4Version.hh>
+#include <G4SystemOfUnits.hh>
```
petricm, line 29: NO `G4SystemOfUnits` please use the CLHEP header

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -0,0 +1,589 @@
...
+      auto sch = sCaloHitColl->create() ;
+
+      const Geant4Calorimeter::Hit* hit = coll->hit(i);
+
+      sch.setCellID( hit->cellID );
+      sch.setPosition({
+           float(hit->position.x()/CLHEP::mm),
+           float(hit->position.y()/CLHEP::mm),
+           float(hit->position.z()/CLHEP::mm)});
+      sch.setEnergy( hit->energyDeposit );
```
petricm, line 505: no cast to unit?

gaede:
Should we now continue this development here or in #740 ?

petricm:
Does not matter, as you wish, you can cheery pick the commits of André or continue here. But dev3 and 4 are still broken because of the pair business so we cannot test this until that is fixed...

andresailer:
All the tests are now passing on dev3/4, will drop the `pair grep`

gaede:
Close and continue in #740 w/ cherry-picked CI stuff from here.

Issue #760, opened 2020-11-26
Bug in MCParticle list in conversion from HepMC to LCIO through DDSIM


eleogran:
I have a HepMC file with the correct MCParticle list.
When giving it as input to ddsim, the LCIO output file results in a wrong MCParticle list.

- HepMC file with 100k events e+e- -> Z -> nu N, with N -> eqq generated with DelphesPythia8:
/afs/cern.ch/work/e/eleogran/public/MCParticleList_issue/HNL.hepmc
- event 198: printout of the list of particle pdg and vertex position from the hepmc file:
/afs/cern.ch/work/e/eleogran/public/MCParticleList_issue/evt_198_mc_info.txt
- ddsim steering file:
/afs/cern.ch/work/e/eleogran/public/MCParticleList_issue/fcc_steer.py
- LCIO output file:
/afs/cern.ch/work/e/eleogran/public/MCParticleList_issue/HNLOutput.slcio
NB: event 198 is event 2 in this file, i.e. to look at it: dumpevent HNLOutput.slcio 2 | less

In the MCParticle list from the lcio file, particles with index = 7 and > 44 have vertex position:
(-8.35e-01, 3.80e-01,-7.65e-02)
rather than 
(-3.73e+02, 1.64e+02,-3.31e+01)

I suspect a problem in the Lorentz boost for the crossing angle, which in my steering file is set:
SIM.crossingAngleBoost = 0.015

If I change this to 0.0, particles with index = 7 and > 44 have vertex position:
(-6.54e-309, 2.88e-309,-5.80e-310)
while the other particles are unchanged.

I have used /cvmfs/clicdp.cern.ch/iLCSoft/builds/2020-02-07/x86_64-slc6-gcc62-opt/.
and run:
$ ddsim --steeringFile fcc_steer.py

andresailer:
As far as I can tell neither setting of Lorentz boost gives you a correct result. With the lorentzboost you get a short lifetime for the N (right handed neutrino?), without the boost there is 0 lifetime. So I think the problem isn't in the Lorentzboost, or at least not exclusively.
The particles between 8 and 44 are not given to geant4 (quarks and "strings" and such), so their information isn't updated.

eleogran:
Correct, neither result is the right one.
As the boost was affecting only the particles with the wrong vertex, I thought that there could be something weird in the boost. Do I understand from your comment that it's a problem of Geant4 then?

andresailer:
[...]

PR #765, opened 2020-12-03
Geant4InputHandling: create G4PrimaryParticle with 4 vector that cons…

andresailer:
…trains also the mass, setting the mass after, affects the momentum

BEGINRELEASENOTES
- DDG4: Geant4InputHandling: create particle with 4 vector (vecP, E), solves issue where the dynamic mass differs from the PDG mass (e.g., via particle.tbl / extraParticles), #760 

ENDRELEASENOTES

I think in our input readers, we always have a consistency between momentum, mass and energy, so this shouldn't be a problem.
I Still need to check, but opening this for discussion at the meeting later today.

cf. https://github.com/AIDASoft/DD4hep/issues/760#issuecomment-737376619
https://github.com/AIDASoft/DD4hep/blob/b5307693a98868079619a835e7381c42f9b2414b/DDG4/src/Geant4InputHandling.cpp#L354
I added `g4->Print()` before and after this line, which gives:
```
==== PDGcode 9900012  Particle name nu_Re
 Assigned charge : 0
     Momentum ( -28.3765[GeV/c], 12.5034[GeV/c], -2.5186[GeV/c] )
     kinetic Energy : 31.0612 [GeV]
     Mass : 0.05 [GeV]
     Polarization ( 0, 0, 0 )
     Weight : 1
<<<< End of link
Massive
==== PDGcode 9900012  Particle name nu_Re
 Assigned charge : 0
     Momentum ( -58.1954[GeV/c], 25.6423[GeV/c], -5.16523[GeV/c] )
     kinetic Energy : 31.0612 [GeV]
     Mass : 50 [GeV]
     Polarization ( 0, 0, 0 )
     Weight : 1
<<<< End of link
```

### DIFF
--- a/DDG4/src/Geant4InputHandling.cpp
+++ b/DDG4/src/Geant4InputHandling.cpp
@@ -351,7 +351,9 @@ static G4PrimaryParticle* createG4Primary(const Geant4ParticleHandle p)  {
     g4 = new G4PrimaryParticle(def, p->psx, p->psy, p->psz, p.energy());
     g4->SetCharge(p.charge());
   }
-  g4->SetMass(p->mass);
+  // The particle is fully defined with the 4-vector set above, setting the mass isn't necessary, not
+  // using the 4-vector, means the PDG mass is used, and the momentum is scaled if the mass is set here
+  // g4->SetMass(p->mass);
   return g4;
 }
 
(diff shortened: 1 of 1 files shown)


petricm:
If we are already at hepmc what about enabling it in the CI `DD4HEP_USE_HEPMC3`

andresailer:
a) I will add the HEPMC3 flag in another PR, as this is unrelated.

b) The `energy` is any calculated with the mass of the particle
https://github.com/AIDASoft/DD4hep/blob/7c2147cb74038a4fd6acadd4fb5dfc9bb8ed4371/DDG4/include/DDG4/Geant4Particle.h#L289-L292

So this change here is always OK.

PR #767, opened 2020-12-07
Improve thread efficiency of Evaluator

Dr15Jones:
BEGINRELEASENOTES
- Made internal classes used by Evaluator to be less stateful
- Switch to using a read/write lock as modifications to the Evaluator are much less frequent than reads.
- Added unit test for Evaluator

ENDRELEASENOTES


Dr15Jones:
FYI @ianna @cvuosalo

PR #775, opened 2021-01-12
AlignDet Test: number conditions depends on DEBUG or not

andresailer:
Fix the failing test because of different number of conditions with DEBUG and without.
@MarkusFrankATcernch does that make sense to you? I am not sure what this is counting in the test

https://github.com/AIDASoft/DD4hep/blob/12caacb209af70835dbdd48efa7ef8d0b77b3dcd/examples/AlignDet/src/AlignmentExample_read_xml.cpp#L102

### DIFF
--- a/examples/AlignDet/CMakeLists.txt
+++ b/examples/AlignDet/CMakeLists.txt
@@ -111,14 +111,22 @@ dd4hep_add_test_reg( AlignDet_Telescope_align_nominal
   )
 #
 #---Testing: Load Telescope geometry and read and print alignments --------
+IF(DD4HEP_BUILD_DEBUG STREQUAL "ON")
+  SET(EXPECTED_CONDITIONS 52)
+  SET(EXPECTED_TS 33)
+ELSE()
+  SET(EXPECTED_CONDITIONS 40)
+  SET(EXPECTED_TS 21)
+ENDIF()
 dd4hep_add_test_reg( AlignDet_Telescope_readback_xml
   COMMAND    "${CMAKE_INSTALL_PREFIX}/bin/run_test_AlignDet.sh"
   EXEC_ARGS  geoPluginRun -volmgr -destroy -plugin DD4hep_AlignmentExample_read_xml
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
This fixes the symptom, but not the cause. I cannot see any reason why there should be a difference in the number of conditions.
I think you can commit it for now, but ultimately I will have to investigate why the numbers are not equal....

andresailer:
Around the time the DEBUG and CONDITIONS MINSIZE etc were introduced you also changed the expected output of this test
E.g.: https://github.com/AIDASoft/DD4hep/commit/c0fb8a85837f3a429760ebd15ee4f23fcc8bda58#diff-ca3028ca0ce04de26ce9f94f94ff9c6fc75aa4f86542a0c72a2e70fd8e8119c3L118-R118

The new_cond.xml file used in the test is written by another testing, depending if DEBUG is ON or OFF the name attribute is filled, and if that is filled the count increases.
E.g.:
FOO/BAR/BAZ increases the count by three.

```xml
      <!-- DDCond conditions for DetElement /world/Telescope/module_5 Total of 8 Entries.  -->
      <detelement path="/world/Telescope/module_5">
        <value key="371A47FF557CADA4" name="FOO" type="float" value="6.11668e+08"/>
        <pressure key="371A47FF7C80AB98" name="BAR" unit="hPa" value="9.79999484e+02"/>
        <sequence key="371A47FFBD303AB3" name="BAZ" type="vector[int]">[ 0 , 1 , 2 , 3 , 4 , 5 ]</sequence>
        <value key="371A47FFBDB2FF6D" name="" type="double" value="6.1166762e+08"/>
        <temperature key="371A47FFC06BF8AB" name="" unit="kelvin" value="2.73000000e+02"/>
        <sequence key="371A47FFD2926577" name="" type="list[int]">[ 0 , 1 , 2 , 3 , 4 , 5 ]</sequence>
        <alignment_delta key="371A47FF6E87AB8E" name="alignment_delta"/>
      </detelement>
```


So, what is the "pool size" and why are the names counted for that?

PR #778, opened 2021-01-29
Move Grammar headers to public include directory

MarkusFrankATcernch:
BEGINRELEASENOTES
- Move the header DD4hep/detail/Grammar_parsed.h to DD4hep/GrammarParsed.h
- Move the header DD4hep/detail/Grammar_unparsed.h to DD4hep/GrammarUnparsed.h
- Try to improve type agnostic ABI cast. Works for down casts only.
ENDRELEASENOTES


andresailer:
The EDM4hep module fails to compile, this is only dev3/4 because it is not build in the other jobs.

Issue #784, opened 2021-02-15
Tube Segment as a SubtractionSolid causes leftover material


niermann999:
When checking the our example geometry it for volume overlaps, it was found that the use of the TGeoTubeSeg instead of TGeoTube as a SubtractionSolid via dd4hep Tube class produces a bit of leftover material/volume boundary where the angle limits close. Placing another volume inside the subtraction then causes overlaps. The problem does not exist with a union of two tubes with different angle limits. TGeoTube does not seem to be accessible through dd4hep shapes?

ROOT      6.23/01
Geant4    4.10.7.0  
dd4hep    latest 15.02.21 (master ba3c7a0f820278626ea0baa11c9ee9c3ad3b1bb3)

code 
```c++
     xml_comp_t x_sub = x_comp.child(_U(subtraction));
     Tube tubeCutout(x_sub.rmin(), x_sub.rmax(), 1.1 * x_comp.length());

      // Create the subtraction
      componentVolume = Volume(
          compName,
          SubtractionSolid(
              trapShape, tubeCutout,
              Position(x_sub.x_offset(), x_sub.y_offset(), x_sub.z_offset())),
          oddd.material(x_comp.materialStr()));
```

petricm:
We have looked into this a bit further we believe there is potentially s small problem with the units conversion. As DD4hep uses radians and full cylinder is in phi 0 - 2π however when we create a TGeoTube we convert 2π to 360 with all the rounding we might end at 359.999999999 and then we hit these line in TGeoTube.cxx

https://github.com/root-project/root/blob/02b422c7224eee65649d5f906b327961e4e2c50e/geom/geom/src/TGeoTube.cxx#L1505-L1516

Which results that `DistFromInside` is calculated wrongly since `if ((fPhi2-fPhi1)>=360.)` giver `false` but should be `true`. It however not clear that this would solve all the problems around tubes.

petricm:
@niermann999 can you try out #793 and report if this potentially solves your problem?

niermann999:
Initially I used a unionsolid of two tubes with a different angles to make the overlap disappear. A little later, however, the overlap reappeared somehow, although to the best of my knowledge I did not changed anything related to the subtraction or the volume placement inside.
[...]

PR #793, opened 2021-03-12
Add special treatment for full tube to Tube creator

petricm:
BEGINRELEASENOTES
- Account for possible units rounding bug in Tube creation (address #784 )

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/Shapes.cpp
+++ b/DDCore/src/Shapes.cpp
@@ -318,7 +318,12 @@ Cone& Cone::setDimensions(double z, double rmin1, double rmax1, double rmin2, do
 
 /// Constructor to be used when creating a new object with attribute initialization
 void Tube::make(const string& nam, double rmin, double rmax, double z, double start_phi, double end_phi) {
-  _assign(new TGeoTubeSeg(nam.c_str(), rmin, rmax, z, start_phi/units::deg, end_phi/units::deg),nam,TUBE_TAG,true);
+  // Check if it is a full tube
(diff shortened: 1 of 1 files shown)


petricm:
I however honestly hope that a tube 0-360 has the same local coordinate system as 90-450 and not rotated for 90?

DDCore/src/Shapes.cpp
```diff
@@ -318,7 +318,12 @@ Cone& Cone::setDimensions(double z, double rmin1, double rmax1, double rmin2, do
 
 /// Constructor to be used when creating a new object with attribute initialization
 void Tube::make(const string& nam, double rmin, double rmax, double z, double start_phi, double end_phi) {
-  _assign(new TGeoTubeSeg(nam.c_str(), rmin, rmax, z, start_phi/units::deg, end_phi/units::deg),nam,TUBE_TAG,true);
+  // Check if it is a full tube
+  if(fabs(end_phi-start_phi-2*M_PI)<10e-6){
+    _assign(new TGeoTubeSeg(nam.c_str(), rmin, rmax, z, 0, 360.),nam,TUBE_TAG,true);
```
andresailer, line 323:
```suggestion
    _assign(new TGeoTubeSeg(nam.c_str(), rmin, rmax, z, start_phi/units::deg, start_phi/units::deg+360.),nam,TUBE_TAG,true);
```

andresailer: Proposed  by @MarkusFrankATcernch

PR #799, opened 2021-03-25
Geant4: flush the buffer to fix the problem with the garbage output a…

### LINKED ISSUE #798: Error in dev3 test
I am trying to figure out if this is caused by the changes in DD4hep, or root...

Failure is because "Exception" is found in the log
https://github.com/AIDASoft/DD4hep/runs/2183002592?check_suite_focus=true#step:4:4630

Not clear why this is printed now.

andresailer:
…t the end

So https://gitlab.cern.ch/geant4/geant4/-/blob/geant4-10.7-release/source/processes/hadronic/models/de_excitation/management/src/G4DeexPrecoParameters.cc#L294
Never uses G4endl;.
And we never use G4endl ourselves. So the buffer never gets flushed until the end.
So we just add one call...

BEGINRELEASENOTES
- DDG4: flush the geant4 strstream buffer after run. fixes #798 

ENDRELEASENOTES

### DIFF
--- a/DDG4/src/Geant4Kernel.cpp
+++ b/DDG4/src/Geant4Kernel.cpp
@@ -300,7 +300,10 @@ int Geant4Kernel::initialize() {
 
 int Geant4Kernel::run() {
   try  {
-    return Geant4Exec::run(*this);
+    auto result = Geant4Exec::run(*this);
+    // flush the geant4 stream buffer
+    G4cout << G4endl;
+    return result;
   }
   catch(const exception& e)   {
     printout(FATAL,"Geant4Kernel","+++ Exception while simulating:%s",e.what());


simonspa:
Very interesting, I just reported that to G4: https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2348 but didn't trace it back to a missing `G4endl;`. In G4 10.6 there is a similar problem with the RadioactiveDecay class, in case you're still supporting that.

andresailer:
@simonspa This fix should prevent at least the "Garbage" printing at the end of the executable in all these cases, also for G4 10.6. I think what really hit us here was a bug in the G4strstreambuffer d'tor that didn't null terminate the buffer before printing it out.
Of course the lack of `G4endl` means that the printout happens at different location to the call to Dump itself as well.

andresailer:
Opened a bug report https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2350

Issue #815, opened 2021-04-23
geoDisplay error: Segmentation: FAILED to create segmentation: GridDRcalo


saraheno:
So now that I have a rough dual crystal calorimeter working (Thanks so much for your help!) I wanted to add the back end spaghetti dual readout hadron calorimeter.  I use a a start the lovely code by Sanghyuan Ko at https://github.com/SanghyunKo/dual-readout

I copied what I thought were the relevent pieces into my area: https://github.com/saraheno/DD4hep/tree/master/examples/DDDualCrystal/src

The code from his repository has names with "DR" in them.

It all seemed to compile fine.

But when I tried to look at his calorimeter using geoDisplay using

geoDisplay DRcalo.xml (this file is at https://github.com/saraheno/DD4hep/blob/master/examples/DDDualCrystal/SCEPCdetector/DRcalo.xml )

I find somehow I am missing some code or some link or something

[eno@hepcms-in2 SCEPCdetector]$ geoDisplay DRcalo.xml
PersistencyIO    INFO  +++ Set Streamer to dd4hep::OpaqueDataBlock
Info in <TGeoManager::TGeoManager>: Geometry default, Detector Geometry created
Info in <TGeoNavigator::BuildCache>: --- Maximum geometry depth set to 100
CompactLoader    INFO  +++ Processing compact file: DRcalo.xml with flag BUILD_DEFAULT
DD4hep           WARN  ++ STD conditions NOT defined by client. NTP defaults taken.
Detector         INFO  *********** Created World volume with size: 1000 1000 1000
Info in <TGeoManager::SetTopVolume>: Top volume is world_volume. Master volume is world_volume
Segmentation     ERROR FAILED to create segmentation: GridDRcalo. [Missing factory]
RunPlugin        ERROR ++ Exception while executing plugin DD4hep_CompactLoader:
                Segmentation: FAILED to create segmentation: GridDRcalo. [Missing factory]
dd4hep: Error interpreting XML nodes of type <readout/>
dd4hep: Error interpreting XML nodes of type <readouts/>
dd4hep: while parsing DRcalo.xml
dd4hep: with plugin:DD4hep_CompactLoader
Application      ERROR geoPluginRun: Uncaught exception: RunPlugin: ++ Exception while executing plugin DD4hep_CompactLoader:
                Segmentation: FAILED to create segmentation: GridDRcalo. [Missing factory]
dd4hep: Error interpreting XML nodes of type <readout/>
[...]

PR #817, opened 2021-04-26
DDCore: move create_segmentation to header file so it can be used outside of DD4hep

andresailer:
@MarkusFrankATcernch as far as I can tell this is needed to be able to do

```c++
#include "DD4hep/detail/SegmentationsInterna.h"
//....
DECLARE_SEGMENTATION(GridDRcalo, create_segmentation<dd4hep::DDSegmentation::GridDRcalo>)
```

for #815 
So that the segmentation_constructor is created and part of the components file?

BEGINRELEASENOTES
- Move `create_segmentation` to DDCore/include/DD4hep/detail/SegmentationsInterna.h  to allow segmentation creation in other libraries

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/detail/SegmentationsInterna.h
+++ b/DDCore/include/DD4hep/detail/SegmentationsInterna.h
@@ -121,5 +121,13 @@ namespace dd4hep {
   /// Default destructor
   template <typename IMP> inline SegmentationWrapper<IMP>::~SegmentationWrapper()  {
   }
+
+  namespace {
+    template<typename T> SegmentationObject*
+    create_segmentation(const BitFieldCoder* decoder)  {
+      return new SegmentationWrapper<T>(decoder);
+    }
+  }
+
 }         /* End namespace dd4hep                       */
 #endif // DD4HEP_DETAIL_SEGMENTATIONSINTERNA_H
--- a/DDCore/src/plugins/ReadoutSegmentations.cpp
+++ b/DDCore/src/plugins/ReadoutSegmentations.cpp
@@ -18,13 +18,6 @@
 using namespace dd4hep;
 using namespace dd4hep::DDSegmentation;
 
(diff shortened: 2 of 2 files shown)


MarkusFrankATcernch:
DECLARE_SEGMENTATION creates the factory entry in the library.
The 'list_components' opens the library during make, reads all these entries
and writes the names into the components file.
If you do not have this entry, the factory mechanism is broken.

andresailer:
Thanks for the confirmation!

petricm:
If users are calling `detail/SegmentationsInterna.h` and this is a necessary to create a segmentation then this header should not be in `detail`.

andresailer:
Yeah, I kind of agree :)
I will try re-arranging so that the necessary pieces get pulled in via Segmentation.h or something like this?

andresailer:
I re-arranged and removed some things in DDSegmentation that wasn't needed as far as I could tell.

I ran into some weird namespace issue, so I added DD4hep/Factories.h to all the Segmentation and replaced the REGISTER_SEGMENTATION macro with DECLARE_SEGMENTATION.
Removed some stuff for the obsolete REGISTER_SEGMENTATION
Removed the file with the now obsolete DECLARE_SEGMENTATION stuff

andresailer:
Moving the DECLARE_SEGMENTATION to the individual segmentations leads these calls to be part of DDCore, which means any plugin linking against DDCore has these in their components file. Will bring back the ReadoutSegmentations.cpp file in plugin and leave the DECLARE_SEGMENTATION calls commented for people creating their own segmentations, in which case they probably create their own plugin in any case and it should be OK.

vvolkl:
@andresailer so user code should also just replace `REGISTER_SEGMENTATION` with `DECLARE_SEGMENTATION`? Could you add a note to the Changelog?

PR #825, opened 2021-05-07
Fix as reported from CMS

MarkusFrankATcernch:
BEGINRELEASENOTES
- Fix as reported from CMS
ENDRELEASENOTES


ianna:
@MarkusFrankATcernch - do you think, this is it?

MarkusFrankATcernch:
@ianna 
Yes. I think this is it. Please give it a try.

ianna:
@cvuosalo - FYI - we need an IB asap

PR #847, opened 2021-07-21
save encoding string for edm4hep format

fdplacido:
BEGINRELEASENOTES
- EDM4hepOutput: On collection creation for EDM4hep, save CellIDEncodingString

ENDRELEASENOTES


andresailer:
There are some failing tests in dev3, but they also fail in the master branch when I just re-ran the CI for master.

Issue #841, opened 2021-06-05
feature: Extend/inherit visualization attributes. 


whit2333:
Is there a way to reuse visualization attributes and extend/inherit them?

For example:

```
<vis name="MyColor"                r="1.0" g="0.0" b="0.0" />
<vis name="MyDetector" alpha="0.5" r="1.0" g="0.0" b="0.0"     showDaughters="false" visible="true" />
<vis name="Detector2"  alpha="0.1" r="1.0" g="0.0" b="0.0"     showDaughters="true"  visible="true" />
```

Instead could be this:

```
<vis name="MyColor"  r="1.0" g="0.0" b="0.0" />
<vis name="MyDetector" extend="MyColor"    alpha="0.5" showDaughters="false" visible="true" />
<vis name="Detector2"  extend="MyDetector" alpha="0.1" showDaughters="true" />
```

Now I can change "MyColor" and it propagates to all extended attributes. 

This would be a very nice feature to add if it does not already exist in some form.

MarkusFrankATcernch:
This functionality does not exist. Could be done at the level of the compact XML 
if "MyColor" would be fully qualified and not only half. 
However, do features like this really help or only add to the confusion?

whit2333:
@MarkusFrankATcernch 
Yes, "MyColor" would be fully qualified. The idea is that it defines part of the subsequent attributes.
For example, this means N slightly different attributes can have their RGB values changed via one attribute. This is **very** useful IMO, especially with detectors many included files.

PR #851, opened 2021-07-23
Added VisAttr ref extension compact construction

whit2333:
Closes https://github.com/AIDASoft/DD4hep/issues/841

BEGINRELEASENOTES
 - Using `ref="OtherVisName"` attribute with the `vis` tag, the visualization attribute is an extension of
`"OtherVisName"` which is used to initialize the new vis attribute.
 - The new VisAttr inherits all the properties  of the ref and additional arguments override these values.

Example where the only difference is the `alpha` value.

```
  <vis name="SiVertexBarrelModuleVis"
       alpha="1.0" r="1.0" g="0.75" b="0.76"
       drawingStyle="wireframe"
       showDaughters="false"
       visible="true"/>

  <vis name="SiVertexEndcapModuleVis"
       ref="SiVertexBarrelModuleVis"
       alpha="0.5"/>
```
ENDRELEASENOTES

### DIFF
--- a/DDCore/src/plugins/Compact2Objects.cpp
+++ b/DDCore/src/plugins/Compact2Objects.cpp
@@ -831,8 +831,8 @@ template <> void Converter<VisAttr>::operator()(xml_h e) const {
     if(!refAttr.isValid() )  {
         throw runtime_error("reference VisAttr " + refName + " does not exist");
     }
-    // Not sure if there is an easy deep copy constructor
-    // Just copying things manually
+    // Just copying things manually.
+    // I think a handle's copy constructor/assignment would reuse the underlying pointer... maybe?
     refAttr.argb(alpha,red,green,blue);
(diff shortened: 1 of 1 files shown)


whit2333:
@MarkusFrankATcernch, can you have a look at this PR?  I'm not sure if the checks fail because of something here.

MarkusFrankATcernch:
@whit2333 
This implementation will not work as expected. 
It will apply defaults for all properties, which are not part of the XML definition.
I have implemented the stuff already some time ago, but did not push. 
If you agree I commit my version and close this PR.
Otherwise you have to NOT apply all defaults if the "ref" property is present. Then I discard my changes.

whit2333:
@MarkusFrankATcernch , You are right. I did push changes.

MarkusFrankATcernch:
@andresailer Please merge this PR. Thanks!

PR #852, opened 2021-07-23
WIP: Added DD4POD output datamodel and plugin

whit2333:
- DD4pod is a small PODIO datamodel and output action plugin which is
designed to mirror Geant4Output2Root
- The datamodel isn't a perfect match (mostly due to missing data types
in PODIO)
- Also, would like to add PhotonCounter but that can be a separate merge
request.

BEGINRELEASENOTES
- DD4pod is a small PODIO datamodel and output action plugin which is
designed to mirror Geant4Output2Root
- The datamodel isn't a perfect match (mostly due to missing data types
in PODIO)

ENDRELEASENOTES

### DIFF
--- a/DDG4/dd4pod/CMakeLists.txt
+++ b/DDG4/dd4pod/CMakeLists.txt
@@ -1,0 +1,72 @@
+cmake_minimum_required(VERSION 3.3 FATAL_ERROR)
+
+message(" ======================= ")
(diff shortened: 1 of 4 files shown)


andresailer:
As EIC/Athena has moved to EDM4hep, I guess this PR will not be needed.

PR #855, opened 2021-07-28
[testing] separate import test for ddg4 and rest of dd4hep

vvolkl:
This avoids a test failure when DD4hep is built with `-DDD4HEP_USE_GEANT4=OFF` - since there is no DDG4 module built in this case, the test trying to import it is doomed to fail. So I propose test for the import of DDG4 in a separate file / test that is only enabled when DDG4 is built.

Due to the `from test_import import *` the new test actually tests also the other imports (again), but at least there is no code duplication.

BEGINRELEASENOTES
- [testing] separate import test for ddg4 and rest of dd4hep

ENDRELEASENOTES

### DIFF
--- a/DDTest/python/test_import_ddg4.py
+++ b/DDTest/python/test_import_ddg4.py
@@ -5,8 +5,8 @@ This file adds a test for the DDG4 python module.
 """
 
 from __future__ import absolute_import
-import pytest
 from test_import import test_module
+
 
 def test_module_ddg4():
     test_module('DDG4')
--- a/DDTest/python/test_import_ddg4.py
+++ b/DDTest/python/test_import_ddg4.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+"""
+Some imports to make sure that the DD4hep environment is complete.
+This file adds a test for the DDG4 python module.
+"""
+
+from __future__ import absolute_import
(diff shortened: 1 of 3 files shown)


MarkusFrankATcernch:
@vvolkl Please check the error

vvolkl:
@MarkusFrankATcernch ah, `from xyz import *` is not allowed, I changed it to use non-wildcard imports so that should be fixed now.

Slightly related: would you accept a PR that prefixes `run_test.sh` and I guess all tests with `dd4hep` (`run_test.sh` -> `dd4hep_run_test.sh`, `test_surface` -> `dd4hep_test_surface`)? Since those are installed to bin, there is potential for confusion and clashes if dd4hep is installed to a shared prefix, as is happening with the lcg views.

DDTest/python/test_import_ddg4.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+moduleNames = [
+    'DDG4',
+    ]
+
+
+@parametrize('moduleName', moduleNames)
+def test_module_ddg4(moduleName):
+    test_module(moduleName)
```
andresailer, line 18:
```suggestion
    test_module('DDG4')
```
and removing all the boilerplate is just as good

PR #856, opened 2021-07-29
CI: disable dev3

andresailer:
CI: disable dev3 until we have time to fix it.


andresailer:
In the end I could make dev3 not "required" so we don't have to remove it.

MarkusFrankATcernch:
Thank you very much! This is very good indeed.

PR #860, opened 2021-08-04
[gdmlFile] optionally import physvol below top level

wdconinc:
This allows syntax like
```xml
  <detectors>
    <detector id="1" name="FT_MT" type="DD4hep_GdmlDetector" vis="InvisibleWithChildren">
      <gdmlFile ref="GdmlDetector.gdml" physvol="lvUpstreamRegion_0"/>
      <parent name="/world"/>
    </detector>
  </detectors>
```
to only load the specific physvol in examples/ClientTests/compact/GdmlDetector.gdml.

The benefit of this is to allow imports of gdml assemblies or volumes into the existing world without having to import a separate gdml world. This can be used to great effect when there would otherwise be overlaps between volumes placed by dd4hep and the world volume in gdml.

Comments welcome.

BEGINRELEASENOTES
- Optionally import gdml physvol below top level, avoiding world

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/gdml/GdmlPlugins.cpp
+++ b/DDCore/src/gdml/GdmlPlugins.cpp
@@ -269,6 +269,7 @@ static Ref_t create_detector(Detector& description, xml_h e, Ref_t /* sens_det *
   string      name    = x_det.nameStr();
   string      par_nam = x_par.nameStr();
   string      gdml   = x_gdml.attr<string>(_U(ref));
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
This looks like a valid enhancement of the GDML plugin.

wdconinc:
Failing check on dev3/x86_64-centos7-clang10-opt currently also fails on master itself.

Issue #862, opened 2021-08-10
DDG4: isotropic gun distribution also forces uniform momentum distribution


wdconinc:
The isotropic gun distributions (e.g. getParticleDirectionUniform, and its siblings) also change the magnitude of the momentum, e.g. in line [Geant4IsotropeGenerator.cpp#L51](https://github.com/AIDASoft/DD4hep/blob/ee51c48c5aaadb0f38d56c7a5b3b8e0c6b890c49/DDG4/src/Geant4IsotropeGenerator.cpp#L51). 

## Steps to reproduce
The following simulation steering file `test.steer`:
```
from DDSim.DD4hepSimulation import DD4hepSimulation
from g4units import mm, GeV, MeV, degree
SIM = DD4hepSimulation()

SIM.gun.energy = 1*GeV
SIM.gun.particle = "e-"
SIM.gun.position = (0.0, 0.0, 0.0)
SIM.gun.direction = (0.0, 0.0, 1.0)
SIM.gun.distribution = "cos(theta)"
SIM.gun.thetaMin = 45*degree
SIM.gun.thetaMax = 135*degree
```
with a simulation that is run as (with a potential change to the dd4hep install path):
```
ddsim -G --steeringFile test.steer --compactFile /opt/local/DDDetectors/compact/SiD.xml --outputFile test.root --numberOfEvents 1000
```
produces a file `test.root`.

## Expected output
The steer file above would seem to indicate a fixed 1*GeV beam energy:
- there is no way to specify an energy range in the command line flags,
- the instructions indicate that `gun.distribution` only changes the direction:
```
  --gun.energy GUN.ENERGY
  --gun.particle GUN.PARTICLE
  --gun.multiplicity GUN.MULTIPLICITY
  --gun.phiMin GUN.PHIMIN
                        Minimal azimuthal angle for random distribution
  --gun.phiMax GUN.PHIMAX
  --gun.thetaMin GUN.THETAMIN
  --gun.thetaMax GUN.THETAMAX
  --gun.direction GUN.DIRECTION
                         direction of the particle gun, 3 vector 
  --gun.distribution {uniform,cos(theta),eta,pseudorapidity,ffbar}
                        choose the distribution of the random direction for theta
                        
                            Options for random distributions:
                        
                            'uniform' is the default distribution, flat in theta
                            'cos(theta)' is flat in cos(theta)
                            'eta', or 'pseudorapidity' is flat in pseudorapity
[...]

PR #872, opened 2021-09-21
MomentumMin and MomentumMax in Geant4IsotropeGenerator

wdconinc:
This PR introduces properties MomentumMin and MomentumMax in the Geant4IsotropeGenerator, so that the isotropic generator can be used with fixed momentum or with a uniform range of momenta. See #862 .

The current behavior (particles are generated with a uniformly distributed momentum between 0 and the specified energy) is maintained with the default settings of MomentumMin (= 0) and MomentumMax (= -1), i.e. when MomentumMax < MomentumMin.

When on the other hand MomentumMax >= MomentumMin, the energy will be uniformly distributed between the two limites.

There are inconsistent user inputs possible which are not captured in the properties approach. A user could set Energy = 1 GeV and MomentumMin = 2 GeV, or MomentumMin = 2 GeV and MomentumMax = 1 GeV. These could be tested for, but since that's not done for ThetaMin and ThetaMax, I've not done it here either.

BEGINRELEASENOTES
- Allow for specifying MomentumMin and MomentumMax in Geant4IsotropeGenerator

ENDRELEASENOTES


andresailer:
As a followup this needs some entries in the ddsim related python classes, so the momentumMin/Max are described there as well.

wdconinc:
I'll put that in a future PR.

PR #880, opened 2021-11-05
WIP: Added the default move constructor to ConditionObject

bencouturier:
Added the default move constructor and default move assignment operator to dd4hep::detail::ConditionObject and dd4hep::NamedObject.
This allows storing ConditionObject in std::vector

BEGINRELEASENOTES
- Added the default move constructor and default move assignment operator to dd4hep::detail::ConditionObject and dd4hep::NamedObject
ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/NamedObject.h
+++ b/DDCore/include/DD4hep/NamedObject.h
@@ -45,11 +45,15 @@ namespace dd4hep {
     NamedObject() = default;
     /// Copy constructor
     NamedObject(const NamedObject& c) = default;
+    /// Move constructor
+    NamedObject(NamedObject&& c) = default;
+
     /// Default destructor
     virtual ~NamedObject() = default;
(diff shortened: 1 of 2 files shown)


MarkusFrankATcernch:
Ben,
I am pretty sure that the tests will succeed. However, I am also very sure this solution will not solve your problem.....
The problem is that the payload cannot be easily be moved. The default move will just copy the byte array,
but the actual contained data there will be deleted and the "moved" instance will have meaningless values
unless these are primitives/scalars.....This will need more rigorous thinking.

bencouturier:
In that case, isn't it up to the deriving classes to implement the move constructor and assignment operator ?

bencouturier:
After our discussion I exp[licitly deleted the move constructor and assignment operators in ConditionObject. I left them in NamedObject though (as the copy constructor there is the default).

Issue #881, opened 2021-11-09
DD4hep fails to build with latest Geant4 11 ( 10.7.ref9)


smuzaffar:
We are trying to build/test new Geant4 11 ( 10.7.Ref9) version for CMS offline software ( https://github.com/cms-sw/cmsdist/pull/7436 ) but looks like currently DD4Hep does not support/build with it. We are getting the compilation errors 9 see below).

 - OS version: SLC7
 - Compiler version: GCC 9.3.0
 - DD4hep version:  v01-18
 - Reproduced by: The build failed with errors like
```
/data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/BUILD/slc7_amd64_gcc900/external/dd4hep/v01-18x-b0ba8d9db156bba37a2d9b5cfd320054/dd4hep-v01-18x/DDG4/src/Geant4Converter.cpp:151:10: error: 'kFASTCOMPONENT' was not declared in this scope; did you mean 'kWLSCOMPONENT'?
  151 |     case kFASTCOMPONENT:                  return make_pair(CLHEP::keV/units::keV, 1.0);
      |          ^~~~~~~~~~~~~~
      |          kWLSCOMPONENT
/data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/BUILD/slc7_amd64_gcc900/external/dd4hep/v01-18x-b0ba8d9db156bba37a2d9b5cfd320054/dd4hep-v01-18x/DDG4/src/Geant4Converter.cpp:152:10: error: 'kSLOWCOMPONENT' was not declared in this scope; did you mean 'kWLSCOMPONENT'?
  152 |     case kSLOWCOMPONENT:                  return make_pair(CLHEP::keV/units::keV, 1.0);
      |          ^~~~~~~~~~~~~~
      |          kWLSCOMPONENT

```
see the full buil log at https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-b6c16f/20363/external/dd4hep/v01-18x-b0ba8d9db156bba37a2d9b5cfd320054/log or https://muzaffar.web.cern.ch/dd4hep-geant11.txt
 - Goal: We would like to test new Geant4 in CMS Offline software and for that we need to build DD4Hep too.

MarkusFrankATcernch:
@smuzaffar https://github.com/AIDASoft/DD4hep/pull/882

smuzaffar:
[...]

PR #884, opened 2021-11-22
Fix for CMS G4 11 beta prerelease ( Issue https://github.com/AIDASoft…

MarkusFrankATcernch:
BEGINRELEASENOTES
- Issue https://github.com/AIDASoft/DD4hep/issues/881
  Next attempt for resolution. Avoid explicit use of the second argument in 
```
  G4int GetConstPropertyIndex(const G4String& key,
                              G4bool warning = false) const;
  // Get the constant property index from the key-name

  G4int GetPropertyIndex(const G4String& key, G4bool warning = false) const;
  // Get the property index by the key-name.
  ```
  Should satisfy existing and future snapshots of Geant4.
ENDRELEASENOTES


smuzaffar:
@MarkusFrankATcernch , can you please also fix https://github.com/AIDASoft/DD4hep/blob/master/DDG4/src/Geant4Converter.cpp#L1200 ?

PR #891, opened 2022-02-24
Fix warnings

### LINKED ISSUE #844: [CMS] Certain warnings not needed because they report intended behavior
As used by CMS, DD4hep issues these warnings and info messages:
```
Evaluator        WARN  +++ Overwriting variable: name=world_x value=101*m  Evaluator::Object : existing variable [setVariable error]
Evaluator        WARN  +++ Overwriting variable: name=world_y value=101*m  Evaluator::Object : existing variable [setVariable error]
Evaluator        WARN  +++ Overwriting variable: name=world_z value=450*m  Evaluator::Object : existing variable [setVariable error]
DD4hep           WARN  ++ Using globally Geant4 unit system (mm,ns,MeV)

PlacedVolume     INFO  REFLECTION: (x.Cross(y)).Dot(z):       -1 Parent: eregalgo:EFAW [TGeoVolume] Daughter: eregalgo:EHAWR [TGeoVolumeAssembly]
```
These messages reported intended behavior. The world variables need to be overwritten for use by CMS, CMS wants to use the Geant4 unit system, and the EFAW assembly should be reflected. In these cases, only a debug message would be needed, not a warning or info. Of course, in other cases, warnings should still appear, just not for intended behavior.

See here for the source of the overwriting warning:
https://github.com/AIDASoft/DD4hep/blob/9bbe12b080ce52615251facc6f6ff5dd58a60e69/DDParsers/src/Evaluator/Evaluator.cpp#L650

As CMS moves to production use of DD4hep, it would be nice to remove unnecessary warnings from our logs.
 
These messages have been appearing for a long time and are still present with version 1.17.0.

andresailer:
BEGINRELEASENOTES
- DDParsers: do not use deprecated boost headers with boost 1.78
- DDG4 Particle.tbl: drop some diffr(active) particles causing warnings with Geant4 11.0
- DetectorIMP: Changing message about Geant4 unit system to INFO, fixes #844
- Volumes: Changing message about REFLECTION to DEBUG, fixes #844

ENDRELEASENOTES


DDParsers/include/Parsers/spirit/GrammarsV2.h
```diff
@@ -33,9 +33,19 @@
 #include <boost/fusion/include/unused.hpp>
 #include <boost/fusion/include/std_pair.hpp>
 
+// version is like X YYY ZZ (no spaces
```
andresailer, line 36:
```suggestion
// version is like X YYY ZZ (no spaces)
```

PR #898, opened 2022-03-24
Reduce coverity warnings

MarkusFrankATcernch:
BEGINRELEASENOTES
- RTeduce coverity warnings
ENDRELEASENOTES


andresailer:
We have issues with the macos builds, I strongly suspect these changes are not breaking  the macos builds, so merge

PR #899, opened 2022-03-28
cmake: quote an argument to avoid error with CMake > 3.18

vvolkl:
When using CMake 3.19 or newer and no CMAKE_CXX_STANDARD has been set before the call to find_package(dd4hep), downstream code will error with:
```
CMake Error at /afs/cern.ch/user/v/vavolkl/cmaketest/tmp/CMakeFiles/CMakeTmp/CMakeLists.txt:13 (add_executable):
  CXX_STANDARD is set to invalid value 'CXX_EXTENSIONS'


CMake Error at CMakeLists.txt:3 (try_compile):
  Failed to generate test project build system.
```
because ${CMAKE_CXX_STANDARD} is empty. I couldn't find any relevant information in the changelog, so this might be a bug in CMake proper. Quoting the argument fixes this in newer dd4hep versions.

BEGINRELEASENOTES
- cmake: DD4hepConfig: Drop unnecessary call to DD4HEP_SETUP_BOOST_TARGETS

ENDRELEASENOTES


andresailer:
But this could result in testing with a different standard (the compiler's default) than finally building, which wouldn't be good.

vvolkl:
Yes, but that is the case already, cmake just seems to swallow the error before version 3.19. 

What about adding a warning if CMAKE_CXX_STANDARD is empty?

cmake/DD4hepBuild.cmake
```diff
@@ -779,7 +779,7 @@ MACRO(DD4HEP_SETUP_BOOST_TARGETS)
   FOREACH(FS_LIB_NAME stdc++fs "" c++fs )
     dd4hep_debug("|++++> linking against ${FS_LIB_NAME}")
     try_compile(HAVE_FILESYSTEM ${CMAKE_BINARY_DIR}/try ${DD4hep_DIR}/cmake/TryFileSystem.cpp
-      CXX_STANDARD ${CMAKE_CXX_STANDARD}
+      CXX_STANDARD "${CMAKE_CXX_STANDARD}"
```
andresailer, line 782: Please remove this change, in case we ever re-arrange the dd4hep cmake we should error out if we don't set the standard before calling this.

andresailer:
Thanks!

PR #903, opened 2022-04-04
edm4hep: Put event and run number into event metadata

tmadlener:
BEGINRELEASENOTES
- Add `eventNumber` and `runNumber` into the event metadata for edm4hep output

ENDRELEASENOTES


andresailer (approved):

tmadlener:
Good question about the event header. Looks like that should have all the information that is necessary. Let me clarify if that is actually the case and if so, I would close this again.

andresailer:
I was thinking that we rather drop the eventHeader, and store this info in the metadata, let's discuss tomorrow in the EDM4hep meeting.

tmadlener:
As discussed during the meeting today: Keep this in the `EventHeader` for now and reassess later, when we might have a more appropriate place to store this in the Frame.

PR #910, opened 2022-05-09
Prevent crash at load time if LD_LIBRARY_PATH is not set

pikacic:
I found in Gaudi a logic error in the way the plugin service initializes the factories search path (see https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/1316) and the issue is present in DD4hep too.

The problem is that the search path string is initialized from the return value of [`std::getenv`](https://en.cppreference.com/w/cpp/utility/program/getenv), which is `nullptr` if the environment variable is not set. Initializing a string from `nullptr` might result in an exception (not caught) or in a segfault (depending on the implementation).

BEGINRELEASENOTES
- Plugin Service: check that the `LD_LIBRARY_PATH` (or `PATH` or `DYLD_LIBRAY_PATH`) are actually set before trying to use them (to avoid a crash)

ENDRELEASENOTES


andresailer (approved):
Thanks!

PR #911, opened 2022-05-17
Add VariantParameters extension + plugin

paulgessinger:
This PR is a proposal for a general purpose parameter map extension that can be attached to any detector element.
It uses `boost::variant` to implement a map with string keys and values of any of string, int, bool, double. 

Also added is a plugin which takes varying arguments: the first one is a name argument which the plugin recursively tries to find (is there a better way to find an element by name?). It then goes through the rest of the arguments which need to take the form: `key[: {int,string,bool,double}] = value`, where the type of the argument can be specified after the colon. If omitted, the type defaults to string. In the XML, this then looks like:

```xml
<plugins>
  <plugin name="DD4hep_ParametersPlugin">
    <argument value="PixelEndcapN"/>
    <argument value="layer_pattern: str=PixelEndcapN\d|PixelEndplate"/>
  </plugin>	
  <plugin name="DD4hep_ParametersPlugin">
    <argument value="PixelEndcapP"/>
    <argument value="layer_pattern=PixelEndcapP\d|PixelEndplate"/>
  </plugin>	
  <plugin name="DD4hep_ParametersPlugin">
    <argument value="PixelBarrel"/>
    <argument value="layer_pattern=PixelLayer\d"/>
  </plugin>	
</plugins>
```

Do you have any thoughts? @andresailer @vvolkl @andiwand @asalzburger

BEGINRELEASENOTES
- Add `VariantParameters` extension + plugin to assign parameters to it by name.

ENDRELEASENOTES

### DIFF
--- a/DDRec/src/plugins/addVariantParameters.cpp
+++ b/DDRec/src/plugins/addVariantParameters.cpp
@@ -89,7 +89,7 @@ static long addVariantParameters(Detector& description, int argc, char** argv) {
       extension->variantParameters[key] = value;
     }
     else if (type == "double") {
-      extension->variantParameters[key] = dd4hep:_toDouble(value);
+      extension->variantParameters[key] = dd4hep::_toDouble(value);
     }
     else if (type == "int") {
       extension->variantParameters[key] = dd4hep::_toInt(value);
--- a/DDRec/src/plugins/addVariantParameters.cpp
+++ b/DDRec/src/plugins/addVariantParameters.cpp
@@ -92,7 +92,7 @@ static long addVariantParameters(Detector& description, int argc, char** argv) {
       extension->variantParameters[key] = dd4hep:_toDouble(value);
     }
     else if (type == "int") {
-      extension->variantParameters[key] = std::stoi(value);
+      extension->variantParameters[key] = dd4hep::_toInt(value);
     }
     else if (type == "bool") {
       if (value == "true") {
--- a/DDRec/src/plugins/addVariantParameters.cpp
+++ b/DDRec/src/plugins/addVariantParameters.cpp
@@ -1,0 +1,112 @@
+//==========================================================================
+//  AIDA Detector description implementation 
+//--------------------------------------------------------------------------
+// Copyright (C) Organisation europeenne pour la Recherche nucleaire (CERN)
+// All rights reserved.
+//
+// For the licensing terms see $DD4hepINSTALL/LICENSE.
+// For the list of contributors see $DD4hepINSTALL/doc/CREDITS.
+//
+// Author     : P.Gessinger
+//
+//==========================================================================
+#include "DD4hep/Detector.h"
+#include "DD4hep/Factories.h"
+#include "DD4hep/Printout.h"
+#include "DD4hep/DetectorTools.h"
+
+#include "DDRec/DetectorData.h"
+
+#include <regex>
+
+
+namespace dd4hep::rec {
+
+static long addVariantParameters(Detector& description, int argc, char** argv) {
+
+  if(argc < 2) {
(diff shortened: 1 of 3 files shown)


DDRec/src/plugins/addVariantParameters.cpp
```diff
@@ -0,0 +1,112 @@
...
+    auto recurse = [detector](auto&& self, DetElement elt) -> DetElement {
+      if(elt.name() == detector) { return elt; }
+      for(auto& child : elt.children()) {
+        DetElement res = self(self, child.second);
+        if (res.isValid()) { return res; }
+      }
+      return DetElement{}; // not found
+    };
+
+    DetElement elt = recurse(recurse, description.world());
```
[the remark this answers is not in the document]
paulgessinger: I'll try that. An exception is probably fine here, right now I just return 0 from the plugin function if it's not found.

andresailer:
Apart from some missing documentation this looks all reasonable to me.

paulgessinger:
Thanks @andresailer. I switched to `description.detector()` (which seems to work), and added doxygen comments. I'm not very familiar with the format you use for these, can you advise if I should put specific annotations on the plugin function?

DDRec/src/plugins/addVariantParameters.cpp
```diff
@@ -0,0 +1,120 @@
...
+    if(extension == nullptr) {
+      extension = new VariantParameters();
+      elt.addExtension<VariantParameters>(extension);
+    }
+
+    if (type == "str" || type == "string") {
+      extension->variantParameters[key] = value;
+    }
+    else if (type == "double") {
+      extension->variantParameters[key] = std::stod(value);
```
andresailer, line 92: I think if we switch to the dd4hep parsing functions, we can use things like units, constants, whatever is known to the grammar, in the values? Or is that already working?

```suggestion
      extension->variantParameters[key] = dd4hep:_toDouble(value);
```
https://dd4hep.web.cern.ch/dd4hep/reference/group__DD4HEP__CORE.html#gab6fd7f4fe0ac9fa2c4f41132be909103

DDRec/src/plugins/addVariantParameters.cpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+    else if (type == "int") {
+      extension->variantParameters[key] = std::stoi(value);
```
andresailer, line 95:
```suggestion
      extension->variantParameters[key] = dd4hep::_toInt(value);
```
https://dd4hep.web.cern.ch/dd4hep/reference/group__DD4HEP__CORE.html#gac41662b126182e5004a5fdbc8304d770

DDRec/src/plugins/addVariantParameters.cpp
paulgessinger: Ah I didn't consider using the full parsing functions. I can look into that though. Does the evaluator support anything other than double results, though? Reading the code it seems to me like it's restricted to double, but here we'd probably at least support string and int values.

andresailer:
Ahh, double colons

DDRec/src/plugins/addVariantParameters.cpp
andresailer: yes, keeping the `type: value`, but going through the dd4hep parsing for ints and doubles.
Not sure it would make sense for strings

paulgessinger: Ah right I see, I didn't realize `_toDouble` and `_toInt` already run the evaluator.

This should be good then, right?

andresailer:
Yes, I think all good now.

andresailer (approved):

PR #917, opened 2022-06-09
Improve error reporting if derived condition dependencies cannot be resolved 

MarkusFrankATcernch:
BEGINRELEASENOTES
-- Improve error reporting if derived condition dependencies cannot be resolved 
-- Add illustrating example

ENDRELEASENOTES


andresailer:
@MarkusFrankATcernch Should this be merged?

PR #933, opened 2022-07-19
Improved GFlash and Geant4 fast simulation handling

MarkusFrankATcernch:
BEGINRELEASENOTES
The fast simulation handling in Geant4 includes the GFlash interface
as implemented in `<geant40-source>/source/parameterisations/gflash`
and a fast simulation interface as provided by 
`<geant40-source>/source/processes/parameterisation`

This MR allows to handle both mechanisms using the same callbacks 
in the Geant4SensitiveActions using the callback specialization:
```
      /// GFLASH/FastSim interface: Method for generating hit(s) using the information of the fast simulation spot object.
      virtual bool processFastSim(const Geant4FastSimSpot* spot, G4TouchableHistory* history)  final;
```
in addition to the regular callback for full simulation:
```
      /// G4VSensitiveDetector interface: Method for generating hit(s) using the G4Step object.
      virtual bool process(const G4Step* step,G4TouchableHistory* history)  final;
```
The `Geant4FastSimSpot` is inspired by the `G4GFlashSpot`, but avoids the specialization towards GFlash
given that the `G4FastHit` and the `G4GFlashEnergySpot` are equivalent.
Together with the helper class `Geant4FastSimHandler` simple sensitive detector callbacks
can be implemented like:
```
    /// GFlash/FastSim interface: Method for generating hit(s) using the information of Geant4FastSimSpot object.
    template <> bool
    Geant4SensitiveAction<Geant4Tracker>::processFastSim(const Geant4FastSimSpot* spot,
							 G4TouchableHistory* /* hist */)
    {
      typedef Geant4Tracker::Hit Hit;
      Geant4FastSimHandler h(spot);
      Hit* hit = new Hit(h.trkID(), h.trkPdgID(), h.deposit(), h.track->GetGlobalTime());
      hit->cellID        = cellID(h.touchable(), h.avgPositionG4());
      hit->energyDeposit = h.deposit();
      hit->position      = h.avgPosition();
      hit->momentum      = h.momentum();
      hit->length        = 0e0;
      collection(m_collectionID)->add(hit);
      return true;
    }
```
Callbacks with this signature are called both for GFlash and the G4 fast simulation.

To simplify the user defined construction of fast simulation shower models
the helper class `Geant4FastSimShowerModel` is implemented as a `Geant4Action`
and hence allows to have options to be set by the user (including python).

As an illustration two fast simulation shower models were implemented in
`<dd4hep>/DDG4/plugins/Geant4P1ShowerModel.cpp` which were directly deduced
from the Geant4 example `<geant4-source>/examples/extended/parameterisations/Par01`.
Please see the source files for details.

To execute GFlash or fast simulation to simple python examples are provided
using simple silicon blocks as sensitive volumes:
- GFlash `<dd4hep-dir>/examples/ClientTests/scripts/SiliconBlockGFlash.py`
- Fast simulation `<dd4hep-dir>/examples/ClientTests/scripts/SiliconBlockFastSim.py`

To enable either the following actions are required:
1. Enable fast simulation for certain particles in the physics list:
```  # Build the physics list:
  phys = geant4.setupPhysics('FTFP_BERT')
  ph = DDG4.PhysicsList(kernel, str('Geant4FastPhysics/FastPhysicsList'))
  ph.EnabledParticles = ['e+', 'e-']
  ph.enableUI()
  phys.adopt(ph)
```
This step is identical for GFlash and fast simulation physics.
2. Enable the shower mode:
```  model = DDG4.DetectorConstruction(kernel, str('Geant4Par01EMShowerModel/ShowerModel'))
  # Mandatory model parameters
  model.RegionName = 'SiRegion'
  model.Material = 'Silicon'
  model.ApplicableParticles = ['e+', 'e-']
  model.Etrigger = {'e+': 0.1 * GeV, 'e-': 0.1 * GeV}
  model.Enable = True
  model.enableUI()
```
Here the options must be set according to the parameters required by the shower models.

The G4 fast simulation interface is only functional for Geant4 version >= 10.07.
For lower versions of Geant4 missing classes are stubbed to allow the compilation.

ENDRELEASENOTES


MarkusFrankATcernch:
@andresailer In the Geant4Sensitives: Is it a good idea to have G4Step as a "const" object?
I have not found any occurrence where this would be violated, but certain calls of the G4Step require write access
like e.g. `SetNonIonizingEnergyDeposit` or `ClearLastStepFlag` but I think this is only called by Geant4 internally.

andresailer:
@MarkusFrankATcernch If it compiles without a const_cast, I don't see a reason not to leave it const. The SensitiveDetectors shouldn't modify the G4Step.

MarkusFrankATcernch:
@engineren
This MR is better to have fast sumlation in dd4hep using gflash or user defined shower model.
You are welcom to join the next DD4hep developers meeting.

EnginEren:
Hi @MarkusFrankATcernch, thanks. I'll let you know when I have updates

MarkusFrankATcernch:
Hmm. I have also built against LCG 101. I get the same output, but it seems to build:
```
 > which cmake
/cvmfs/sft.cern.ch/lcg/views/LCG_101/x86_64-centos7-gcc11-dbg/bin/cmake
 > geant4-config --version
10.7.2
 > cmake --version
cmake version 3.20.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

MarkusFrankATcernch:
I continue then aftwards simply like this:
```
    source ${DD4hepINSTALL}/bin/thisdd4hep.sh;
    mkdir examples;
    export DD4hepExamplesINSTALL=......
    cd examples;
    cmake ${CMAKE_COMPILER_CHOICE} \
	  -DDD4HEP_USE_GEANT4=ON \
	  -DBoost_NO_BOOST_CMAKE=ON \
	  -DDD4HEP_USE_LCIO=OFF \
	  -DDD4HEP_USE_TBB=OFF \
	  -DBUILD_TESTING=ON \
	  -DROOT_DIR=$ROOTSYS \
	  -DDD4hep_DIR=${DD4hepINSTALL}  \
	  -DCMAKE_INSTALL_PREFIX=${DD4hepExamplesINSTALL}  \
	  -DCMAKE_BUILD_TYPE=Debug \
	  ../../DD4hep/examples;
    make -j 22;
    make -j 12 install;
```
This works fine for me.

wdconinc:
> The SensitiveDetectors shouldn't modify the G4Step.

Some sensitive detectors, e.g. Geant4OpticalCalorimeterAction, kill the incident optical photon and absorb the energy completely. See e.g.
https://github.com/AIDASoft/DD4hep/blob/master/DDG4/plugins/Geant4SDActions.cpp#L312
The only reason why I think this doesn't fail on compilation is because Geant4OpticalCalorimeterAction is not used anywhere and probably no compilation is even attempted for this template specialization.

This is failing in our EIC code for the exact same reason (but in a PMT detector SD action that absorbs the optical photon).

PR #945, opened 2022-08-03
ddsim: remove `--gun.energy` option (as confusing alias for `--gun.momentumMax`)

veprbl:
BEGINRELEASENOTES
- Removed `--gun.energy` option from ddsim. The `--gun.momentumMax` should be used instead.
ENDRELEASENOTES

The issue is that, currently, `--gun.energy x` generates particles in the range [0, x], which is a bit surprising.

I was getting a confusing results from using this option (imagine getting a flat spectrum from what is supposed to be a fixed energy gun). It doesn't do what the name suggests and there is no help message to clarify that. I propose to remove it.

### DIFF
--- a/DDG4/python/DDSim/Helper/Gun.py
+++ b/DDG4/python/DDSim/Helper/Gun.py
@@ -14,7 +14,6 @@ class Gun(ConfigHelper):
 
   def __init__(self):
     super(Gun, self).__init__()
-    self.energy = 10 * GeV
     self.particle = "mu-"
     self.multiplicity = 1
     self._position = (0.0, 0.0, 0.0)
@@ -28,7 +27,7 @@ class Gun(ConfigHelper):
     self.thetaMax = None
     self._momentumMin_EXTRA = {'help': "Minimal momentum when using distribution (default = 0.0)"}
     self.momentumMin = None
-    self.momentumMax = None
+    self.momentumMax = 10 * GeV
 
     self._distribution_EXTRA = {'choices': ['uniform', 'cos(theta)',
                                             'eta', 'pseudorapidity',
@@ -114,9 +113,6 @@ class Gun(ConfigHelper):
   def setOptions(self, ddg4Gun):
     """set the starting properties of the DDG4 particle gun"""
     try:
-      ddg4Gun.energy = self.energy  # ddg4Gun.energy actually sets momentum
-      ddg4Gun.MomentumMin = 0.0
-      ddg4Gun.MomentumMax = self.energy
       ddg4Gun.particle = self.particle
       ddg4Gun.multiplicity = self.multiplicity
       ddg4Gun.position = self.position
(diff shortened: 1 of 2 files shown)


DDG4/python/DDSim/Helper/Gun.py
```diff
@@ -114,9 +113,6 @@ def position(self, val):
   def setOptions(self, ddg4Gun):
     """set the starting properties of the DDG4 particle gun"""
     try:
-      ddg4Gun.energy = self.energy  # ddg4Gun.energy actually sets momentum
```
veprbl, line 117: I'm starting to doubt this. It seems like `DDG4.GeneratorAction(kernel, "Geant4ParticleGun/" + "Gun").energy` should may have been useful together with some other option.

An alternative to this PR is to revert the confusing behaviour introduced in 91683a6fe8c18ae8d882499d8ebf4db67d7b684d (cc @wdconinc).

wdconinc: Let me express my full support for not having a default that throws over a range from 0 to `energy` when the user specifies a single `energy` value.

andresailer: I agree, this is more trouble and a real stumbling block.
What do you think about keeping `energy` as an alias for momentumMax (via `@property`)

wdconinc: The reasonable (for high energy physics) approximation of using 'energy' interchangeably for total energy, kinetic energy, and momentum in the context of geant4 simulations really breaks down when in our EIC environment we study e.g. lower limit 150 MeV protons in our tracker (the 150 MeV in this context is momentum).

I personally think that 'energy' is confusing and should not be used without explicit 'total' or 'kinetic'. When momentum min and max setters were introduced, the energy setter was only kept for exact backwards compatibility, not because I'd suggest anyone to use it :-) If there's a way to phase it out, I'd wholeheartedly support that :-)

andresailer: Let's add a "deprecated" warning then, and drop it soon after that.

andresailer: I guess what I wanted to have is a simple way to set mono-energetic particle.
And that used to be the case, but this was eventually (inadvertently) changed https://github.com/AIDASoft/DD4hep/pull/734/files#diff-e6b2229d02a03edb64bc49337cda4bf87560460d23790cc468958479b8e9c72e

I think we need to better define what we want the behavior to be...

Keeping (restoring) the possibility to set the mono-energetic sounds rather more attractive from my point of view.

andresailer: So I restored the behaviour in #949 added some doc text etc. Let me know what you think...

andresailer:
Superseded by #949

PR #946, opened 2022-08-04
CI: change the software stacks used, add output grouping

andresailer:
BEGINRELEASENOTES
- CI changes, for discussion

ENDRELEASENOTES

- [x] Add a build based on key4hep stack
- [x] disable EDM4hep in lcg stack based builds


andresailer:
dev4 x86_64-centos7-gcc11-opt utterly failed to be installed on cvmfs today, I hope it will be better tomorrow

andresailer:
Something still goes wrong for the g4units case on dev3 (ROOT Master branch)

```
   [surface] ##################### TEST_FAILED ######################
  [surface] ### ERROR:      MaterialData.radiationLength() == 93.4961 * mm  : [0.936612] != [93.6612]
  [surface] ########################################################
  [surface] test PASSED:       MaterialData.interactionLength() == 457.532 * mm  : [457.533] == [457.533]
```

These values should just be taken from the XML file for the test, no calculations performed?

https://github.com/AIDASoft/DD4hep/blob/7cbae31f15835815374bb1ce0b6e744a21b6417d/DDTest/units.xml#L51-L56

andresailer:
Couple checks:
* with DD4HEP_USE_GEANT4_UNITS=ON dd4hep::mm is equal to 1
* the ROOT from dev3 does include the merged PR
* RL and NIL from the xml are ignored

andresailer:
The unit issue should be fixed by https://github.com/root-project/root/pull/11115

andresailer:
The key4hep stack is using a different assimp library than lcg stack, and there seems to be an issue with collada based meshes being loaded by that new library.

MarkusFrankATcernch:
@andresailer There is not a lot one can do...The number of vertices somehow differ. 
This may also be some numerical discrepancy, where TGeo inserts an additional facet since the number of vertices only differ by 3.

andresailer:
@MarkusFrankATcernch I was going to ignore the duck test for the key4hep stacks

vvolkl:
I can also change the version of assimp in the stack, but it does sound like the test should not fail if the discrepancy is not an actual issue.

andresailer:
I will try to draw the duck, and see if it still quacks like :duck:

andresailer:
The duck still looks the same, and the difference, apart from the order of what is printed is just three entries with identical numbers to three other entries.

PR #950, opened 2022-08-10
Support for channeling physics in Geant4

MarkusFrankATcernch:
BEGINRELEASENOTES
On request from out LHC colleagues DD4hep supports channeling physics in Geant4.
To support channeling physics in Geant4 it is necessary to enable the transparent 
creation of G4ExtendedMaterial and G4LogicalCrystalVolume instances.
This can be done in DD4hep with the supply of material and volume properties in DD4hep.

Material properties can be supplied in the compact description like here:
```
    <material name="Ice">
      <D type="density" value="1.0" unit="g/cm3"/>
      <composite n="2" ref="H"/>
      <composite n="1" ref="O"/>
      <constant name="Geant4-plugin:material"            option="ChannelingCrystalMaterial"/>
      <constant name="Geant4-ignore:crystal_data"        option="${DD4hepExamplesINSTALL}/examples/DDG4/data/Si220pl/Si220pl"/>
      <constant name="Geant4-ignore:crystal_orientation" option="(1,0,0)"/>
    </material>
```
The property starting with the tag `Geant4-plugin` is used to determine the proper plugin
preparing the G4ExtendedMaterial instance.
All properties with tags `Geant4-ignore` are not passed as material properties to Geant4.
Properties starting with this tag can be used by users to configure the G4ExtendedMaterial instance.

Volumes can as well have properties as this C++ example shows:
```
  Volume vol = Volume("Volume", solid, material);
  xml_elt_t ec = ...;
  vol.addProperty(ec.attr<std::string>(_U(name)), ec.attr<std::string>(_U(value)));
```
with values supplied by XML:
```
  <property name="Geant4-plugin" value="ChannelingCrystalVolume"/>
```
For more details see the example `<DD4hep>/examples/DDG4/compact/Channeling.xml`.
Again the property `Geant4-plugin` is used to call a plugin to create sub-classes
of G4LogicalVolume like G4LogicalCrystalVolume. 

An example was prepared in `<DD4hep>/examples/DDG4` containing:
- a compact description: `compact/Channeling.xml`
- an example factory for the G4ExtendedMaterial: `src/ChannelingCrystalMaterial.cpp`
- an example factory for the G4LogicalCrystalVolume: `src/ChannelingCrystalVolume.cpp`
- A script invoking Geant4 to test this setup: `scripts/Channeling.py`

ENDRELEASENOTES


chiaramaccani:
Dear Markus,
if I want to apply your example code to another project should I copy the factory for the G4ExtendedMaterial ( (src/ChannelingCrystalMaterial.cpp) and the factory for G4LogicalCrystalVolume (src/ChannelingCrystalVolume.cpp) to my source directory, right?
should I also add somewhere the path to Geant4 in order to make the compiler find the G4 include file?
Thank you very much,
Chiara

MarkusFrankATcernch:
If you n4eed to change the classes/factories, not only copy them, but also rename the classes/factories. Not that by some accident you have to deal with name clashes and unconsciously execute the wrong code.
You must have already Geant4 to build DD4hep and unless you use a stone age version all required includes should be present.
Have a look on the dependencies in the example you will need the same for your channeling Geant4 module.

chiaramaccani:
ok thank you.
My problem is that if I just copy the two files and compile with cmake, when I run make, the compiler shows the error 

/afs/cern.ch/user/c/cmaccani/TestChanneling/src/ChannelingCrystalVolume.cpp:17:10: fatal error: G4ExtendedMaterial.hh: No such file or directory
   17 | #include <G4ExtendedMaterial.hh>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.

like it couldn't be able to find G4 include files anymore...

MarkusFrankATcernch:
When you build your library you will have a stetement like this:
```
  dd4hep_add_plugin(MyChannellingLib
    SOURCES   src/*.cpp
    USES      DD4hep::DDCore DD4hep::DDG4 Geant4::Interface ROOT::Core ROOT::Geom ROOT::GenVector ROOT::RIO
    )
```
The dependecy on Geant4::Interface will add the proper include path.

chiaramaccani:
Dear Markus Frank,
I think there could be a small mistake in the code.
At line 67 in ChannelingCrystalMaterial.cpp it is written 
```
orientation = mat.propertyRef("Geant4-ignore:crystal_orientation");
```

but since orentation is a constant property I think it should be
```
orientation = mat.constPropertyRef("Geant4-ignore:crystal_orientation");
```

Fixing this, it works as expected also in my application.
Thank you very much.

andresailer:
Thanks @chiaramaccani 
I added your fix in #963

Issue #913, opened 2022-05-23
XML parsing with non-standard linux-locale


benjaminhuth:
I'm not sure if this is really a bug, but certainly a unexpected behavior at least for me: 

I tried to parse a DD4hep XML-file (this one specifically, though it shouldn't matter for the issue: https://gitlab.cern.ch/acts/OpenDataDetector/-/blob/main/xml/OpenDataDetector.xml) with the ACTS DD4hep interface, and got the following error:.

`RuntimeError: dd4hep: Evaluator::Object : syntax error : value=10.0*m [Evaluation error]`

After quite some time of debugging I found the error in the configured locale, where the `LC_NUMERIC` entry was set to `de_DE.UTF-8`, which caused the failure of the floating point parsing (it expected a comma). Changing this value to the default `C`-locale resolved the error.

From my perspective the dependence on the configured locale for the parsing causes hard to debug errors and should be fixed (or at least checked with a warning).

 - OS version: e.g. Fedora 5.17.6-200.fc35.x86_64
 - Compiler version: GCC 11.3.1
 - DD4hep version: v01-17

MarkusFrankATcernch:
After quite some investigations:
Cannot be fixed if the old locales in C and C++ should be restored.

PR #954, opened 2022-08-11
Set locale to C before parsing XML files

MarkusFrankATcernch:
BEGINRELEASENOTES
Before starting to parse XML files, programatically set the locale to "C" if any of the three
- LC_NUMERIC,
- LC_TIME,
- LC_CTYPE
Is not set to "C".

Addresses issue https://github.com/AIDASoft/DD4hep/issues/913.
ENDRELEASENOTES

### DIFF
--- a/DDCore/src/DetectorLoad.cpp
+++ b/DDCore/src/DetectorLoad.cpp
@@ -21,6 +21,7 @@
 // C/C++ include files
 #include <stdexcept>
 #include <clocale>
+#include <cstring>
 
 #ifndef __TIXML__
 #include "xercesc/dom/DOMException.hpp"
--- a/DDCore/src/DetectorLoad.cpp
+++ b/DDCore/src/DetectorLoad.cpp
@@ -13,7 +13,6 @@
 
 // Framework include files
 #include "DD4hep/DetectorLoad.h"
-#include "DD4hep/Detector.h"
 #include "DD4hep/Printout.h"
 #include "DD4hep/Plugins.h"
 #include "XML/XMLElements.h"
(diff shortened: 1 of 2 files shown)


veprbl:
Since DD4hep is a library, it would be nice to also restore the locale after you are done with it.

MarkusFrankATcernch:
@veprbl You have a point here. I have think how to do this in some elegant way.

veprbl:
I was fixing a similar issue for YODA. Here is how it went:
https://gitlab.com/hepcedar/yoda/-/commit/33437490c9101e8f6fa4aaa12e032ee91e95e32f

MarkusFrankATcernch:
@veprbl I think this approach does not work. I can only switch between full locales outside of streams. All internal stuff is then happening e.g. in XercesC etc....I reverted.

veprbl:
There is also a routine to change and restore the locales, it seems to be portable enough to at least compile with gcc and for macOS.

MarkusFrankATcernch:
@veprbl Naaaa....look, this is what happens. Normally a class like this could guard all major XML parsing blocks:
```
    std::locale locale_current  {""};  // OS locale, NOT program locale
    std::locale locale_C {"C"};  // C locale
    TemporaryLocale()   {                        <<< Locale is program locale
      std::locale::global(locale_C);           <<< Locale becomes C
    }
    ~TemporaryLocale()   {
      std::locale::global(this->locale_current);   <<<< Locale becomes OS locale
    }                                                                   <<<< locale is no longer the same as before the constructor.
```
There is no call to get the current program locale. Hence, there is no simple way to construct such a guard.

veprbl:
I don't remember all the details, but I vaguely remember that, indeed, you can't do a lot with locales from the C++ standard library. But looking at my commit, it seems like the `uselocale` function return a handler for a previous locale.

PR #967, opened 2022-08-30
feat: Geant4OpticalTrackerAction which stops-and-kills optical photons

wdconinc:
In the EIC detectors we use various photo-sensitive pixel planes in the DIRC and RICH detectors. This is similar to a tracker in the overall SD action sense, but instead of requiring optical photons energy deposition in the material of the sensor (e.g. SiPM, LAPPD, AC-LGAD), we often don't model the optical properties of the materials in the sensors and simply stop-and-kill the optical photon when it reaches the sensor. This is an SD action that achieves this.

I have no geometry to easily test the `processFastSim` functionality with.

BEGINRELEASENOTES
- Geant4OpticalTrackerAction which stops-and-kills optical photons for e.g. SiPM sensor planes

ENDRELEASENOTES

### DIFF
--- a/DDG4/plugins/Geant4SDActions.cpp
+++ b/DDG4/plugins/Geant4SDActions.cpp
@@ -144,6 +144,90 @@ namespace dd4hep {
     typedef Geant4SensitiveAction<Geant4Tracker> Geant4TrackerAction;
 
     // ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+    //               Geant4SensitiveAction<Geant4PhotonCounter>
+    // ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+    /** \addtogroup Geant4SDActionPlugin
+     *
+     * @{
+     * \package Geant4OpticalTrackerAction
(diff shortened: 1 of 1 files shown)


andresailer:
Hi Wouter,
Do you have a simple detector that we could add to the examples to use this plugin?

wdconinc:
> simple detector

Well... Probably not an entire RICH? Though maybe @c-dilks would be open to turning our dual RICH into a standalone example that demonstrates this. Otherwise we can also do our proximity-focused RICH which is a bit simpler.

c-dilks:
Hi @wdconinc, I opened https://github.com/wdconinc/DD4hep/pull/2 targeting your fork's branch, which adds a PFRICH. It needs some more work to make it nice, but I think I see the increase in the number of hits. See the description within for more details.

wdconinc:
> Do you have a simple detector that we could add to the examples to use this plugin?

Example added as #974. Not sure if it's what you'd consider 'simple' but it is at least much simpler than how we are using it in production :-D

So, I'd suggest either merging this and moving example discussion to the example PR, or closing this and merging the tracker SD action as part of the example PR.

andresailer:
Thanks! I will merge this, and then #974 can be rebased to add the example.

PR #968, opened 2022-08-30
feat: support `--runType qt` in `ddsim`

wdconinc:
This adds a run type `qt` for `ddsim`. Sometimes it's convenient to be able to open the Qt interface, open a visualization, and browse the geometry tree.

BEGINRELEASENOTES
- ddsin: Add `--runType qt` support for enabling the graphical Qt interface for geant4
- ddsim: change shell type to tcsh: enabling tab completion when Geant4 offers it

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -149,7 +149,8 @@ class DD4hepSimulation(object):
                         default=ConfigHelper.makeList(self.compactFile), type=str,
                         help="The compact XML file, or multiple compact files, if the last one is the closer.")
 
-    parser.add_argument("--runType", action="store", choices=("batch", "vis", "run", "shell", "qt"), default=self.runType,
+    parser.add_argument("--runType", action="store", choices=("batch", "vis", "run", "shell", "qt"),
+                        default=self.runType,
                         help="The type of action to do in this invocation"  # Note: implicit string concatenation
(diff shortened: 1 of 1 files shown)


wdconinc:
@dd4hepbot fix style ;-)

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -108,9 +108,6 @@ def __init__(self):
 
     self._argv = None
 
-    # use TCSH geant UI instead of QT
-    os.environ['G4UI_USE_TCSH'] = "1"
```
andresailer, line 112: Is this no longer used by Geant4?
If it is still used, please keep it if `runType != 'qt'`
This prevented many a crash in the past.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -324,6 +323,8 @@ def run(self):
 
     if self.runType == "vis":
       simple.setupUI(typ="csh", vis=True, macro=self.macroFile)
```
andresailer, line 325:
```suggestion
      simple.setupUI(typ="tcsh", vis=True, macro=self.macroFile)
```

andresailer: Also in line 331, where I can't comment any more.

wdconinc: Didn't know there was a difference... Changed all `csh` into `tcsh`.

PR #972, opened 2022-09-06
thisdd4hep.sh: use local variables where possible

veprbl:
Use of nicely named variables such as $p may clash with local user environments. A similar change was implemented in ROOT a long time ago:
https://github.com/root-project/root/commit/bba80a143655a4aac63cb5adb2e9f2d73bcd9566

BEGINRELEASENOTES
- Refactor thisdd4hep.sh to use local variables where possible to prevent user environment corruption.

ENDRELEASENOTES

### DIFF
--- a/cmake/thisdd4hep.sh
+++ b/cmake/thisdd4hep.sh
@@ -10,8 +10,8 @@
 #
 #-----------------------------------------------------------------------------
 dd4hep_parse_this()   {
-    SOURCE=${1}
-    package=${2};
+    local SOURCE=${1}
+    local package=${2};
     if [ "x${SOURCE}" = "x" ]; then
         if [ -f bin/this${package}.sh ]; then
             THIS="$PWD"; export THIS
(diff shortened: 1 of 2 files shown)


andresailer:
Does this file also need adaptations? https://github.com/AIDASoft/DD4hep/blob/master/cmake/thisdd4hep_package.sh.in

veprbl:
> Does this file also need adaptations? https://github.com/AIDASoft/DD4hep/blob/master/cmake/thisdd4hep_package.sh.in

Thanks for pointing that one out. `local` is only working inside `function`s, which are not present there. So, the answer is "no".

andresailer:
Ok, thanks! if there are no variables that clash then I will merge your changes!
Many thanks!

veprbl:
I was only hit by the use of `$p`. If someone is hit by a global scope variable such as `$SOURCE` they might have to rename it to something like `$_DD4HEP_SOURCE` and that is a different kind of change.

PR #975, opened 2022-09-14
materialBudget: enable use with asymmetric detectors

wdconinc:
All EIC detectors are asymmetric. Having this as a quick tool is useful, but we need to be able to run it for both positive and negative eta. Introduces thetaMin/thetaMax/etaMin in the steering file.

BEGINRELEASENOTES
- materialBudget: enable use with asymmetric detectors: adding thetaMin/thetaMax/etaMin to the steering file

ENDRELEASENOTES

### DIFF
--- a/UtilityApps/src/materialBudget.cpp
+++ b/UtilityApps/src/materialBudget.cpp
@@ -247,7 +247,7 @@ int main_wrapper(int argc, char** argv)   {
       }
 
 
-      double binX = ( etaMax > 0. ? (etaMin + (0.5+i)*dEta) : -(thetaMin + theta)/M_PI*180. ) ;
+      double binX = ( etaMax > 0. ? (etaMin + (0.5+i)*dEta) : -theta/M_PI*180. ) ;
 
       det.hx->Fill( binX , sum_x0 ) ;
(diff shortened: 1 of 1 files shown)


UtilityApps/src/materialBudget.cpp
```diff
@@ -233,7 +247,7 @@ int main_wrapper(int argc, char** argv)   {
       }
 
 
-      double binX = ( etaMax > 0. ? (0.5+i)*dEta : -theta/M_PI*180. ) ;
+      double binX = ( etaMax > 0. ? (etaMin + (0.5+i)*dEta) : -(thetaMin + theta)/M_PI*180. ) ;
```
andresailer, line 250: The `theta` calculation from line 255 already takes `thetaMin` into account, doesn't it?
So that should be the same as before, or not?
```suggestion
      double binX = ( etaMax > 0. ? (etaMin + (0.5+i)*dEta) : -theta/M_PI*180. ) ;
```

wdconinc: Ah, yes, you're right.

andresailer:
Thanks!

PR #976, opened 2022-09-14
geoConverter plugin to pure ROOT TGeo format (not ROOT persistency)

wdconinc:
As we continue to upstream some of our local EIC tools, here is functionality we have been using to export dd4hep to a TGeo tree in a ROOT file. As far as we know, this is not otherwise possible with command-line tools. This is different from the exports to ROOT persistency provided by DD4hep_Geometry2ROOT and intended for exports to TGeo objects that only need a standard ROOT installation to visualize (as primary application).

BEGINRELEASENOTES
- `geoConverter -compact2tgeo` to write TGeo geometry to ROOT file

ENDRELEASENOTES

### DIFF
--- a/UtilityApps/src/converter.cpp
+++ b/UtilityApps/src/converter.cpp
@@ -23,9 +23,10 @@ namespace {
     std::cout <<
       "geoConverter -opt [-opt]                                                        \n"
       "        Action flags:               Usage is exclusive, 1 required!             \n"
-      "        -compact2description               Convert compact xml geometry to description.     \n"
+      "        -compact2description        Convert compact xml geometry to description.\n"
       "        -compact2gdml               Convert compact xml geometry to gdml.       \n"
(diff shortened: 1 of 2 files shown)


UtilityApps/src/converter.cpp
```diff
@@ -98,6 +102,8 @@ namespace {
       run_plugin(description,"DD4hepGeometry2GDML",output,&argv[output]);
     else if ( compact2pand )
       run_plugin(description,"DD4hepGeometry2PANDORA",output,&argv[output]);
+    else if ( compact2tgeo )
+      run_plugin(description,"DD4hepGeometry2TGeo",argc-output,&argv[output]);
```
andresailer, line 106: I certainly don't understand the logic behind `argc-output` here, but also not really behind the `output` arguments in the other cases.

andresailer: Thank you for the explanation!
Let's leave things as they are then!

PR #979, opened 2022-09-23
Use explicit namespace in DDCond dictionary

pikacic:
Having a `using namespace` in the ROOT dictionary headers is a bad idea because those headers are used in the global translation unit used by ROOT (cling or cppyy).

In my specific case I get an error because in LHCb we have a class called `Condition` (without namespace) and clang is not able to decide what to use:
```
G__DDCond dictionary payload:145:126: error: reference to 'Condition' is ambiguous
template class std::list< std::pair< std::pair< std::string, std::pair< std::pair<std::string,int>, IOV::Key> >, std::vector<Condition> > >;
                                                                                                                             ^
Forward declarations from .../LHCbDict.rootmap:8:7: note: candidate found by name lookup is 'Condition'
class Condition;
      ^
.../include/DD4hep/Conditions.h:51:9: note: candidate found by name lookup is 'dd4hep::Condition'
  class Condition: public Handle<detail::ConditionObject> {
        ^
```

This pull requests modify the declaration of the classes to add to the dictionary of DDCond  so that the namespace is explicit.

Note that other dictionaries (not addressed here) have the same problem (there's even a `using namespace std`).

BEGINRELEASENOTES

- Use explicit namespaces in DDCond dictionary

ENDRELEASENOTES

### DIFF
--- a/DDCond/src/ConditionsDictionary.h
+++ b/DDCond/src/ConditionsDictionary.h
@@ -37,39 +37,36 @@ namespace { class ConditionsDictionary {};   }
 #pragma link C++ namespace dd4hep::detail;
 #pragma link C++ namespace dd4hep::cond;
 
-using namespace dd4hep;
-using namespace dd4hep::cond;
-
 // These are necessary to support interactivity
-#pragma link C++ class UserPool-;
-#pragma link C++ class UpdatePool-;
-#pragma link C++ class ConditionsPool-;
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch (approved):
Absolutely valid request.
I was not aware that cling actually does not resolve the proper names in the generated dictionaries. 
I will also change the other dd4hep dictionaries accodingly.

Issue #879, opened 2021-11-02
DDG4 simulation with CAD geometry !!?


Shivang-UHM:
Hi,

Here are the details of the DD4hep installation

 - OS version: e.g. VBox with Ubuntu 20.04.2 LTS 
 - Compiler version: GCC 9.3
 - DD4hep version:  v01-16-01
 - Reproduced by: Geant4 10.07p01, ROOT version: 6.22, cmake: 3.16.3
 - Input: HMBv3.xml

 - Goal: Geant4 (DDG4) simulation of detector geometry having both CAD structures and DDG4 Box segment
  
So, we have a simple geometry of our Muon Beamline (detector). It consists of 2 tracker planes and a set of 6 calorimeters stacked in the matrix of 2:3. And these tracker planes and calorimeters are exported in STL format from the CAD software Solidworks. We have taken the help from the DDCAD examples and imported the CAD geometries using XML description. 
First, we wanted to record some hits in the calorimeter by showering particles. So, we declared Calorimeter Geometry as sensitive. The compact XML description of geometry is attached (HMBv3.xml). We haven't written our own constructor and hoping existing detector constructors should be good for starting.
After initiating the ddsim simulation or python-based script it shows some initial Geant4 things and then the system kind of freezes. It counts some G4TriangularFacets and sometimes ends with the message that one of the logical volumes is outside the world volume, sometimes segmentation fault and sometimes I need to restart my VBox. 
I have tried to capture some logs please have a look and suggest where am I going wrong.
Could be some fundamental blunder as I'm just a beginner in DD4hep and still trying to understand your wonderful effort (DD4hep).
Any example where you have imported a CAD geometry and ran a Geant4 (DDG4) simulation? 
Attachment: Xml description along with cad files, python script and output log.
[DD4hep_HMB.zip](https://github.com/AIDASoft/DD4hep/files/7464411/DD4hep_HMB.zip)

Looking forward to your kind response.

Thank you,
Shivang Tripathi,
IDLab, UH Manoa

MarkusFrankATcernch:
I tried out your example.
1) The SciFi STL files do not properly translate to Geant4. There must be some errors in them.
2) The calorimeter (6CalAssembly_aluminumMaterial.STL) can be translated to Geant4. 

However, you need some sort of segmentation:
``` 
 <readouts>
    <readout name="CaloHits">
[...]

PR #983, opened 2022-10-09
fix: set tessellated solid closure flag in geant4

wdconinc:
The `G4TessellatedSolid` created in the `Geant4ShapeConverter` is by default [not closed](https://github.com/Geant4/geant4/blob/v11.0.3/source/geometry/solids/specific/src/G4TessellatedSolid.cc#L165) (v11.0.3). This sets the closure of the `G4TessellatedSolid` based on the closure of `TGeoTessellated` (with [`IsBodyClosed`](https://github.com/root-project/root/blob/v6-26-06/geom/geom/inc/TGeoTessellated.h#L141)), to ensure that closed solids are converted into closed solids.

BEGINRELEASENOTES
- Ensure that closed tessellated solids are converted to closed solids in geant4 

ENDRELEASENOTES

### DIFF
--- a/DDG4/src/Geant4ShapeConverter.cpp
+++ b/DDG4/src/Geant4ShapeConverter.cpp
@@ -274,6 +274,7 @@ namespace dd4hep {
         }
         g4->AddFacet(g4f);
       }
+      g4->SetSolidClosed(sh->IsClosedBody());
       return g4;
     }
     


DDG4/src/Geant4ShapeConverter.cpp
```diff
@@ -274,6 +274,7 @@ namespace dd4hep {
         }
         g4->AddFacet(g4f);
       }
+      g4->SetSolidClosed(sh->IsClosedBody());
```
MarkusFrankATcernch, line 277: I think I made a bug when implementing this and forgot the closure. Thanks a lot.
I have some hope that this fix will also resolve the issue:
https://github.com/AIDASoft/DD4hep/issues/879#issuecomment-1074165657

Guts feeling says this issue is connected, since Geant4 does not model assemblies as shapeless volumes, but places the daughters directly into the assembly mother...

Thanks for spotting this!

Issue #850, opened 2021-07-23
Incorrect Trap shape 


DraTeots:
- OS version: linux mint 20.02 (Ubuntu 20.04 based)
 - Compiler version: gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
 - DD4hep version: tag v01-17
 - Reproduced by: ROOT 6-24-02, 6-20-04. Geant v10.7.1
 - When I try to create Trap prizm through `Trap(double pz, double py, double px, double pLTX)` constructor, it creates the wrong shape both for ROOT and Geant4 part. 
 
 Trap I'm trying to create: 

```c++
// Look full code below
Trap prizm_shape(360 * mm, 300 * mm, 237.4608 * mm, 50 * mm);
```

What I see in TGeo viewer: 

![image](https://user-images.githubusercontent.com/4732545/126736183-55ac970d-38f8-42f9-9185-74d558061e8f.png)

What I see in Geant4 using DDSIM 

![image](https://user-images.githubusercontent.com/4732545/126736219-28f3c532-4f06-4a2f-b184-63e0268589ff.png)

So both ROOT and Geant4 conversion are affected. The picture is not correct as there should be right angles.

Just in case the same prism created in **pure** Geant4 (or how it should look) 

![image](https://user-images.githubusercontent.com/4732545/126736308-21d79d71-e090-4cf8-b643-7fab2a7e0ccc.png)

The code to reproduce is really slim and straight forward: 

```c++
#include "DD4hep/DetFactoryHelper.h"
#include "DD4hep/OpticalSurfaces.h"
#include "DD4hep/Printout.h"
#include "DDRec/Surface.h"
#include <XML/Helper.h>

using namespace std;
using namespace dd4hep;

static Ref_t createDetector(Detector& desc, xml_h e, SensitiveDetector sens)
{
    xml_det_t xml_det   = e;
    std::string det_name = xml_det.nameStr();
    int det_id   = xml_det.id();

    DetElement   det(det_name, det_id);
    Volume       mother_vol = desc.pickMotherVolume(det);
    Material quartz = desc.material("Quartz");

    Trap prizm_shape(360 * mm, 300 * mm, 237.4608 * mm, 50 * mm);
    Volume prizm("prizm", prizm_shape, quartz);

    PlacedVolume det_plvol = mother_vol.placeVolume(prizm, Position(50, 0, 0));
    det.setPlacement(det_plvol);
}
[...]

PR #986, opened 2022-10-12
fix: 4-argument TGeoTrap needs an atan

wdconinc:
The [TGeoTrap class](https://root.cern.ch/doc/master/classTGeoTrap.html#a0e27ff95f2ffa2eb8c320a7778e1f403) (for which an object is created here) has a slightly different calling syntax than what is done internally in [G4Trap](https://apc.u-paris.fr/~franco/g4doxy/html/classG4Trap.html#074ca31710e0a11f9c7b6b13ddd1eabf). We need to take the atan here to get to the angle that TGeoTrap wants.

This is #850 again. Likely not caught in testing then due to atan(x) =~ x for small x in radians. But the previous expression also didn't agree dimensionally, in particular `double fTalpha1 = 0.5*(pLTX - pX)/pY;` but then `fTalpha1/units::deg`.

BEGINRELEASENOTES
- Shapes::Trap: Fix the 4-argument trapezoidal prism constructor (again)

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/Shapes.cpp
+++ b/DDCore/src/Shapes.cpp
@@ -611,21 +611,21 @@ Trap::Trap(const string& nam,
 /// Constructor to be used when creating a new anonymous object with attribute initialization
 void Trap::make(const string& nam, double pZ, double pY, double pX, double pLTX) {
   double fDz  = 0.5*pZ;
-  double fTthetaCphi = 0;
-  double fTthetaSphi = 0;
+  double fTheta = 0;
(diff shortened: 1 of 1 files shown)


andresailer:
@MarkusFrankATcernch Re the discussion from the DD4hep meeting: The issue is that Root expects the argument to be an angle in degrees and applies DegToRad and tan on top of the argument, so it shouldn't be done before

PR #987, opened 2022-10-17
DDEve: root header was renamed

andresailer:
Change in https://github.com/root-project/root/pull/11554

BEGINRELEASENOTES
- DDEve: Adapt to renamed header in upcoming ROOT 6.28

ENDRELEASENOTES


andresailer:
Dev3 fails to build, because the new view wasn't created because DD4hep fails. Local test with this fix can be successfully build against HEAD of ROOT

PR #989, opened 2022-10-17
feat: automatically interpret hepmc3 input files

wdconinc:
When passing a HepMC3 input file with extension `.hepmc3`, `ddsim` does not know what to do with it. This allows it to interpret this extension correctly.

BEGINRELEASENOTES
- ddsim: Automatically interpret '.hepmc3' input files

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -403,7 +403,7 @@ class DD4hepSimulation(object):
       elif inputFile.endswith(".hepevt"):
         gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepevt%d" % index)
         gen.Input = "Geant4EventReaderHepEvtLong|" + inputFile
-      elif inputFile.endswith(".hepmc"):
+      elif inputFile.endswith((".hepmc", ".hepmc3")):
         if self.hepmc3.useHepMC3:
           gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index)
           gen.Parameters = self.hepmc3.getParameters()
@@ -411,13 +411,6 @@ class DD4hepSimulation(object):
         else:
           gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index)
           gen.Input = "Geant4EventReaderHepMC|" + inputFile
-      elif inputFile.endswith(".hepmc3"):
-        if self.hepmc3.useHepMC3:
-          gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index)
-          gen.Parameters = self.hepmc3.getParameters()
-          gen.Input = "HEPMC3FileReader|" + inputFile
-        else:
-          raise RuntimeError("HepMC3 input file support not enabled")
       elif inputFile.endswith(".pairs"):
         gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/GuineaPig%d" % index)
         gen.Input = "Geant4EventReaderGuineaPig|" + inputFile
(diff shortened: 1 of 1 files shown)


DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -428,6 +428,13 @@ def run(self):
         else:
           gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index)
           gen.Input = "Geant4EventReaderHepMC|" + inputFile
+      elif inputFile.endswith(".hepmc3"):
+        if self.hepmc3.useHepMC3:
+          gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index)
+          gen.Parameters = self.hepmc3.getParameters()
+          gen.Input = "HEPMC3FileReader|" + inputFile
+        else:
+          raise RuntimeError("HepMC3 input file support not enabled")
```
andresailer, line 437: I would propose to implement the hepmc3 ending check to
https://github.com/AIDASoft/DD4hep/blob/0515488c0f98f3c1a25f5d36a8b9547ad35b6a77/DDG4/python/DDSim/DD4hepSimulation.py#L546-L554
This would give an error earlier and is less likely to result in a long confusing traceback from undoing some python/c++ interplay that might already have happened here?

And then one can just change line 423 to
```python
if inputFile.endswith((".hepmc", ".hepmc3")):
```

wdconinc: I don't think combining in L423 is the right place since for hepmc3 files we cannot fall through to Geant4EventReaderHepMC which only reads hepmc version 2.

andresailer: The `.hepmc3` check for `self.hepmc3.useHepMC3 == True` is already done in `__checkFileFormat`, there is no fall through.

andresailer: How is that possible? `self.hepmc3` is set up in the `__init__` ?!?
Or do you mean if useHepMC3 is changed in the steering file or elsewhere? (Which I don't know why I allowed that in the first place in this case...)

wdconinc: > How is that possible?

`self.hepmc3` is set up in `__init__`, but is indeed changed in the steering file (potentially). One reason: hepmc(2) files don't need the hepmc3 library and can use their own reader, so in that case you can set `hepmc3.useHepMC3 = false`.

andresailer: But that wouldn't be a problem, would it? If the hepmc2 file is ending with hepmc then the file format check doesn't care if useHepMC3 is on or off, and then in the selection of the input plugin it would matter if useHepMC3 was turned off in the steering file.

wdconinc: Maybe I've just completely confused myself. In any case, I committed what I think you mean. This would bail out earlier for hepmc3 files if there is no HepMC3 support. However, if a HepMC3 file is passed with hepmc extension, it passes the early check and will only get caught later. The hepmc3 extension is more the exception than the rule, I find.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -544,6 +544,8 @@ def __checkFileFormat(self, fileNames, extensions):
       fileNames = [fileNames]
     if not all(fileName.endswith(extensions) for fileName in fileNames):
       self._errorMessages.append("ERROR: Unknown fileformat for file: %s" % fileNames)
+    if not self.hepmc3.useHepMC3 and fileName.endswith(".hepmc3"):
+      self._errorMessages.append("ERROR: HepMC3 files require the use of HepMC3 library")
```
wdconinc, line 548: This probably needs to be
```
    if not self.hepmc3.useHepMC3 and any(fileName.endswith(".hepmc3") for fileName in fileNames):
```

andresailer: yes, correct.

DDG4/python/DDSim/DD4hepSimulation.py
wdconinc: As far as I can tell, this thread is now resolved and I will mark as such.

andresailer (approved):
Thanks!

PR #1012, opened 2022-11-10
feat: allow asymmetric tracking region, default to symmetric

wdconinc:
At the EIC (and other asymmetric colliders) the detector layout is asymmetric. Thus, we wish to define asymmetric tracking regions. This PR introduces `tracker_region_zmin` which is the (signed) negative z position of the tracking region. If not specified, it is set to the negative of zmax and an info message is printed (ensuring backwards compatibility).

BEGINRELEASENOTES
- DDG4: Geant4TCUserParticleHandler: Allow for asymmetric tracking region with `tracker_region_zmin`

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDSim/Helper/ParticleHandler.py
+++ b/DDG4/python/DDSim/Helper/ParticleHandler.py
@@ -141,8 +141,8 @@ class ParticleHandler(ConfigHelper):
       try:
         user.TrackingVolume_Zmin = DDG4.tracker_region_zmin
       except AttributeError as e:
-        logger.info("Attribute tracker_region_zmin for asymmetric tracker region missing %s", e)
-        logger.info("  will use symmetric region defined by tracker_region_zmax")
+        logger.debug("Attribute tracker_region_zmin for asymmetric tracker region missing %s", e)
+        logger.debug("  will use symmetric region defined by tracker_region_zmax")
         user.TrackingVolume_Zmin = str(-float(user.TrackingVolume_Zmax))
 
       logger.info(" *** definition of tracker region *** ")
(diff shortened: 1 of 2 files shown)


DDG4/python/DDSim/Helper/ParticleHandler.py
```diff
@@ -131,16 +131,26 @@ def setupUserParticleHandler(self, part, kernel, DDG4):
...
         logger.error("Attribute of tracker region missing in detector model %s", e)
         logger.error("   make sure to specify the global constants tracker_region_zmax and tracker_region_rmax ")
         logger.error("   this is needed for the MC-truth link of created sim-hits  !  ")
         logger.error(" Or Disable the User Particle Handler with --part.userParticleHandler=''")
         exit(1)
+
+      try:
+        user.TrackingVolume_Zmin = DDG4.tracker_region_zmin
+      except AttributeError as e:
+        logger.info("Attributes tracker_region_zmin for asymmetric tracker region missing %s", e)
```
andresailer, line 144:
```suggestion
        logger.info("Attribute tracker_region_zmin for asymmetric tracker region missing %s", e)
```

andresailer (approved):
Looks good to me. The only thing I am wondering about is the new info print out, if that should be debug. But if no one objects I'll merge next week...

wdconinc:
I'll change it to debug. I was also looking for other docs on this tracking region functionality, but couldn't find it. If there are any other locations it can be included under `docs/`?

andresailer:
There is some description in https://dd4hep.web.cern.ch/dd4hep/usermanuals/DDG4Manual/DDG4Manual.pdf section 5.3 Monte-Carlo Truth Handling

wdconinc:
I finally had some time to look at the section of the docs. It is more general in scope, not specifically about `Geant4TCUserParticleHandler`, so adding this asymmetric tracking region would require describing also this particular implementation of a `Geant4UserParticleHandler`. Therefore, I think it is probably fine to leave this as is.

Issue #1010, opened 2022-11-08
Changing Geant4 physics flags 


simonge:
Hi All,

I am wanting to turn on the synchrotron physics process from Geant4EmExtraPhysics. Usually either achieved before initialization in the geant4 macro file using e.g.
`/physics_lists/em/SyncRadiation true`
or in the code with:
`synActivated(true)`

Is there a way to call this without writing a plugin? If I understand correctly the ability to call PreInit Geant4 macro commands before the initialization would solve our issue and introduce this functionality for any other such processes?

@wdconinc

wdconinc:
My thoughts on this. I know we could write a plugin for this, Geant4EmExtraPhysics or so, but this seems like adding more maintenance burden to dd4hep. I wonder if there is a better way to resolve this inside dd4hep.

As @simonge pointed out, the key issue is that this physics requires flags to be set before initialization and cannot be loaded after initialization. For example, let's say we want synchrotron radiation enabled in the FTFP_BERT list, then we'd load [FTFP_BERT.cc](https://github.com/Geant4/geant4/blob/master/source/physics_lists/lists/src/FTFP_BERT.cc), which registers `RegisterPhysics( new G4EmExtraPhysics(ver) );`, but that doesn't do anything since G4EmExtraPhysics starts with [pretty much everything disabled](https://github.com/Geant4/geant4/blob/master/source/physics_lists/constructors/gamma_lepto_nuclear/src/G4EmExtraPhysics.cc#L131).

So, between [loading the physics list](https://github.com/AIDASoft/DD4hep/blob/master/DDG4/python/DDSim/DD4hepSimulation.py#L475) and [initializing the kernel](https://github.com/AIDASoft/DD4hep/blob/master/DDG4/python/DDSim/DD4hepSimulation.py#L486) we need to inject the enabling of synchrotron radiation. That appears to be something for [`setupUserPhysics`](https://github.com/AIDASoft/DD4hep/blob/master/DDG4/python/DDSim/Helper/Physics.py#L146), so maybe we can add something like this to the steering file:
```python
def setupSynchrotonRadiation(kernel):
    kernel.getUI().ApplyCommand('/physics_lists/em/SyncRadiation true')
    return None
SIM.physics.setupUserPhysics(setupSynchrotronRadiation)
```
... but that's where I got...

MarkusFrankATcernch:
Please close if required functionality is present...

andresailer:
[...]

Issue #1004, opened 2022-11-02
DDG4: Particle charge is not properly propagated from the input stage to the output stage


MarkusFrankATcernch:
DDG4: Particle charge is not properly propagated from the input stage to the output stage
- Geant4Particle's charge is in units of 1/3 of the elementary charge
- If at all what is put in place is a value in units of the elementary charge
- needs fixing of several event readers' and the particle gun/generic particle generator
- the charge gets lost in the Geant4ParticleHandler (fix in the pipeline)

MarkusFrankATcernch:
After some tests and investigation:
This issue cannot be resolved, because MC records may also be read without Geant4 and a physics list being initialized.
Hence the event readers cannot implement such a feature if the event record does not support charge, but only the PDG code.

Conclusion is to best update the charge of particles at the output stage using the PDG code or to implement an event action, which performs this actions right before the output stage.

PR #1021, opened 2022-11-21
Multiple commits. See release notes.

MarkusFrankATcernch:
BEGINRELEASENOTES
- Implemented issue https://github.com/AIDASoft/DD4hep/issues/1010: Changing Geant4 physics flags 
  o Extended the Geant4UIManager to support command queues at various stages of the program:
    -- configure
    -- initialize
    -- pre-run
    -- post-run
    -- terminate
  o Added interface function to directly communicate with the G4UI using string commands at any time
    from python.
  o Extended Geant4Kernel functionality to allow client actions to subscribe to stages of the program
    and perform dedicated actions on:
    -- configure
    -- initialize
    -- terminate
    As an example please see DDG4/src/Geant4UIManager.cpp

- Dropped issue https://github.com/AIDASoft/DD4hep/issues/1004 after some tests and investigation:
  This issue cannot be resolved, because MC records may also be read without Geant4 and a physics list being initialized.
  Hence the event readers cannot implement such a feature if the event record does not support charge, but only the PDG code.
  Conclusion is to best update the charge of particles at the output stage using the PDG code or to implement an event action,
  which performs this actions right before the output stage.
- Simplified component properties. 
- Improvements to DDDigi (work in progress)

ENDRELEASENOTES


DDG4/python/DDG4.py
```diff
@@ -131,14 +131,29 @@ def _registerGlobalFilter(self, filter):
   self.get().registerGlobalFilter(Interface.toAction(filter))
 
 
+def _evalProperty(data):
+  """
+    Function necessary to extract real strings from the property value.
+    Strings may be emraced by quotes: '<value>'
```
andresailer, line 137:
```suggestion
    Strings may be embraced by quotes: '<value>'
```

DDG4/python/DDG4.py
```diff
@@ -131,14 +131,29 @@ def _registerGlobalFilter(self, filter):
...
+  """
+  try:
+    if isinstance(data, str):
+      return eval(data)
```
andresailer, line 141:
```suggestion
      import ast
      return ast.literal_eval(data)
   except ValueError:
```
https://docs.python.org/3.7/library/ast.html#ast.literal_eval
Safely(!) eval 
```
>>> import ast
>>> ast.literal_eval('foo')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python2.7/ast.py", line 80, in literal_eval
    return _convert(node_or_string)
  File "/usr/lib64/python2.7/ast.py", line 79, in _convert
    raise ValueError('malformed string')
ValueError: malformed string
>>> ast.literal_eval('"foo"')
'foo'
```

DDG4/src/Geant4UIManager.cpp
```diff
@@ -39,25 +40,81 @@ namespace   {
...
+  /// Get the pointer to the User Interface manager
+  G4UImanager* mgr = G4UImanager::GetUIpointer();
+  /// Start UI instance
+  if ( m_haveUI ) {
+    m_ui = startUI();
+  }
+  /// Execute the chained command statements
+  for(const auto& c : m_configureCommands)  {
+    info("++ Executing configure command:%s",c.c_str());
+    mgr->ApplyCommand(c.c_str());
```
andresailer, line 80: Should add a check for success

PR #1022, opened 2022-11-22
Take into account suggestions from PR #1021

MarkusFrankATcernch:
BEGINRELEASENOTES
- Take into account suggestions from PR #1021
ENDRELEASENOTES


andresailer:
Thanks!

PR #1025, opened 2022-11-25
Make ddsim executable

jmcarcell:
Since `ddsim` comes from copying the template and the template doesn't have execution rights at the moment `ddsim` is not an executable. Tested successfully.

BEGINRELEASENOTES

- Make ddsim executable by default

ENDRELEASENOTES


andresailer:
I am fairly sure this
https://github.com/AIDASoft/DD4hep/blob/c1060cfa9136a062fd8a263a14afc248ca8842bd/DDG4/CMakeLists.txt#L153
make ddsim be executable.
Under what circumstances do you need the `.in` file to be executable as well?

Issue #1036, opened 2023-01-05
problem with MCParticles folder produced in the root output of ddsim?


saraheno:
Hi,
Perhaps I have a misunderstanding about what is supposed to be in the MCParticles folder.  I am running single particles and I want to get the energy of the incident particle.  I thought that would be the first particle in this folder, and that seems to work when I run muons.  But when I run electrons, there is usually no particle in the list that corresponds to the gun energy.

I am running https://github.com/saraheno/SingleDualCrystal  with the command

 ddsim --compactFile=./DRSingleCrystal.xml --runType=batch -G --steeringFile SCEPCALsteering.py --outputFile=out.root --part.userParticleHandler='' -G --gun.position="0.,30.,0." --gun.direction "0 -1 0" --gun.energy "10*GeV" --gun.particle="e-" -N 5

and am looking at MCParticles in out.root

(the actually code I use to look at out.root is https://github.com/saraheno/SingleDualCrystal/blob/main/compact/CrystalAna.C

Should I not expect to find the initial gun particle in this list?  If not, is there any thing I can add to get this information into the output root file or do I need to code this myself?

thanks
Sarah Eno

andresailer:
Which version of dd4hep are you using? To make sure you have this fix #949

saraheno:
I pulled it from the hub just a few days ago, so I should have this fix.

I think my problem is that the electron is producing optical photons in the air before it reaches my detector, and so the first particle doesn't get "marked" to be saved in the MCparticles folder?  Maybe I need some way to ensure it is marked right at the start?  I think I know that this line https://github.com/saraheno/SingleDualCrystal/blob/565d4e7bc096620106ceaabd20e49af5b6758c55/src/DualCrystalCalorimeterSDAction.cpp#L256  is what puts particles into the MCparticles folder once it reaches my detector?

andresailer:
This is somehow related to the Cerenkov and Scintillation physics treatment.
When I change to
```
     scint.TrackSecondariesFirst = False
     cerenkov.TrackSecondariesFirst = False
```

I get the primary energy of the electrons in the root output files. Somehow treating the optical photons before the electron is finished causes its values to be changed.

saraheno:
Is there someway/places I can access the particle just as first read from the gun, and force it to be stored in MCParticles?

andresailer:
I hope we can find a general solution to this problem.

PR #1046, opened 2023-01-18
ParticleHandler: account for modified stepping actions, where tracking…

andresailer:
… of a particle could be paused and later restarted.

Pick up the particle information that we have stored and go on from there

BEGINRELEASENOTES
- ParticleHandler: account for modified stepping actions, where tracking of a particle could be paused and later restarted.

ENDRELEASENOTES

For #1036 I get this for 2 events now

```
root [2] EVENT->Scan("sqrt(MCParticles.psx^2 + MCParticles.psy^2 + MCParticles.psz^2 + MCParticles.mass^2):MCParticles.reason:MCParticles.pdgID:MCParticles.mask", "MCParticles.mask==1")
***********************************************************************
*    Row   * Instance * sqrt(MCPa * MCParticl * MCParticl * MCParticl *
***********************************************************************
*        0 *        0 *     10000 *       158 *        11 *         1 *
*        1 *        0 *     10000 *       158 *        11 *         1 *
***********************************************************************
==> 2 selected entries
(long long) 2
```
So the energy is the same as the initial one (10000) as opposed to some smaller number.

### DIFF
--- a/DDG4/include/DDG4/Geant4ParticleHandler.h
+++ b/DDG4/include/DDG4/Geant4ParticleHandler.h
@@ -119,6 +119,9 @@ namespace dd4hep {
       Particle          m_currTrack;
       /// Map with stored MC Particles
       ParticleMap       m_particleMap;
+      /// Map with stored MC Particles that were suspended by the stepping action
+      ParticleMap       m_suspendedPM;
+      bool              m_haveSuspended = false;
       /// Map associating the G4Track identifiers with identifiers of existing MCParticles
       TrackEquivalents  m_equivalentTracks;
 
--- a/DDG4/src/Geant4ParticleHandler.cpp
+++ b/DDG4/src/Geant4ParticleHandler.cpp
@@ -316,16 +316,6 @@ void Geant4ParticleHandler::end(const G4Track* track)   {
   Geant4ParticleHandle ph(&m_currTrack);
   const int g4_id = h.id();
 
-  if(track->GetTrackStatus() == fSuspend) {
-    m_haveSuspended = true;
-    //track is already in particle map, we pick it up from there in begin again
-    if(m_particleMap.find(g4_id) != m_particleMap.end()) return;
-    //track is not already stored, keep it in special map
-    auto iPart = m_suspendedPM.emplace(g4_id, new Particle());
-    (iPart.first->second)->get_data(m_currTrack);
-    return; // we trust that we eventually return to this function with another status and go on then
-  }
-
   int track_reason = m_currTrack.reason;
   PropertyMask mask(m_currTrack.reason);
   // Update vertex end point and final momentum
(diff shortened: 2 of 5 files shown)


andresailer:
@MarkusFrankATcernch Does this look sensible to you?

MarkusFrankATcernch:
If the tracking action is the first call to the particle handler after the restacking, this is should be correct.
However, I have the suspicion that all properties of the existing "currentTrack" may all somehow lost, because this
track and its properties is not yet saved away to the collection possibly be kept in the output record.

Hence the question is: Whan does the restacking happen? before or after a couple of steps were already processed?

MarkusFrankATcernch:
if before the restacking the member
```
void Geant4ParticleHandler::end(const G4Track* track)
```
is called this approach should be correct. Be aware that this track may be in both collections:
- Geant4ParticleHandler::m_equivalentTracks and
- Geant4ParticleHandler::m_particleMap
in `Geant4ParticleHandler::endEvent(const G4Event* event)` preference must be given to `Geant4ParticleHandler::m_particleMap` when rebasing the tracks.

andresailer:
I implemented the treatment also for particles that might not be in the particleMap before they are suspended. It looks fine to me, but that code is never actually run so far.

andresailer:
I am adding a test

andresailer:
Added a test that exercises the suspendedPM related code

DDG4/include/DDG4/Geant4ParticleHandler.h
```diff
@@ -119,6 +119,9 @@ namespace dd4hep {
       Particle          m_currTrack;
       /// Map with stored MC Particles
       ParticleMap       m_particleMap;
+      /// Map with stored MC Particles that were suspended by the stacking action
```
andresailer, line 122:
```suggestion
      /// Map with stored MC Particles that were suspended by the stepping action
```

DDG4/src/Geant4ParticleHandler.cpp
```diff
@@ -115,6 +115,8 @@ bool Geant4ParticleHandler::adopt(Geant4Action* action)    {
 void Geant4ParticleHandler::clear()  {
   detail::releaseObjects(m_particleMap);
   m_particleMap.clear();
+  // m_suspendedPM should already be empty and cleared...
+  assert(m_suspendedPM.empty());
```
andresailer, line 119:
```suggestion
  assert(m_suspendedPM.empty() && "There was something wrong with the particle record treatment, please open a bug report!");
```

DDG4/src/Geant4ParticleHandler.cpp
```diff
@@ -293,7 +314,10 @@ void Geant4ParticleHandler::begin(const G4Track* track)   {
 void Geant4ParticleHandler::end(const G4Track* track)   {
   Geant4TrackHandler h(track);
   Geant4ParticleHandle ph(&m_currTrack);
-  int g4_id = h.id();
+  const int g4_id = h.id();
+
+  std::cout << "ending " << h.statusName() << " " << g4_id  << std::endl;
```
andresailer, line 319:
```suggestion
```

PR #1044, opened 2023-01-18
Add track momentum to the MonteCarloContribution of Geant4Hit structure

MarkusFrankATcernch:
BEGINRELEASENOTES
- This PR addresses issue https://github.com/AIDASoft/DD4hep/issues/1037
  Rather than adding the beta as a double it was chosen to add at each step the track momentum
  in floats which for 4 more bytes offers more information.
ENDRELEASENOTES

PR #1047, opened 2023-01-19
Geant4Data: bring back a 4 argument version of Hit constructor used i…

andresailer:
…n lcgeo

This 4 argument constructor is used in lcgeo (k4geo) code, so lcgeo is currently not compiling.
(Update for #1044)

* Maybe also used in other places, so better to keep the interface?


MarkusFrankATcernch:
This is fine. I simply wanted to ensure that position and momentum of the initial point also propagate to the MC truth contribution.

PR #1052, opened 2023-01-29
Remove not implemented declaration of member function PlacedVolume::flags()

veprbl:
BEGINRELEASENOTES
- Removed not implemented declaration of member function `PlacedVolume::flags()`

ENDRELEASENOTES


veprbl:
Turns out, it's the `Volume` that has a `flags` field.

PR #1053, opened 2023-01-30
Require compatible podio version

paulgessinger:
I believe DDDigi implicitly requires podio>=0.16.0 if edm4hep is enabled. CMake does not check this so far, I'm adding a CMake-level check here.

Let me know what you think!

BEGINRELEASENOTES
- Check podio version for compatibility in cmake
ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,12 @@ endif()
 
 if(DD4HEP_USE_EDM4HEP)
   find_package(EDM4HEP REQUIRED)
+
+  set(podio_required 0.16.0)
+  find_package(podio REQUIRED ${podio_required})
+  if(NOT podio_VERSION VERSION_GREATER_EQUAL podio_required)
+    message(FATAL_ERROR "podio found version ${podio_VERSION} is insufficient (${podio_required})")
+  endif()
 #  DD4HEP_SETUP_EDM4HEP_TARGETS()
 endif()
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -151,6 +151,12 @@ endif()
 
 if(DD4HEP_USE_EDM4HEP)
   find_package(EDM4HEP REQUIRED)
+
+  set(podio_required 0.16.0)
+  find_package(podio REQUIRED ${podio_required})
```
andresailer, line 156:
```suggestion
  find_package(podio ${podio_required} REQUIRED)
```
And that should error by itself, so the next check isn't needed.
(https://cmake.org/cmake/help/v3.25/command/find_package.html#basic-signature [version] comes before [REQUIRED])

andresailer:
And if you could change this
https://github.com/AIDASoft/DD4hep/blob/4f6c0b50e3e1e16cb627f64a0d5f0d23ef1e007d/.github/workflows/clang-tidy.yml#L11

To
```yaml
LCG: ["dev4/x86_64-centos7-clang12-opt"] 
```

That should fix the error there.

Thanks!

CMakeLists.txt
paulgessinger: You're right. I thought it was because the versions were all the same major version (0) but that seems not to have been the case.

andresailer (approved):
Thanks!

PR #1055, opened 2023-02-02
DD4hep.cmake: respect DYLD_LIBRARY_PATH when running listcomponents

veprbl:
Dropping DYLD_LIBRARY_PATH produces macOS-only issues when ROOT libraries are linked with -Drpath=ON. Users of a broken DYLD_LIBRARY_PATH should be able to unset it manually.

This partially reverts 65a090ff5191ad73d4a120bc4e18ab68159cbaae .

BEGINRELEASENOTES
- Pass `DYLD_LIBRARY_PATH` environment to listcomponents_dd4hep when generating rootmaps in CMake

ENDRELEASENOTES

### DIFF
--- a/.github/utils/ctest2junit.xsl
+++ b/.github/utils/ctest2junit.xsl
@@ -1,0 +1,116 @@
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
+  <xsl:output method="xml" indent="yes"/>
+  <xsl:template match="/Site">
+    <testsuite>
+      <xsl:variable name="BuildName"><xsl:value-of select="@BuildName"/></xsl:variable>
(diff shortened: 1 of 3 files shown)


andresailer:
What kind of issues does this cause when rpath is ON?

veprbl:
> What kind of issues does this cause when rpath is ON?

ROOT reports its CMake `ROOT::*` targets in `$ROOTSYS/cmake/ROOTConfig-targets-release.cmake` with `@rpath` instead of the absolute path. So if DD4hep is built without `$ROOTSYS/lib` in rpath, there is no way to fall back on DYLD_LIBRARY_PATH from thisroot.sh.

andresailer (approved):
Our tests on mac11 still work well enough with this change.
I don't remember why the difference was introduced...

PR #1060, opened 2023-02-16
Prevent a compact file from being loaded twice

bencouturier:
Draft attempt at preventing compact files from being loaded twice.
The issues are that:
- the set containing already the list of files already loaded is in an anonymous namespace, but its access is  not multithread safe.
- the path of the XML is just normalized, as the dd4hep::Path::full_path does not seem to return the full path, so we are sensitive to the way the path is passed 

BEGINRELEASENOTES

- Prevent a compact file from being loaded twice if the debug option "incguard" has been set.

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/Parsers/detail/Conversions.h
+++ b/DDCore/include/Parsers/detail/Conversions.h
@@ -18,7 +18,7 @@
  *  Note: Do NEVER include this file directly!
  *
  *  Use the specific include files in the XML or JSON directory!
- *  Also NO header guards!
+ *
  */
 
 
--- a/DDCore/include/XML/UnicodeValues.h
+++ b/DDCore/include/XML/UnicodeValues.h
@@ -225,6 +225,7 @@ UNICODE (idspecref);
 UNICODE (ignore);
 UNICODE (include);
 UNICODE (includes);
+UNICODE (incguard);
 UNICODE (incoming_r);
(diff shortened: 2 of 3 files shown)


MarkusFrankATcernch:
As the tests show, this implementation shoots a bit too low.
The list of parsed xml files must be placed into some context related to the main parsing instance.

bencouturier:
At first I was not sure how to do that as new instances of the Converter templated class are created when needed.
We could add the set of parsed files as an optional argument to the constructor or the operator() and pass it from the first Converter<Compact> call. I can implement something like that tomorrow if that is fine for you.

bcouturi:
Adding the list of processed files as an argument to the "Converter" does not work as we load plugins.
One way that would work is attaching that list to the detector Description, which is done in the current commit.
The current solution is not ideal either as I ad to do a Compact loading specific method to the Detector interface, I am trying to use the extension mechanism instead.

MarkusFrankATcernch:
@bencouturier If you think this is good like this, please remove the "Draft", then I shall merge.

bcouturi:
The MR works fine for me, I'd say that the like issue is how to get an absolute path of the file, but that does not seem blocking.

MarkusFrankATcernch:
@ben You have to rebase before I can merge.....

PR #1065, opened 2023-03-01
feat: add energy deposition test for OpticalTracker example

c-dilks:
BEGINRELEASENOTES
- Add test for average energy deposition in `OpticalTracker` example, for `Geant4SensitiveAction<Geant4OpticalTracker>`

ENDRELEASENOTES

### DIFF
--- a/examples/OpticalTracker/scripts/test_energy_deposition.C
+++ b/examples/OpticalTracker/scripts/test_energy_deposition.C
@@ -1,22 +1,20 @@
-#include "CLHEP/Units/SystemOfUnits.h"
-
 void test_energy_deposition(TString sim_file_name="sim.root") {
 
   // test requirements
-  const Double_t min_edep = 2*CLHEP::eV;
-  const Double_t max_edep = 5*CLHEP::eV;
+  const Double_t min_edep = 2 /*eV*/;
+  const Double_t max_edep = 5 /*eV*/;
 
   // get average energy deposition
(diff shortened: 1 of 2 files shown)


c-dilks:
This test should pass now, since #1067 is resolved.

c-dilks:
Hmm... I'm not sure how to correctly handle units in the test. The `key4hep` job does not have access to `CLHEP` units, and at least in my local build, `dd4hep::eV != CLHEP::eV`.

examples/OpticalTracker/scripts/test_energy_deposition.C
```diff
@@ -0,0 +1,22 @@
+#include "CLHEP/Units/SystemOfUnits.h"
```
andresailer, line 1:
```suggestion
#include "DD4hep/DD4hepUnits.h"
```

andresailer: And dd4hep::ev etc

andresailer: Depending on if gean4 units are used or not one has to convert the energy in the file. Or convert the energy to the expected unit before it is written to file.

andresailer: Actually, here one should just use geant4 units in all cases.

c-dilks:
Oh, thanks @andresailer! Sorry I got distracted working on other things...

PR #1068, opened 2023-03-01
fix: fill energy deposition for `Geant4OpticalTracker`

### LINKED ISSUE #1067: Energy deposition from `Geant4SensitiveAction<Geant4OpticalTracker>` is not filled
The energy deposition for sensitive action `Geant4OpticalTracker` is no longer filled; instead, energy deposition is always zero. I added a new test in #1065 in the `OpticalTracker` example, which uses this action; this test passes in an older release, but is currently failing on `master`.

According to the git blame of `DDG4/plugins/Geant4SDActions.cpp`, the issue may have been introduced by 11be96f.

 - Package version: `master`
 - Reproduced by: new (failing) test https://github.com/AIDASoft/DD4hep/pull/1065

c-dilks:
Fix #1067

I'm not sure if this is the correct way to fix this, but it does restore the energy deposition distribution for the `OpticalTracker` example. In particular:
- does `hit->truth` need to be filled too?
- `step->GetTotalEnergyDeposit()` in the `print` statement is still always zero (but `hit->energyDeposit` looks correct)

BEGINRELEASENOTES
- fix: fill energy deposition for `Geant4SensitiveAction<Geant4OpticalTracker>`

ENDRELEASENOTES

### DIFF
--- a/DDG4/plugins/Geant4SDActions.cpp
+++ b/DDG4/plugins/Geant4SDActions.cpp
@@ -171,7 +171,9 @@ namespace dd4hep {
       double    hit_len   = direction.R();
       
       Hit* hit = new Hit(h.trkID(), h.trkPdgID(), h.deposit(), tim, hit_len, pos, mom);
-      hit->cellID = cellID(step);
+      hit->truth         = Hit::extractContribution(step);
+      hit->energyDeposit = hit->truth.deposit;
+      hit->cellID        = cellID(step);
       if (track->GetDefinition() != G4OpticalPhoton::OpticalPhotonDefinition()) {
         track->SetTrackStatus(fStopAndKill);
       }


MarkusFrankATcernch:
Hi,
Thanks for spotting this!

If I trace the changes down, then the difference in the energy deposit for optical photons in the OpticalTracker SD is:
-- BAD: `h.deposit()` aka `G4Step::GetTotalEnergyDeposit()`
-- GOOD: deposit from MC contribution: `deposit =
    (h.trackDef() == G4OpticalPhoton::OpticalPhotonDefinition()) ? h.trkEnergy() : h.totalEnergy(); `
where `totalEnergy = GetTotalEnergyDeposit` and `trkEnergy = track->GetTotalEnergy()`
which of course makes a difference.

Is this correct ?
If YES, then one could do quite some optimization here, because the SD callback `Geant4SensitiveAction<Geant4OpticalTracker>::process()`  does the same calculations
as  `Geant4HitData::extractContribution(...)`, which is called in addition.

c-dilks:
> Is this correct ?

I'm not sure. @wdconinc spent some time working on this SD action, so perhaps he could comment.

Issue #1072, opened 2023-03-04
Allow etaMin, etaMax bounds in Geant4IsotropeGenerator and DDSim/Helper/Gun.py


wdconinc:
Just ran a lot of simulations with `sim.gun.etaMin = -1` and  `sim.gun.etaMax = +1`... which turned out to not be a setting that exists in DD4hep, but DDSim didn't complain about it.

This issue reports two aspects:
- [x] using a steering file with non-existent settings did not lead to complaints (not sure how dd4hep would find out),
- [x] I'm going to be contributing etaMin and etaMax as a PR (which is why I picked "new feature" as category).

To reproduce:
```
cat << EOF > gamma_10GeV_eta.steer
from DDSim.DD4hepSimulation import DD4hepSimulation
from g4units import mm, GeV, MeV, degree
SIM = DD4hepSimulation()
SIM.gun.energy = 10*GeV
SIM.gun.particle = "gamma"
SIM.gun.position = (0.0, 0.0, 0.0)
SIM.gun.direction = (0.0, 0.0, 1.0)
SIM.gun.distribution = "uniform"
SIM.gun.etaMin = -1.0
SIM.gun.etaMax = +1.0
EOF
ddsim --compactFile $DD4hepINSTALL/DDDetectors/compact/SiD.xml -G -N 10 --steeringFile gamma_10GeV_eta.steer 
```

andresailer:
Dynamically adding properties to instances is of course a core python feature. We can probably prevent this with "__slots__"?

andresailer:
Still work to do.

andresailer:
I implemented a check for unknown properties.

PR #1075, opened 2023-03-13
feat: add etaMin and etaMax in ddsim (Helper/Gun.py)

wdconinc:
This adds the arguments `--gun.etaMin` and `--gun.etaMax`. If both `thetaMin` and `etaMax` are specified, `etaMax` has precedence. Addresses part of #1072.

BEGINRELEASENOTES
- Support for etaMin and etaMax in ddsim.

ENDRELEASENOTES


wdconinc:
Not sure why the key4hep test fails...

PR #1088, opened 2023-04-06
fix units in Geant4FastSimShowerModel::constructSensitives

gaede:
BEGINRELEASENOTES
- fix units in `Geant4FastSimShowerModel::constructSensitives`
     - used for energy check in `Geant4FastSimShowerModel::check_trigger`  for fast simulation

ENDRELEASENOTES

Logically this should be correct:  `dd4hep::_toDouble(prop.second)/dd4hep::GeV*CLHEP::GeV` - we have a dd4hep property for an energy and need it in Geant4/CLHEP units for GeV. However this is a factor 1000. too large (old code was 1000. too small).  @MarkusFrankATcernch Am I missing sth. here ? What is the correct way to convert dd4hep unit to g4 ?


gaede:
If I understand this correctly, we use G4units in ddsim exclusively. Then there should be no conversion applied at all in `Geant4FastSimShowerModel::constructSensitives`...

PR #1095, opened 2023-04-21
CI: adding test result upload

andresailer:
This adds upload and parsing of the test results and a display of the result in a comment to the PR, which is updated when there are results from updated commits. Should make it slightly easier to see error logs for failed tests.

The "Upload Event Payload" is needed for PRs from outside forks to know which artifacts to download.

Thanks to @olemorud for the developments as part of root-project CI

Looks like this: https://github.com/andresailer/DD4hep/pull/3#issuecomment-1519076356

some cmake versions are too old and not supporting the junit output.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -45,7 +45,7 @@ jobs:
           ninja install
           . ../bin/thisdd4hep.sh
           echo "::group::Test"
-          ctest --output-on-failure -j4
+          ctest --output-on-failure -j4 --output-junit TestResults_1.xml
           echo "::group::CMakeExamples"
           cd ../examples/
           mkdir build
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -57,7 +63,13 @@ jobs:
...
           ninja install
           echo "::group::TestExamples"
-          ctest --output-on-failure -j2
+          ctest --output-on-failure -j2 --output-junit TestResults_2.xml
+
+    - name: Upload test results
+      uses: actions/upload-artifact@v3
+      with:
+        name: Test Results ${{ env.ARTIFACT_NAME }}
+        path: /home/runner/work/DD4hep/DD4hep/build/TestResults_*.xml
```
andresailer, line 72:
```suggestion
        path: |
           /home/runner/work/DD4hep/DD4hep/build/TestResults_1.xml
           /home/runner/work/DD4hep/DD4hep/examples/build/TestResults_2.xml
```

Issue #1091, opened 2023-04-19
Repetition of sub-detector consisting of slices of different shapes 


Gongzi-Zhang:
Hi:

I am trying to repeat a sub-detector consisting of slices of different shapes, and was prompted that **'Wrong bounding box for volume HCAL5_envelope'** . I want to ask is there a way to do the job without typing each sub-detector separately? 

Many thanks.

Here is my code and sorry for the mess, it looks good in the editor panel, I don't know why the terrible display. 

(edit: you have to use triple tick marks \`\`\`, not just one  around code blocks... (and optionally add the language for highlighting) look at your post after my edits)

```xml
 <detector id="5" name="HCAL5" type="ZDC_Sampling" readout="HCALHits" vis="GreenVis">
     <position x="2.37*cm" y="2.37*cm" z="511.28*cm"/>                         
      <rotation x="0" y="0" z="0"/>                                             
      <dimensions x="4.74*cm" y="4.74*cm" z="16.92*cm"/>                        
      <layer repeat="6">                                                        
        <box name="Absorber" material="Steel235" x="4.74*cm" y="4.74*cm" z="2*cm"  vis="AnlGray" sensitive="false"/>
        <box name="Air_gap"  material="Air"      x="4.74*cm" y="4.74*cm" z="0.52*cm" sensitive="false"/>
        <polyhedra name="Scint" material="Polystyrene" numsides="6" rmin="0*cm" rmax="2*cm" dz="0.3*cm"  vis="AnlOrange" sensitive="true"/>
      </layer>                                                                  
    </detector>    
```

MarkusFrankATcernch:
@Gongzi-Zhang Sorry for the late reply.....
If I understand correctly these are actually 2 questions:

1. Why does the envelope cause overlaps (Bounding box errors).

2. Does one have to describe every HCALX (X=1,...n) as a separate subdetector?

add 1) Can it be that the envelope has "touching surfaces"? 
           Even if these are mathematically correct, tiny numerically differences can cause such overlaps. 
           Try to make either the inside volumes some epsilon smaller (~1e-10 or so) or the envelopes bigger.
           1e-10 does not change any physics.

add 2) In principle you can describe anything as coarse as you want. Inside the `<detector/>` tag you only have to supply the
            necessary constants to parametrize what you want to describe. If these are many HCAL(1....N), 1 single one or a few is
            your decision...and depends on what the result is supposed to do. Just try to have sort of "balanced" volumes: there 
[...]

PR #1098, opened 2023-04-25
Upgrade MiniTel example to serve as example for issue

MarkusFrankATcernch:
BEGINRELEASENOTES
- MiniTel Generate example should replect the detector in issue (https://github.com/AIDASoft/DD4hep/issues/1091)

ENDRELEASENOTES


MarkusFrankATcernch:
@andresailer In test job https://github.com/AIDASoft/DD4hep/actions/runs/4797217643/jobs/8535995123
There seems to be a mismatch in the Geant4 version. Is this true?

andresailer:
@MarkusFrankATcernch I think the problem is rather missing geant4 include dirs in ROOT_INCLUDE_PATH
I wanted to debug this soon. And the DDCad problem is hopefully solved with https://github.com/AIDASoft/DD4hep/pull/1099

andresailer:
Actually the Geant4 include path given to the command in the test is wrong, there is a superfluous lib64 
`-I/cvmfs/sw.hsf.org/spackages7/geant4/11.1.1/x86_64-centos7-gcc11.2.0-opt/jlctn/lib64/include/Geant4`,

andresailer:
The location of the Geant4Config.cmake etc. changed in 11.1.1? now it is in `lib64/cmake/Geant4`, instead of `lib64/Geant4-11.1.0`, and we derive the wrong location of the include directories from that.

PR #1099, opened 2023-04-25
DDCad: test: use newer duck for newer assimp versions; geant4 11.1.1 fix

andresailer:
BEGINRELEASENOTES

- thisdd4hep.sh: make the Geant4 path discovery compatible with Geant4 11.1.1

ENDRELEASENOTES

make the colada duck example compatible with newer assimp versions


andresailer:
Also a problem with the ROOT libraries location in the key4hep stack

andresailer:
I suspect that the key4hep artifacts have the same name so those tests are only shown once :/
Anyway, will merge this and change the key4hep artifact name somewhere else.

PR #1100, opened 2023-04-26
CMake: update requirement to 3.14 minimum

andresailer:
BEGINRELEASENOTES
- CMake: increase required version of Cmake from 3.12 to 3.14

ENDRELEASENOTES


.github/workflows/linux.yml
```diff
@@ -127,7 +127,7 @@ jobs:
     - name: Prepare Artifact Name
       if: success() || failure()
       run: |
-          artifact_name="${{ matrix.LCG }}"
+          artifact_name="${{ matrix.view_path }}"
```
andresailer, line 130:
```suggestion
          artifact_name="${{ matrix.view-path }}"
```

PR #1106, opened 2023-05-04
Put error context into the exception message instead of dumping to std::cout in DetectorImp::getRefChild()

veprbl:
BEGINRELEASENOTES
- DetectorImp::getRefChild() will no longer print to error context to std::cout, it will instead appear in the message of the runtime_exception thrown.

ENDRELEASENOTES


DDCore/src/DetectorImp.cpp
```diff
@@ -642,15 +642,20 @@ Handle<NamedObject> DetectorImp::getRefChild(const HandleMap& e, const string& n
...
-      cout << "   " << cnt << "  " << (*i).first << endl;
-    throw runtime_error(nam+": Cannot find a child with the reference name:" + name);
+    std::stringstream err;
+    err << "getRefChild: Failed to find child with name: " << name
+        << " Map " << nam << " contains " << e.size() << " elements: {";
+    for (it = e.begin(); it != e.end(); ++it) {
+      if (it != e.begin()) {
+        err << ", " << endl;
+      }
+      err << it->first << endl;
```
wdconinc, line 655: Did you want an `endl` here? This results in e.g. 
```
[2023-05-06 07:41:33.824] [JEventProcessorPODIO] [error] Omitting PODIO collection 'EcalBarrelScFiRecHits' due to exception: getRefChild: Failed to find child with name: EcalBarrelScFiHits Map contains 28 elements: {B0ECalHits
, 
B0TrackerHits
, 
DIRCBarHits
, 
DRICHHits
, 
EcalBarrelSciGlassHits
, 
EcalEndcapNHits
, 
EcalEndcapPHits
, 
EcalEndcapPInsertHits
, 
ForwardOffMTrackerHits
, 
ForwardRomanPotHits
, 
HcalBarrelHits
, 
HcalEndcapNHits
, 
HcalEndcapPInsertHits
, 
LFHCALHits
,
...
```

veprbl: Good catch, fixed.

PR #1108, opened 2023-05-07
Geant4Output2EDM4hep: delay frame writing until commit

wdconinc:
This aims to allow for writing multiple collections with the same name, which is currently not possible because `m_frame` doesn't accept the next `put` if the collection name is already in use. This shifts the collecting of hits from different geant4 hit collections into the plugin, and moves the collections over to the `m_frame` at the end of the event only.

Fixes https://github.com/AIDASoft/DD4hep/issues/1111.

BEGINRELEASENOTES
- Delay writing EDM4hep collections until `commit` at end of event, allowing for multiple collections with same name

ENDRELEASENOTES


wdconinc:
Compiles now, segfaults in `commit`.

wdconinc:
Maybe extract is just not necessary (or even harmful). When extract gets a node handle, that node handle retains ownership and a move on the mapped value doesn't relinquish that ownership. When the nose handle goes out of scope it takes the value down with it, before it has been put in the frame.

Instead, maybe this is a case where we can use a map to a unique_ptr instead. Then, when we want to transfer ownership we can do that without needing extract and we can just erase the entry.

wdconinc:
Tests fail `t_DDCMS_Geant4_LONGTEST` in key4hep sw-nightlies, but that appears to be unrelated. The rest works for me.

wdconinc:
@tmadlener Maybe you are interested in taking a look here. One performance aspect (maybe) is that we create a podio collection for all geant4 hit collections, even if there were no hits in them. I suspect that collection initialization is not exactly reducing the performance of geant4 simulations, but is there any way in which we can avoid creating podio collections for empty geant4 hit collections in an event?

wdconinc:
Ah, and sorting the collections in the frame alphabetically might be nice (easier to find in ROOT). Not sure if that's the job of podio or of the dd4hep plugin.

tmadlener:
Regarding the comment with a link to the blog post about insertion performance, why not make the maps use the collection types directly instead of `unique_ptr`? Then simply using `operator[]` should do the trick on the first call.

Skipping empty collections from Geant4 might save a bit of performance, but on the other hand it might introduce some issues down the line, where if you skip empty collections you would have to potentially put in some checks to see if the collection is there, rather than just having an empty loop. In a functional framework approach, IMHO having empty collections instead of missing collections is easier to deal with, so I would keep empty collections from the beginning. Also because this makes it easier to spot typos in configuration files. In the end it's a policy decision, where for me the advantages of having empty collections outweigh potential (hopefully tiny) performance hits.

> Ah, and sorting the collections in the frame alphabetically might be nice (easier to find in ROOT). Not sure if that's the job of podio or of the dd4hep plugin.

I think that would be a podio responsibility, making sure to create the branches in the correct order. If I don't get to this the next few days, could you remind me and open a feature request?

wdconinc:
> why not make the maps use the collection types directly instead of `unique_ptr`?

Yeah, that's now possible and clearly a better solution that the `unique_ptr`. I forget what the motivation for the `unique_ptr` was again originally.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -428,9 +444,8 @@ void Geant4Output2EDM4hep::saveCollection(OutputContext<G4Event>& /*ctxt*/, G4VH
   debug("+++ Saving EDM4hep collection %s with %d entries.", colName.c_str(), int(nhits));
   //-------------------------------------------------------------------
   if( typeid( Geant4Tracker::Hit ) == coll->type().type()  ){
-    edm4hep::SimTrackerHitCollection sthc;
     for(unsigned i=0 ; i < nhits ; ++i){
-      auto sth = sthc->create();
+      auto sth = m_trackerHits[colName]->create();
```
tmadlener, line 448: Can `saveCollection` be called on multiple threads concurrently? In that case the previous version was thread safe, while this will no longer be and we would need to take care of that. The implicit map insertion might be a problem, but also `create` will race when called from multiple threads on the same collection.

MarkusFrankATcernch:
@tmadlener Thomas,
are you sure this can ever work?
Typically the handles require a "medium" where the objects live in or are the handles more or less simply shared pointers?
I always assumed the handles must reside inside the "frame"/TTree. Then  you cannot see any handle without also having
to care about the frame at the same time.

I must also say that I do not really see the use case except the academic one.
Parallelism is not really the issue when talking about I/O after Geant4. In any case any I/O operation with ROOT 
must be serialized  with a global lock to prevent crashes. ROOT is very picky on this.

tmadlener:
My main point was that if `saveCollection` can be called from multiple threads concurrently, than there are possible race conditions in the proposed version here. In the previous version each `saveCollection` had its own, independent collection(s) to operate on and `Frame::put` is thread safe. Now if more than one thread enters `saveCollection` with the same name, then the calls to `Collection::create` can race. (The implicit creation of a collection via `map::operator[]` should be fine, since it doesn't invalidate any iterators).

On the other hand if `saveCollection` is guaranteed to never be called by multiple threads, than there is no problem and this version if fine. I simply wasn't sure whether this is the case.

wdconinc:
Is there anything I can do to help test here? I'm afraid I'm not qualified to judge the intention or guarantees in regards to multithreaded running.

tmadlener:
I think we concluded in one of the last DD4hep developers meetings that my concerns are not really relevant in this case, since all of this happens on a single thread, so that this should work as it is.

wdconinc:
@MarkusFrankATcernch My apologies for having remaining issues in this PR. Of course it was not my intention to have programming mistakes, nor was it my intention to drop this on anyone else for debugging. I knew I would be mostly unavailable from mid-May to mid-June to look into this further and I tried to complete it before that time. Except for the possible multithreading concerns, this PR did run fine in my local tests (presumably because we have at least one hit all the time), and the programming mistake was introduced right before I left. I let several people here know that I would be away to explain the lack of activity and to allow someone to look into this if they considered it a priority, not to rely on them to debug this (I don't like posting my travel schedule publicly on GitHub for everyone to see, so you were not aware). So, apologies for this mistake, and I'll expand my local testing to try to catch issues in the future. I appreciate your patience and willingness to work with the broader community, even if that does introduce potential for mistakes sometimes.

PR #1113, opened 2023-05-19
Remove the Frame re-implementation and use the one from podio

tmadlener:
BEGINRELEASENOTES
- Remove the copy&past `DigiFrame.h` and use the upstream `podio/Frame.h` as all functionality is available from there.
- Increase the minimum podio version to 0.16.3

ENDRELEASENOTES


MarkusFrankATcernch:
@tmadlener Thomas, what should I do with these 2 MRs ?
They are very much ahead of their time, since the poidio in the hsf cvmfs is too old.....

tmadlener:
I mainly opened them to not forget about them. We can also hide them behind some pre processor versioning, and remove anything that requires pre v1.0 podio, once there is podio v1.0.

tmadlener:
Now that almost all CI environments have caught up with a recent enough version of podio. Should I still add some version `#if`s or is it OK to simply update the minimal version of podio that is necessary? (That is in any case already part of this PR).

MarkusFrankATcernch:
@tmadlener Should I merge ? The missing test is not required....

PR #1138, opened 2023-06-23
Fix DetElementCreator plugin 

tmadlener:
BEGINRELEASENOTES
- Make the default return value of `DetElementCreator::process` non-zero to avoid erroneously raising an exception after finishing. 
- Add a test case that checks that things work

ENDRELEASENOTES

~I am not entirely sure if the `DDDB` examples were turned off on purpose, there seem to be quite some test failures in there. It might be easier to move the new test case to somewhere else, I just put it into `DDDB` because it comes with a compact file that uses the `DetElementCreator` plugin. I can also move all the "noise" into a separate PR, so that this only contains what is really necessary.~

### The issue that is fixed by this
Without the (one line) fix, I run into the following issue:
```
[...]
RunPlugin        ERROR ++ Exception while executing plugin DD4hep_CompactLoader:
		dd4hep: apply-plugin: Failed to execute plugin DD4hep_PlacedVolumeProcessor
dd4hep: with plugin:DD4hep_PlacedVolumeProcessor
dd4hep: Error interpreting XML nodes of type <plugin/>
dd4hep: Error interpreting XML nodes of type <plugins/>
dd4hep: while parsing compact.xml
dd4hep: with plugin:DD4hep_CompactLoader
Application      ERROR geoPluginRun: Uncaught exception: RunPlugin: ++ Exception while executing plugin DD4hep_CompactLoader:
		dd4hep: apply-plugin: Failed to execute plugin DD4hep_PlacedVolumeProcessor
dd4hep: with plugin:DD4hep_PlacedVolumeProcessor
dd4hep: Error interpreting XML nodes of type <plugin/>
dd4hep: Error interpreting XML nodes of type <plugins/>
dd4hep: while parsing compact.xml
dd4hep: with plugin:DD4hep_CompactLoader
```

The exception is raised here:
https://github.com/AIDASoft/DD4hep/blob/2a492e2b8f2d6506c49cb7be80e0aeb10549874b/DDCore/src/DetectorImp.cpp#L854-L857

The return value in turn is set here:
https://github.com/AIDASoft/DD4hep/blob/2a492e2b8f2d6506c49cb7be80e0aeb10549874b/DDCore/src/plugins/StandardPlugins.cpp#L1406-L1411

Which finally calls `PlacedVolumeManager::process`
https://github.com/AIDASoft/DD4hep/blob/2a492e2b8f2d6506c49cb7be80e0aeb10549874b/DDCore/src/VolumeProcessor.cpp#L33

The `process` that is called in this case is `DetElementCreator::process`. One of the paths through this looks like this
https://github.com/AIDASoft/DD4hep/blob/2a492e2b8f2d6506c49cb7be80e0aeb10549874b/DDCore/src/gdml/DetElementCreator.cpp#L317-L331

the other looks like this
https://github.com/AIDASoft/DD4hep/blob/2a492e2b8f2d6506c49cb7be80e0aeb10549874b/DDCore/src/gdml/DetElementCreator.cpp#L422-L426

If I am not mistaken, this means that even if the `DetElementCreator` does nothing (i.e. automatic success?), it still returns 0. That at some point signals failure to the machinery. Hence, changing the default return value to 1, would fix this. **I am not entirely sure about all the assumptions in this, but this would fix things for me**.


MarkusFrankATcernch:
@tmadlener Did you re-enable DDDB ?
I thought DDDB was not built anymore. 
It is entirely obsolete.

tmadlener:
Ah, I wasn't aware of that. I will then move "my" test to somewhere non-obsolete. Any immediate suggestions for that?

MarkusFrankATcernch:
The generic container for all tests which are not really logically connected is "ClientTests".

tmadlener:
I moved the compact and necessary input file to ClientTests and added the test there. I also removed the previous changes to DDDB and also disabled `DD4HEP_BUILD_DEBUG` again for CI.

tmadlener:
New test seems to run and succeed, e.g.: https://github.com/AIDASoft/DD4hep/actions/runs/5355697069/jobs/9714291446?pr=1138#step:4:3663

PR #1140, opened 2023-07-03
Store the CellIDEncoding strings in the metadata Frame

tmadlener:
BEGINRELEASENOTES
- Create a separate `metadata` Frame in the edm4hep output and put the `CellIDEncoding` strings there in order to follow the new convention for this introduced by key4hep/k4FWCore#100.

ENDRELEASENOTES

If the whole `LazyEncodingExtraction` is too much of a premature optimization I can also take it out again and just put the interesting two lines into `saveCollection` directly.

I still have to think about an easy test for this (and where to put it).

### DIFF
--- a/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
+++ b/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
@@ -60,6 +60,7 @@ namespace dd4hep {
       stringmap_t                   m_eventParametersInt;
       stringmap_t                   m_eventParametersFloat;
       stringmap_t                   m_eventParametersString;
+      stringmap_t                   m_cellIDEncodingStrings{};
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
@tmadlener 
If we really ant to save extended metadata in the files, then these wild ad-hoc approaches at some point will lead
to chaotic data addings, where at the end nobody will know anymore how legally the data part should look like.
We will need some glossary about the keywords and the formatting of the data part.

tmadlener:
Yes, this convention needs to be properly documented in podio (at least). Ideally, we "hide" as much as possible of this behind some utility functionality, but we are not there yet. The problem is that the cellIDs need to be there otherwise many things will not work downstream, so for now I have to hardcode the convention in here to not block other things.

PR #1147, opened 2023-07-18
Propagate run Parameters from input to output

### LINKED ISSUE #1114: Persist HepMC3 GenRunInfo attributes into metadata tree
I would like to be able to save the attributes found in the header of hepmc3 files into the edm4hep output file.

The specific case is to keep track of the EIC beam settings used when creating the events so that the analysis can automatically choose the correct parameters in our reconstruction software.
[https://github.com/eic/epic/issues/433](https://github.com/eic/epic/issues/433)

andresailer:
BEGINRELEASENOTES
- Add reading of HepMC3 run info and forwarding it to output files: fixes #1114. Reliably this will only work for HepMC3 >= 3.2.6.

ENDRELEASENOTES

TODO:
- [x] Check that this works
- [x] Make it nicer
- [x] ~~Re-use EventParameters for RunParameter storage?~~ Cannot do that, cannot specialize the ingestParameters function more than once
- [x] Write release notes

### DIFF
--- a/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
+++ b/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
@@ -17,6 +17,7 @@
 #include <DD4hep/Detector.h>
 #include <DDG4/EventParameters.h>
 #include <DDG4/Geant4OutputAction.h>
+#include <DDG4/RunParameters.h>
 
 /// edm4hep include files
 #include <edm4hep/MCParticleCollection.h>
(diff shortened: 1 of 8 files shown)


andresailer:
As far as I understand, this will only work with HepMC3 >= 3.2.6 https://gitlab.cern.ch/hepmc/HepMC3/-/issues/44

MarkusFrankATcernch:
@andresailer This issue is highly connectedd to:  PR https://github.com/AIDASoft/DD4hep/pull/1140
Are you sure to add parameters as 4 tuples (string, int, float, double) with name-value pairs?
This to me looks quite limiting.
Would it not be better to have predefined xml snippets and then online store maps <string, string> ?
Such snippets could be:
```
<param name="author-name" type="string"  value="A.Sailer"/>
<param name="generator" type="string"  value="Jetset 8.6"/>
<param name="world-x" type="double" unit="m"  value="30"/>
<param name="world-y" type="double" value="30*m"/>
<param name="center-of-mass-energy" type="energy"  unit="Gev" value="5000.6"/>
<param name="ECAL:encoding" type="encoding"  value="system:8,barrel:3,module:4,layer:6,slice:5,x:32:-16,y:-16"/>
etc.
```

andresailer:
Hi Markus, this is related to #1114

I don't understand what you mean though about the 4 tuples.

andresailer:
So it probably also works with hepmc before 3.2.6, but the ReaderAscii only parses the run_info when actually reading events, so we skip(-1), which is otherwise a no-op, I hope also for all the other readers (it is for the Root Reader).

andresailer:
I think the skip(-1) is not working. I am confused.

MarkusFrankATcernch:
Negative skips can only work for direct access files (root etc.)
For sequential files this generally is difficult unless one wants to keep a  seek entry for every event accessed in the file.

andresailer:
> Negative skips can only work for direct access files (root etc.) For sequential files this generally is difficult unless one wants to keep a seek entry for every event accessed in the file.

I though the reader would parse all the text until it hits an event but not the event, but I misread the code and the (-1) was exiting even earlier.

https://gitlab.cern.ch/hepmc/HepMC3/-/blob/92f12d3d2af65c2e142c164fe3d7bed77aa766cc/src/ReaderAscii.cc#L54

Now I read an event and then create a new reader with the same file. How bad can this be :thinking:

MarkusFrankATcernch:
No. Look. This cannot be. If we made a design error we should correct it.

andresailer:
It is a problem with the hepmc reader, not in dd4hep.

andresailer:
Apparently, segfault is how bad this can be, so back to the drawing board.

andresailer:
The segfault was due to moving the reader creation to the beginRun only. But the tests were directly calling readparticles, without there being beginRun. Now a reader is created either at begin run, or in readParticles (as before).

Added also a HepMC Version check so that the code should compile...

andresailer:
As far as I am concerned, this is as good as it gets for the moment.

PR #1148, opened 2023-07-26
feat: convert several DDRec constructors to use `const Detector&`

wdconinc:
This started as an attempt to use a `CellIDPositionConverter` in a class that only has access to a `const Detector&`, but then turned in to figuring out a few other constructors where we can add const-ness.

The main hurdle is addressed in https://github.com/AIDASoft/DD4hep/commit/4c26c0375502cdd2bc6d94b44a7dfd8fac45eba8, where the former approach was returning the map entry `m_detectorTypes[type]` whenever `type` is not a valid key, creating a new entry in the process. This doesn't work when const.

I changed this to adding an empty key entry when `m_detectorTypes` is first filled, `m_detectorTypes[""] = {}`, and this one can be returned when `type` is not a valid key, without violating const-ness.

### Potential implications
If a sensitive detector doesn't specify a type, then `type()` returns `""` too,
```cpp
string DetElement::type() const {
  return m_element ? m_element->GetTitle() : "";
}
```
There doesn't seem to be any guards (that I could find) against someone defining a sensitive detector with an empty string as type. Feedback welcome. I'm filing this as draft first.

BEGINRELEASENOTES
- Allow several DDRec c'tors to work on `const Detector&` instead of `Detector&`

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/DetectorImp.cpp
+++ b/DDCore/src/DetectorImp.cpp
@@ -540,6 +540,7 @@ Material DetectorImp::material(const string& name) const {
 
 /// Internal helper to map detector types once the geometry is closed
 void DetectorImp::mapDetectorTypes()  {
+  m_detectorTypes[""] = {};
   for( const auto& i : m_detectors )   {
     DetElement det(i.second);
     if ( det.parent().isValid() )  { // Exclude 'world'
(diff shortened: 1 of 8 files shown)


DDCore/src/DetectorImp.cpp
```diff
@@ -571,15 +572,15 @@ vector<string> DetectorImp::detectorTypes() const  {
...
+const vector<DetElement>& DetectorImp::detectors(const string& type, bool throw_exc) const {
   if ( m_manager->IsClosed() ) {
     if ( throw_exc )  {
       DetectorTypeMap::const_iterator i=m_detectorTypes.find(type);
       if ( i != m_detectorTypes.end() ) return (*i).second;
       throw runtime_error("detectors("+type+"): Detectors of this type do not exist in the current setup!");
     }
     // return empty vector instead of exception
-    return m_detectorTypes[ type ] ;
+    return m_detectorTypes.at("") ;
```
andresailer, line 583: Now if throw_exc is false empty vector is always returned.
Rather
```c++
  static vector<DetElement> empty;
  if ( m_manager->IsClosed() ) {
    DetectorTypeMap::const_iterator i=m_detectorTypes.find(type);
    if ( i != m_detectorTypes.end() ) return (*i).second;
    if (throw_exc) {
      throw runtime_error("detectors("+type+"): Detectors of this type do not exist in   the current setup!");
     }
    // return empty vector instead of exception
    return empty;
  }
```

wdconinc: Thanks, I guess my mind had that logic but the code didn't ;-)

MarkusFrankATcernch:
@wdconinc Wouter, can we merge this PR ?

wdconinc:
Ah, yeah, I didn't realize this was marked as draft still.

PR #1151, opened 2023-08-02
Fix two compiler warnings

jmcarcell:
BEGINRELEASENOTES
- Limits object: fix string creation of Limit::toString, this will now also print the relevant "particles"
- Shapes: remove "move" from return of dimensions()

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/Objects.cpp
+++ b/DDCore/src/Objects.cpp
@@ -452,7 +452,7 @@ string Limit::toString() const {
   string res = name + " = " + content;
   if (!unit.empty())
     res += unit + " ";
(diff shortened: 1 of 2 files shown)


DDCore/src/Objects.cpp
```diff
@@ -452,7 +452,7 @@ string Limit::toString() const {
   string res = name + " = " + content;
   if (!unit.empty())
     res += unit + " ";
-  res + " (" + particles + ")";
+  res += " (" + particles + ")";
```
andresailer, line 455: This isn't a trivial change.

PR #1154, opened 2023-08-05
FastSimTests: increase number of events

andresailer:
Increase the number of events to check if we maybe were just lucky before, or if there really is an issue in the latest Geant4 version

BEGINRELEASENOTES
- Thank you for writing the text to appear in the release notes. It will show up
  exactly as it appears between the two bold lines
- ...

ENDRELEASENOTES


andresailer:
Tests fail with time out

Issue #1111, opened 2022-08-12
Writing multiple podio collections with the same name in DD4hep... 


wdconinc:
I don't really know where to post this issue exactly, since it seems to fall at the intersection of DD4hep, EDM4hep, and podio, but here goes... This is an issue reported on (if not properly reported on github) at one of the key4hep meetings in the spring. I'm also sorta debugging this again while I'm writing it, so forgive me for labeling this `[DRAFT]`.

In our DD4hep detector description, we use different `detector` sections to describe logically similar detectors. For example, the various endcap planes in our [tracker](https://github.com/eic/epic/blob/84e15195ab812157d0f95b0a06ec0c4d343528b6/compact/central_tracker_hybrid_v2.xml) are different in size and chip arrangement, yet they share all other characteristics and should logically be expected to write to a single hit collection, e.g. `TrackerEndcapHits`.

However, when writing to a single collection from multiple detectors, these collections are not read correctly by our reconstruction code (the actual issue, I'll come back to this).

So, as an ugly workaround we write to a slightly differently name collection for each detector (e.g. `TrackerEndcapHits1`, `TrackerEndcapHits2`, etc..., see these [readout definitions](https://github.com/eic/epic/blob/84e15195ab812157d0f95b0a06ec0c4d343528b6/compact/central_tracker_hybrid_v2.xml#L1395)). In reconstruction we then collate these again with a [SimTrackHitsCollector](https://github.com/eic/juggler/blob/861a64e312d29e1cf463c037ab085770044ee94d/JugDigi/src/components/SimTrackerHitsCollector.cpp). This is not really scalable, since our reconstruction steering scripts now have to depend on exactly how many planes we have (something we are still determining).

What appears to happen when we write to the single hits collection (as e.g. in [this branch](https://github.com/eic/epic/blob/6a2e484a4a1b3fc75a38cd6c82b7c21528690415/compact/central_tracker_hybrid_v2.xml#L1395)) is that we get simulated output files that contain a single collection, but the counters are off.

- [ ] add details on how the output files are 'off'.

Issue #1153, opened 2023-08-04
Error in GFLash test


andresailer:
- OS: any
 - compiler: any tested with dev3/dev4
 - Commit: master
 - Reproducer: building and testing against dev3 or dev4 LCG environment
 - Output: https://github.com/AIDASoft/DD4hep/actions/runs/5655190432/job/15595805064
 - Goal: the test shouldn't fail

```
[snip]
Gun              INFO  Particle [0] e+           Mom:50.000 GeV vertex:( 0.000  0.000  0.000)[mm] direction:(-0.634  0.485 -0.603)
ParticleHandler  INFO  +++ Event 0 Begin event action. Access event related information.
Gun              INFO  Particle [0] e+           Mom:50.000 GeV vertex:( 0.000  0.000  0.000)[mm] direction:(-0.202  0.068  0.977)
ParticleHandler  INFO  +++ Event 1 Begin event action. Access event related information.
Geant4Kernel     FATAL +++ Exception while simulating: BitFieldElement 'x': out of range : -2147483648 for width 12
Geant4Kernel     INFO  ++ Terminate Geant4 and delete associated actions.
RootOutput       INFO  +++ Closing ROOT output file SiliconBlock_GFlash_2023-08-03_19-49.root
```

andresailer:
Somewhere we get some NANs for the position of a hit
```
221: void dd4hep::sim::TrackerCombine::start(const dd4hep::sim::Geant4FastSimSpot*)
221: Spot  x=    22.0924 y=    2.24376 z=    249.554 energy 0
221: X ID-4
221: Y ID0
221: void dd4hep::sim::TrackerCombine::start(const dd4hep::sim::Geant4FastSimSpot*)
221: Spot  x=        nan y=        nan z=        nan energy 52.8525
221: X ID-2147483648
221: Y ID-2147483648

```

PR #1157, opened 2023-08-21
Geant4Output2EDM4hep: allow reuse of collection names... again

wdconinc:
[This](https://github.com/AIDASoft/DD4hep/issues/1111) seems to have gotten broken again [here](https://github.com/AIDASoft/DD4hep/blob/b784f3899b7cae7ef1e50b1b5b4cee6bbbfe61f8/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L476). We do NOT want to assign an empty collection since it overwrites the hits that are already in the collection; we just want to ensure that the named collection exists and the map should take care of the construction if it doesn't.

The ClientTest 'framework' doesn't easily allow for a unit test to prevent this from regressing (it's mostly based on regexes on failures), but please comment on ideas for tests there.

BEGINRELEASENOTES
- Geant4Output2EDM4hep: allow use of identical collection names across multiple detectors

ENDRELEASENOTES

### DIFF
--- a/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
+++ b/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
@@ -505,7 +505,7 @@ void Geant4Output2EDM4hep::saveCollection(OutputContext<G4Event>& /*ctxt*/, G4VH
   //-------------------------------------------------------------------
   if( typeid( Geant4Tracker::Hit ) == coll->type().type()  ){
     // Create the hit container even if there are no entries!
-    auto& hits = m_trackerHits[colName] = edm4hep::SimTrackerHitCollection();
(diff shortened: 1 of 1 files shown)


wdconinc:
Failing tests are in both cases `FATAL +++ Exception while simulating: BitFieldElement 'x': out of range : -2147483648 for width 12` and not related to this PR:
- https://github.com/wdconinc/DD4hep/actions/runs/5929991832/job/16078793371#step:4:4676
- https://github.com/wdconinc/DD4hep/actions/runs/5929991832/job/16078793179#step:4:5199

They also occur on master, e.g. https://github.com/AIDASoft/DD4hep/actions/runs/5761723352/job/15620106338#step:4:5201.

andresailer:
Yes, the test failures are under investigation in #1153

PR #1159, opened 2023-08-23
Improved _getEnviron to deal with multiple variables in the evaluated string

bcouturi:
Improved _getEnviron to deal with the case when multiple variables are to be evaluated in the string.
Previously it was only looking for the first ${ and the last }. I did a qucik fix that looks for all ${ } in the string. Using a parser or a library to do this would be better, but it seems that DDCore can be built without boost ?
 

BEGINRELEASENOTES
- Improved _getEnviron to deal with the case when multiple variables are to be evaluated in the string
ENDRELEASENOTES

### DIFF
--- a/DDCore/src/Handle.cpp
+++ b/DDCore/src/Handle.cpp
@@ -280,26 +280,41 @@ namespace dd4hep  {
   }
 
   /// Evaluate string constant using environment stored in the evaluator
-  string _getEnviron(const string& env)   {
-    size_t id1 = env.find("${");
-    size_t id2 = env.rfind("}");
-    if ( id1 == string::npos || id2 == string::npos )   {
-      return "";
-    }
-    else  {
-      stringstream err;
-      string v   = env.substr(id1,id2-id1+1);
-      auto   ret = eval.getEnviron(v, err);
-      if ( ret.first != tools::Evaluator::OK) {
-	cerr << env << ": " << err.str() << endl;
-	throw runtime_error("dd4hep: Severe error during environment lookup of " + env +
-			    " " + err.str());
+  string _getEnviron(const string& env)   {    
+    // We are trying to deal with the case when several variables are being replaced in the
+    // string.   
+    size_t current_index = 0;
+    stringstream processed_variable;
+    while (true) {
+      // Looking for the start of a variable use, with the syntax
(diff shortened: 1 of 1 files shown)


DDCore/src/Handle.cpp
```diff
@@ -280,26 +280,41 @@ namespace dd4hep  {
...
+    stringstream processed_variable;
+    while (true) {
+      // Looking for the start of a variable use, with the syntax
+      // "path1/path2/${VAR1}/path3/${VAR2}"
+      size_t id1 = env.find("${", current_index);
+      // variable start found, do a greedy search for the variable end 
+      if (id1 != string::npos) {
+        size_t id2 = env.find("}", id1);
+        if (id2 == string::npos) {
+          runtime_error("dd4hep: Syntax error, bas variable syntax: " + env);
```
andresailer, line 296:
```suggestion
          runtime_error("dd4hep: Syntax error, bad variable syntax: " + env); 
```

DDCore/src/Handle.cpp
```diff
@@ -280,26 +280,41 @@ namespace dd4hep  {
...
+    // We are trying to deal with the case when several variables are being replaced in the
+    // string.   
+    size_t current_index = 0;
+    stringstream processed_variable;
+    while (true) {
+      // Looking for the start of a variable use, with the syntax
+      // "path1/path2/${VAR1}/path3/${VAR2}"
+      size_t id1 = env.find("${", current_index);
+      // variable start found, do a greedy search for the variable end 
+      if (id1 != string::npos) {
```
andresailer, line 293:
```suggestion
      if (id1 == string::npos) {
        // In this case we did not find the ${ to indicate a start of variable,
        // we just copy the rest of the variable to the stringstream and exit
        processed_variable << env.substr(current_index);
        break;
       }
```
And then you don't need an `else, like you do for `id2`

bcouturi: Good idea

bcouturi (commented):
Included suggestions from A.Sailer

andresailer (approved):

andresailer:
Expected failures, and the "rich" again: 
`TEST: failed with average number of hits = 248.72 (expected 230+/-15)`

PR #1161, opened 2023-08-24
Added implementation for hexagonal segmentation and also staggered square segmentations

sebouh137:
Implemented segmentation variations with hexagonal and square cells, with and without staggering.  See  [arXiv:2308.06939](https://arxiv.org/abs/2308.06939)

BEGINRELEASENOTES
- DDSegmentation: added segementation HexGrid for hexagonal segmentation
- DDSegmentation: added CartesionGridXYStaggered for rectangular segmentation that can be staggered for every other layer

ENDRELEASENOTES


andresailer:
Hi @sebouh137 , many thanks for your contribution!

I am wondering if we could merge the SquareGrid with the existing CartesionGrid segmentation implementation?

Also it should be possible to use the "MultiSegmentation" and different offsets to achieve the same result as with the staggered SquareGrid?

Finally some example XML snippets would be good, a test that uses the hexgrid segmentation event better.

Thanks again!

DDCore/src/segmentations/HexGrid.cpp
```diff
@@ -0,0 +1,150 @@
+/*
```
andresailer, line 1: License notice missing

DDCore/src/segmentations/HexGrid.cpp
```diff
@@ -0,0 +1,150 @@
...
+ * HexGrid.cpp
+ *
+ *  Created on: August 9, 2023
+ *      Author: Sebouh J. Paul, UC Riverside
+ */
+#include "DD4hep/Factories.h"
+#include "HexGrid.h"
```
andresailer, line 8:
```suggestion
#include "DDSegmentation/HexGrid.h"
```

DDCore/src/segmentations/HexGrid.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+namespace dd4hep {
+  namespace DDSegmentation {
+
+    /// Default constructor used by derived classes passing the encoding string
+    HexGrid::HexGrid(const std::string& cellEncoding) :
+      Segmentation(cellEncoding) {
+      _type = "HexGridXY";
+	_description = "Hexagonal segmentation in the local XY-plane";
```
andresailer, line 17: Indentation is different here to previous line

DDCore/src/segmentations/HexGrid.cpp
```diff
@@ -0,0 +1,150 @@
...
+	  }
+	}
+	return cellPosition;
+    }
+
+    inline double positive_modulo(double i, double n) {
+      return std::fmod(std::fmod(i,n) + n,n);
+    }
+
+    //inline double positive_floor(
```
andresailer, line 83:
```suggestion
```

DDCore/src/segmentations/HexGrid.cpp
```diff
@@ -0,0 +1,150 @@
...
+	
+	double a=positive_modulo(y/(std::sqrt(3)*_sideLength),1);
+	double b=positive_modulo(x/(3*_sideLength),1);
+	int ix = std::floor(x/(3*_sideLength/2.))+		
+	  (b<0.5)*(-abs(a-.5)<(b-.5)*3)+(b>0.5)*(abs(a-.5)-.5<(b-1)*3);
+	int iy=std::floor(y/(std::sqrt(3)*_sideLength/2.));
+	iy-=(ix+iy)&1;
+
+	
+	//int ix=int(floor((localPosition.X - _offsetX) / (3/2*_sideLength)));
```
andresailer, line 125: remove commented code

DDCore/src/segmentations/SquareGrid.cpp
```diff
@@ -0,0 +1,105 @@
+/*
```
andresailer, line 1: License notice missing

DDCore/src/segmentations/SquareGrid.cpp
```diff
@@ -0,0 +1,105 @@
...
+	}
+	
+	return cellPosition;
+    }
+
+    inline double positive_modulo(double i, double n) {
+      return std::fmod(std::fmod(i,n) + n,n);
+    }
+
+    //inline double positive_floor(
```
andresailer, line 68:
```suggestion
```

andresailer (changes requested):

DDCore/src/segmentations/SquareGrid.cpp
```diff
@@ -0,0 +1,105 @@
+/*
+ *  SquareGrid.cpp
+ *  A square grid in X,Y which has options for staggering in the S2 pattern
+ *  Created on: August 9, 2023
+ *      Author: Sebouh J. Paul, UC Riverside
+ */
+#include "DD4hep/Factories.h"
+#include "SquareGrid.h"
```
andresailer, line 8:
```suggestion
#include "DDSegmentation/SquareGrid.h"
```

DDCore/src/segmentations/HexGrid.cpp
sebouh137: fixed with commit [84373a2](https://github.com/sebouh137/DD4hep/commit/84373a217539ac024beafacec1501327381224ab)

DDCore/src/segmentations/HexGrid.cpp
sebouh137: [84373a2](https://github.com/sebouh137/DD4hep/commit/84373a217539ac024beafacec1501327381224ab)

DDCore/src/segmentations/HexGrid.cpp
sebouh137: Added license info in [ade020d](https://github.com/sebouh137/DD4hep/commit/ade020d946c0fab99b8266a4a7823d225f314255)

DDCore/src/segmentations/SquareGrid.cpp
sebouh137: Added in commit [9b7833c](https://github.com/sebouh137/DD4hep/commit/9b7833c04431f588d29e13489c6ae65902e82534)

andresailer:
Hi @sebouh137 Any thoughts on what I proposed before https://github.com/AIDASoft/DD4hep/pull/1161#issuecomment-1692957930 ?

sebouh137:
Hi @andresailer.  

Sorry for the delay.  Here are my thoughts about your comments.  I'll get around to implementing these changes either by the end of this week or early next week.  

"I am wondering if we could merge the SquareGrid with the existing CartesionGrid segmentation implementation?"

I agree that would be much simpler.  When I was writing this code and testing it for a paper, I could have modified the CartesianGrid segmentation code, but it seemed easier at the time to write a new class as a plugin rather than re-compile the whole dd4hep code (especially since dd4hep was pre-compiled inside a container).  But yes it would be possible to replace the SquareGrid with an implementation of a staggering option in CartesianGrid, separately for x, y and z as appropriate.

"Also it should be possible to use the "MultiSegmentation" and different offsets to achieve the same result as with the staggered SquareGrid?"
I tried setting the offset fields in CartesianGrid to be a formula like offset_x="side_length*(layer%2)" in the xml file, but that caused  work.   Since I couldn't find any documentation on how to do that in the xml file, I ended up writing a segmentation class of my own.  I think adding staggering options for CartesianGrid would make more sense

"Finally some example XML snippets would be good, a test that uses the hexgrid segmentation event better."
That is a good idea.  

Thanks again!

DDCore/src/segmentations/CartesianGridXY.cpp
```diff
@@ -22,6 +22,8 @@ CartesianGridXY::CartesianGridXY(const std::string& cellEncoding) :
 	registerParameter("grid_size_y", "Cell size in Y", _gridSizeY, 1., SegmentationParameter::LengthUnit);
 	registerParameter("offset_x", "Cell offset in X", _offsetX, 0., SegmentationParameter::LengthUnit, true);
 	registerParameter("offset_y", "Cell offset in Y", _offsetY, 0., SegmentationParameter::LengthUnit, true);
+	registerParameter("stagger_x", "Option to stagger the layers in x (ie, add grid_size_x/2 to offset_x for odd layers)", _staggerX, 0);
+	registerParameter("stagger_y", "Option to stagger the layers in y (ie, add grid_size_y/2 to offset_y for odd layers)", _staggerY, 0);
```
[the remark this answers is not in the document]
sebouh137: ok thanks, I'll create a parameter called "stagger_keyword", and have that default to "layer".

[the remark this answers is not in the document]
andresailer: Thanks!

DDCore/include/DDSegmentation/HexGrid.h
```diff
@@ -0,0 +1,131 @@
...
+      void setFieldNameX(const std::string& fieldName) {
+        _xId = fieldName;
+      }
+      /// set the field name used for Y
+      void setFieldNameY(const std::string& fieldName) {
+        _yId = fieldName;
+      }
+      /// set the keyword used to determine which volumes to stagger
+      void setStaggerKeyword(const std::string& staggerKeyword) {
+        _staggerKeyword = staggerKeyword);
```
andresailer, line 98:
```suggestion
        _staggerKeyword = staggerKeyword;
```

DDCore/src/segmentations/HexGrid.cpp
```diff
@@ -0,0 +1,148 @@
...
+    inline double positive_modulo(double i, double n) {
+      return std::fmod(std::fmod(i,n) + n,n);
+    }
+
+    
+    /// determine the cell ID based on the position
+    CellID HexGrid::cellID(const Vector3D& localPosition, const Vector3D& /* globalPosition */, const VolumeID& vID) const {
+        CellID cID = vID ;
+	int layer= _decoder->get(cID,_staggerKeyword);
+	double _gridSizeY=std::sqrt(3)*_sideLength/2.;
```
andresailer, line 96: These variables are currently not used, but I think you can use them below?

sebouh137: I just removed the lines that define the unused variables _gridSizeY and _gridSizeX.

sebouh137:
@andresailer I have attached here a set of xml files that implement various combinations of hexagons and squares.  Thank you for your review of my pull request, and let me know if there's anything else that is needed to complete this.  
[xml_example.tar.zip](https://github.com/AIDASoft/DD4hep/files/12540730/xml_example.tar.zip)

sebouh137:
Dear @andresailer,
 I found where the problem with the check workflows failed this morning.  The test-suite must not have defined a field called "layer".  So I modified my code so that if staggering is not enabled, it will not attempt to find the value of "layer" (or whatever stagger keyword is provided).  Please rerun the check workflows again now that I've fixed this.  
Thanks, Sebouh

DDCore/src/segmentations/CartesianGridXY.cpp
```diff
@@ -51,16 +63,28 @@ CartesianGridXY::~CartesianGridXY() {
...
   CellID CartesianGridXY::cellID(const Vector3D& localPosition, const Vector3D& /* globalPosition */, const VolumeID& vID) const {
         CellID cID = vID ;
-	_decoder->set( cID,_xId, positionToBin(localPosition.X, _gridSizeX, _offsetX) );
-	_decoder->set( cID,_yId, positionToBin(localPosition.Y, _gridSizeY, _offsetY) );
+	if (_staggerX || _staggerY){
+		int layer= _decoder->get(cID,_staggerKeyword);
+		_decoder->set( cID,_xId, positionToBin(localPosition.X, _gridSizeX, _offsetX+_staggerX*_gridSizeX*(layer%2)/2) );
+		_decoder->set( cID,_yId, positionToBin(localPosition.Y, _gridSizeY, _offsetY+_staggerY*_gridSizeY*(layer%2)/2) );
+	} else {
+		_decoder->set( cID,_xId, positionToBin(localPosition.X, _gridSizeX, _offsetX);
```
andresailer, line 85:
```suggestion
		_decoder->set( cID,_xId, positionToBin(localPosition.X, _gridSizeX, _offsetX));
```

DDCore/src/segmentations/CartesianGridXY.cpp
```diff
@@ -51,16 +63,28 @@ CartesianGridXY::~CartesianGridXY() {
...
+		_decoder->set( cID,_yId, positionToBin(localPosition.Y, _gridSizeY, _offsetY);
```
andresailer, line 86:
```suggestion
		_decoder->set( cID,_yId, positionToBin(localPosition.Y, _gridSizeY, _offsetY));
```

DDCore/src/segmentations/CartesianGridXY.cpp
```diff
@@ -22,8 +22,14 @@ CartesianGridXY::CartesianGridXY(const std::string& cellEncoding) :
 	registerParameter("grid_size_y", "Cell size in Y", _gridSizeY, 1., SegmentationParameter::LengthUnit);
 	registerParameter("offset_x", "Cell offset in X", _offsetX, 0., SegmentationParameter::LengthUnit, true);
 	registerParameter("offset_y", "Cell offset in Y", _offsetY, 0., SegmentationParameter::LengthUnit, true);
+	registerParameter("stagger_x", "Option to stagger the layers in x (ie, add grid_size_x/2 to offset_x for odd layers)", _staggerX, 0,
+		SegmentationParameter::NoUnit, true);
+	registerParameter("stagger_y", "Option to stagger the layers in y (ie, add grid_size_y/2 to offset_y for odd layers)", _staggerY, 0, 
+		SegmentationParameter::NoUnit, true);
 	registerIdentifier("identifier_x", "Cell ID identifier for X", _xId, "x");
 	registerIdentifier("identifier_y", "Cell ID identifier for Y", _yId, "y");
+	registerParameter("stagger_keyword", "Volume ID identifier used for determining which volumes to stagger", _staggerKeyword, (std::string)"layer",
```
andresailer, line 31: why `registerParameter`, and not `registerIdentifier`?

sebouh137: If I understand it correctly, registerIdentifier creates a new field within the cellD.  Whereas what I am trying to do here is let the program know which existing volumeID field to use to determine whether to apply staggering within the given volume (e.g., by its layer number).

sebouh137:
Dear @andresailer ,
I am not sure what exactly is going on here, and I would like more details of what it's checking so that I can narrow down whatever I need to change in order to make it pass.  All of the tests pass, except for the four below.  What is it taking the checksum of, and what reference is it comparing them to? 
Thanks,
Sebouh

The following tests FAILED:
  	 22 - t_CLICSiD_check_checksum_EcalBarrel (Failed)
  	 23 - t_CLICSiD_check_checksum_full (Failed)
  	175 - t_MiniTel_check_checksum_Minitel3 (Failed)
  	176 - t_MiniTel_check_checksum_full (Failed)
  Errors while running CTest

DDCore/src/segmentations/CartesianGridXY.cpp
andresailer: Thanks for the explanation! That makes sense then.

andresailer:
Hi @sebouh137,

The checksum is a validity check for detector models to ensure that a detector model still looks exactly the same when changes are made in DD4hep or how DD4hep is compiled, for example when enabling GEANT4_UNITS. This checksum includes the segmentations, so with the modifications this is somehow changed. 

Is is probably OK to just update the expected checksums, but maybe @MarkusFrankATcernch can comment on this?

wdconinc:
The segmentation description in the checksum enumerates all parameters, https://github.com/AIDASoft/DD4hep/blob/094aa6cee2c0a38b7db4a9dc0b95f4759f6a1981/DDCore/src/plugins/DetectorChecksum.cpp#L897.

That is likely the cause for the change in checksum since new parameters were added to CartesianGridXY.

MarkusFrankATcernch:
So you changed the segmentation and consequently the checksum changed. No ?

This is in principle the expected behavior.
It is another question what should actually part of the checksum.
I normally believe that the segmentations should be part of them, because they actually are 
highly related to the readout geometry (sort of the segmentation approach in Geant4). 
This readout geometry is fundamental to a sub-detector and cannot be that easily changed either.
But maybe we want to have this configurable....

In principle the dump shows the various contributions to the checksum:
```+++ 1    Minitel1                             de fb9c568cca5fc642 +place    52327f747ebfed63 +daughters cbf29ce484222325 +children b56401ff46338125
  +++ 1                                                             +readout  a54220f77daa9eed +iddesc    8a34357f3cc91ee3 +segment  85d8e05343f2c736
  +++ 1    Minitel2                             de 26ddb5c697d5a3f2 +place    b1216132567d8f10 +daughters cbf29ce484222325 +children 52cc8ff6b960a770
  +++ 1                                                             +readout  8b859f9d89006082 +iddesc    53089002050c5a70 +segment  4446d2a5bd91f515
  +++ 1    Minitel3                             de a381e53d77452f9a +place    8461628938ce25d0 +daughters cbf29ce484222325 +children 30b6d41e344e570d
  +++ 1                                                             +readout  3edc69afe0b565c5 +iddesc    a0a7efb65254fea1 +segment  40ff023e9838d7b2
  +++ 0    world                                de b49f13e86137f0ac +place    18ff43e2fb898b14 +daughters cbf29ce484222325 +children bf1b5030a49d8dac
```

Are your changes only affecting the "segment" part of the elements?

sebouh137:
Dear @MarkusFrankATcernch @wdconinc and @andresailer ,
This is correct that I have added new parameters to CartesianGridXY.  The new parameters are options to enable "staggering" (ie, having a cyclical layer-dependent offset in x and y).  By default, staggering would is disabled, in order to avoid messing up existing geometries.  Further, these parameters are set as optional, so that no errors would occur from having them missing in such existing geometries.  
Thanks and best regards,
Sebouh Paul.

MarkusFrankATcernch:
@sebouh137 @andresailer @wdconinc 
From all I can see, there is no way to have a backwards compatible solution to this problem.
If we require backwards compatibility for the checksums, then the current changes to the XY segmentations
must move to the a hexagonal segmentation implementation.
The call 
```
      /// Access to all parameters
      virtual Parameters parameters() const;
```
in `DDSegmentation::Segmentation.h` returns all parameters - not just the relevant subset...
there are only two possibilities (both not backwards compatible):
- Segmentations are not part of the checksum (probably also not readouts and IDDescriptors, as they are related.
- Hexagonal segmentations may inherit from CartesianGridXY, but are a seperate implementation. This way existing CartesianGridXY segmentations stay intact.

Both is not really satisfactory, but I see no other way.

sebouh137:
Dear @MarkusFrankATcernch ,
In that case, I will leave HexGrid as I have implemented it.  I will remove changes from CartesianGridXY, and create a separate class "CartesianGridXYStaggered", which would have all of the staggering.  This should allow everything to be backwards compatible.

MarkusFrankATcernch:
@sebouh137
I know that code duplication is uncool, but I believe this is unfortunately the only way.

sebouh137:
> Dear @MarkusFrankATcernch ,
> In that case, I will leave HexGrid as I have implemented it. I will remove changes from CartesianGridXY, and create a separate class "CartesianGridXYStaggered", which would have all of the staggering. This should allow everything to be backwards compatible.

I have now implemented this change.  You can run the checks again now.  
Best regards,
Sebouh

DDCore/src/segmentations/CartesianGridXYStaggered.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+#if __cplusplus >= 201103L
+  return {_gridSizeX, _gridSizeY};
+#else
+  std::vector<double> cellDims(2,0.0);
+  cellDims[0] = _gridSizeX;
+  cellDims[1] = _gridSizeY;
+  return cellDims;
+#endif
+}
```
andresailer, line 100: This is missing the function declaration, and we can safely assume c++11 is available now.

```suggestion
std::vector<double> cellDimensions(const CellID& cellID) const {
  return {_gridSizeX, _gridSizeY};
}
```

sebouh137:
Dear @andresailer,
I see where this is failing to build.  While editing my files, I had forgotten to add "CartesianGridXYStaggered::" to the cellDimensions method in CartesianGridXYStaggered.cpp.  I think it should work now.  
Best regards,
Sebouh

andresailer:
> Dear @andresailer, I see where this is failing to build. While editing my files, I had forgotten to add "CartesianGridXYStaggered::" to the cellDimensions method in CartesianGridXYStaggered.cpp. I think it should work now. Best regards, Sebouh

Actually I forgot the `CartesianGridXYStaggered::` in https://github.com/AIDASoft/DD4hep/pull/1161/commits/dccd89f91faf28dbc4a9d70d2557e890be8b7ae2 , but the function declaration was missing before :smile: 

Thanks for adding the missing piece!

sebouh137:
Dear @andresailer ,
All of the changes you have requested have been marked as resolved, but it is still showing "one change requested".  I can't see any non-resolved change on here.  Further, all of the checks have now passed.  Let me know if there's anything else that you need on my end.  
Thanks,
Sebouh


image

andresailer (approved):

andresailer:
Thanks again for your contribution!

Issue #1194, opened 2023-12-05
DD4hep incompatible with Python 3.12


paulgessinger:
`dd4hep_base.py`, `dddigi.py` and `DDG4.py` use the `imp` module, which has been deprecated for about 9 years, and was finally removed in Python 3.12.

The module finding in `dd4hep_base.py` is trivial to replace, the dynamic module creation in `dddigi.py` and `DDG4.py` I'm not exactly sure about.

I only observed an error due to the top-level `import imp` in `dd4hep_base.py`. 
Is there a good way to test these dynamic lookups and module generation? 

Otherwise what's the best path forward here?

PR #1195, opened 2023-12-05
Replace imp module usage

paulgessinger:
`imp` was removed in Python 3.12.

See #1194 

BEGINRELEASENOTES
- Replace usage of the `imp` module that was removed in Python 3.12 with `importlib` and `types`.

ENDRELEASENOTES


MarkusFrankATcernch:
@paulgessinger Thanks for helping in the maintenance!

paulgessinger:
I think there might be a problem in lower python versions (from the CI output I wasn't sure what the test failures were exactly caused by). I'm trying to debug locally. The replacements I put in should be valid starting in python 3.4, so I don't think it's a fundamental error.

paulgessinger:
Hm, the pytest failures seem to be gone. There's now still two failures in ctest in `g4units`, but from a quick look they don't seem to be related to the python?

PR #1198, opened 2023-12-11
Set the edm4hep::MCParticle momentum as doubles once it becomes possible

tmadlener:
BEGINRELEASENOTES
- Set the momenta of the `edm4hep::MCParticle` as double instead of float once it becomes possible.

ENDRELEASENOTES

This should be backward compatible and keep working once key4hep/EDM4hep#237 is merged.


tmadlener:
Waiting for the EDM4hep change to go in first as that doesn't break DD4hep. This PR just avoids the precision loss from the explicit cast to `float` that is currently there.

tmadlener:
I think this can be merged now. I rebased it to trigger CI again.

MarkusFrankATcernch:
@tmadlener Should I merge ?

tmadlener:
Yes, this can be merged.

PR #1205, opened 2023-12-15
Geant4PrimaryHandling: fix issue with multiple vertices in Geant4 GeneralParticleSource

### LINKED ISSUE #1204: DDSim crashes when GPS use more than 1 source at once
- Package version: 1.21, local installation
 - short description: I was trying to run GPS, which allows to define several sources of primaries. It seems to work when only 1 source is called to generate primaries, but if all sources are fired at once when this command is added `/gps/source/multiplevertex true`, ddsim explodes.
 - Steps to reproduce the error:
```
git clone -b ddsim_bug_gps git@github.com:atolosadelgado/dd4hep_cube.git
cd dd4hep_cube
cmake -B build -S . -D CMAKE_INSTALL_PREFIX=install -DCMAKE_BUILD_TYPE=Release
cmake --build build -j 8 -- install
export LD_LIBRARY_PATH=$PWD/install/lib:$LD_LIBRARY_PATH
ddsim --compactFile ./compact/simple_detector.xml --runType run --enableG4GPS --macroFile test34.mac 
```
The macro file contain the definition of 2 sources of primaries, copypasted from `geant4/share/Geant4-11.0.2/examples/extended/eventgenerator/exgps/macros/test34.mac`

ddsim throws the following exception:
```
GenerationInit   INFO  +++ Initializing event 1. Within run:0 event 1.
Geant4PrimaryEvent ERROR +++ Interaction with ID '3' exists and cannot be added twice!
Geant4Kernel     FATAL +++ Exception while simulating:Geant4PrimaryEvent: +++ Interaction with ID '3' exists and cannot be added twice!
Geant4Kernel     INFO  ++ Terminate Geant4 and delete associated actions.
DDSim            INFO DDSim            INFO  Total Time:   4.28 s (User), 0.24 s (System)
```

Let me know if I can help further on this issue.

andresailer:
BEGINRELEASENOTES
- Geant4PrimaryHandling: fix issue with multiple vertices in Geant4 GeneralParticleSource, fixes #1204 

ENDRELEASENOTES

~~This is a little bit hacky, but as long as no-one goes over 10k primary sources we are OK.~~

There are actually two issues: One already pointed out in #1204 
which is triggered in
```c++
       prim->add(m_mask, inter);
```
Because we cannot add multiple vertices with the same mask

And with the m_mask replaced by an increasing maskCounter, we run into

https://github.com/AIDASoft/DD4hep/blob/3a6b9ff5b2463ceb780cb89c6f92ee7291ff49fd/DDG4/src/Geant4InputHandling.cpp#L120-L137

Where `vid` is always 0, because we take the length of a just created vector.

Better fixes would probably to switch the loops around in the Geant4GeneratorWrapper to create one interaction, which has multiple vertices, instead of multiple interactions with one vertex each.

But I want to see if these changes cause any other test failures as well, so I already open the PR

### DIFF
--- a/DDG4/src/Geant4GeneratorWrapper.cpp
+++ b/DDG4/src/Geant4GeneratorWrapper.cpp
@@ -77,13 +77,7 @@ void Geant4GeneratorWrapper::operator()(G4Event* event)  {
   // Now generate the new interaction
   generator()->GeneratePrimaryVertex(event);
 
-  /// Add all the missing interactions (primary vertices) to the primary event record.
-  int maskCounter = 100000 + m_mask;
-  for(G4PrimaryVertex* gv=event->GetPrimaryVertex(); gv; gv=gv->GetNext())  {
-    if ( primaries.find(gv) == primaries.end() )   {
-      Geant4PrimaryInteraction* inter = createPrimary(maskCounter, primaryMap, gv);
-      prim->add(maskCounter, inter);
-      maskCounter += 1;
-    }
-  }
+  // Add all the missing interactions (primary vertices) to the primary event record.
+  Geant4PrimaryInteraction* inter = createPrimary(m_mask, primaryMap, primaries);
+  prim->add(m_mask, inter);
 }
--- a/DDG4/src/Geant4GeneratorWrapper.cpp
+++ b/DDG4/src/Geant4GeneratorWrapper.cpp
@@ -70,12 +70,12 @@ void Geant4GeneratorWrapper::operator()(G4Event* event)  {
   Geant4PrimaryMap*   primaryMap = context()->event().extension<Geant4PrimaryMap>();
   set<G4PrimaryVertex*> primaries;
   
+  // Now generate the new interaction
+  generator()->GeneratePrimaryVertex(event);
+
   /// Collect all existing interactions (primary vertices)
   for(G4PrimaryVertex* v=event->GetPrimaryVertex(); v; v=v->GetNext())
     primaries.insert(v);
-
-  // Now generate the new interaction
-  generator()->GeneratePrimaryVertex(event);
 
   // Add all the missing interactions (primary vertices) to the primary event record.
   Geant4PrimaryInteraction* inter = createPrimary(m_mask, primaryMap, primaries);
--- a/DDG4/src/Geant4GeneratorWrapper.cpp
+++ b/DDG4/src/Geant4GeneratorWrapper.cpp
@@ -78,10 +78,12 @@ void Geant4GeneratorWrapper::operator()(G4Event* event)  {
   generator()->GeneratePrimaryVertex(event);
 
   /// Add all the missing interactions (primary vertices) to the primary event record.
+  int maskCounter = 100000 + m_mask;
   for(G4PrimaryVertex* gv=event->GetPrimaryVertex(); gv; gv=gv->GetNext())  {
     if ( primaries.find(gv) == primaries.end() )   {
-      Geant4PrimaryInteraction* inter = createPrimary(m_mask, primaryMap, gv);
-      prim->add(m_mask, inter);
+      Geant4PrimaryInteraction* inter = createPrimary(maskCounter, primaryMap, gv);
+      prim->add(maskCounter, inter);
+      maskCounter += 1;
     }
   }
 }
--- a/DDG4/src/Geant4InputHandling.cpp
+++ b/DDG4/src/Geant4InputHandling.cpp
@@ -88,6 +88,11 @@ static void collectPrimaries(Geant4PrimaryMap*         pm,
                              Geant4Vertex*             particle_origine,
                              G4PrimaryParticle*        gp)
 {
+  //if the particle is in the map, we do not have to do anything
+  if ( pm->get(gp) )  {
+    return;
+  }
+
   int pid = int(interaction->particles.size());
   Geant4Particle* p = createPrimary(pid,particle_origine,gp);
   G4PrimaryParticle* dau = gp->GetDaughter();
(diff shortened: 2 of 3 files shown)


MarkusFrankATcernch:
@andresailer 
This solution may destroys entirely the MC history mechanism as soon as there are multiple.
For the particle provenance every source must have it's own mask, independent of the primaries of each source.
If this is not the case, any later relationship between hits and particles for e.g. spill-over or multiple interactions
is just void.

-- Why is it not possible to add multiple primary vertices per interaction ?
   I cannot really see why.
-- If every GPS has its own mask, which it MUST have, this should work out just fine. Each GPS will be associated to a primary interaction and all particle contributions after simulation can be clearly disentangled.

andresailer:
Hi @MarkusFrankATcernch 

Thanks for your comments! But could you be more specific about where in the code you see an issue?

> For the particle provenance every source must have it's own mask, independent of the primaries of each source.

Every source still has its own mask, just that the GPS one has multiple masks now, or did I mess something up with my changes there?

> If this is not the case, any later relationship between hits and particles for e.g. spill-over or multiple interactions is just void.

There are still checks in the code that prevent multiple interactions with the same mask from happening.

> Why is it not possible to add multiple primary vertices per interaction? I cannot really see why.

Just more re-writing of the code, which has possibly other implications elsewhere.

> If every GPS has its own mask, which it MUST have, this should work out just fine. Each GPS will be associated to a primary interaction and all particle contributions after simulation can be clearly disentangled.

Maybe we need to define what we understand with GPS? The DD4hep incarnation of the GPS, or the Geant4 setup for the GPS? In the example macro from #1205 There is one Geant4 GPS with two different sources.

> Please do not merge.

No worries!

DDG4/src/Geant4InputHandling.cpp
```diff
@@ -103,14 +103,13 @@ static void collectPrimaries(Geant4PrimaryMap*         pm,
 
   if ( dau )   {
     Geant4Vertex* dv = new Geant4Vertex(*particle_origine);
-    int vid = int(interaction->vertices.size());
```
andresailer, line 106: This is the size of a map, seems a bit random to then assign vertices to that key, which may or may not be occupied. So I changed this to use "mask" instead below.

DDG4/src/Geant4InputHandling.cpp
```diff
@@ -121,18 +120,19 @@ static void collectPrimaries(Geant4PrimaryMap*         pm,
 Geant4PrimaryInteraction* 
 dd4hep::sim::createPrimary(int mask,
                            Geant4PrimaryMap* pm,
-                           const G4PrimaryVertex* gv)
+                           std::set<G4PrimaryVertex*>const& primaries)
 {
   Geant4PrimaryInteraction* interaction = new Geant4PrimaryInteraction();
-  Geant4Vertex* v = createPrimary(gv);
-  int vid = int(interaction->vertices.size());
```
andresailer, line 128: As `interaction` was created just above, this was always `0`, so I use mask instead.

DDG4/src/Geant4InputHandling.cpp
```diff
@@ -180,13 +180,14 @@ static void appendInteraction(const Geant4Action* caller,
   }
   Geant4PrimaryInteraction::VertexMap::iterator ivfnd, iv, ivend;
   for( iv=input->vertices.begin(), ivend=input->vertices.end(); iv != ivend; ++iv )   {
-    ivfnd = output->vertices.find((*iv).first) ; //(*iv).second->mask);
+    int theMask = input->mask;
+    ivfnd = output->vertices.find(theMask);
```
andresailer, line 184: We should now always use "mask" to add things to the `->vertices`.
But probably with the other changes, the changes here are no longer necessary.

MarkusFrankATcernch:
@andresailer 
I had some chat with Alvaro. So the result of these was:

- He wanted to shoot two(or more) close particles into the detector ans see which separation he can resolve.
- To solve this he constructed for Geant4 a combined particle source being the union of two individual particle sources.
- If he would have done this purely inside dd4hep, he would not even have seen this deficiency.

But he did not....
- In dd4hep the combined source appears as one single source giving multiple primary vertices. However, since it produces (unforeseen by me initially) multiple primary vertices, these get all attached to a primary interaction with the same mask. This effectively caused the exception.
- From the dd4hep philosophy one should assign multiple primary vertices of one source (even if this source is the union of other sources) to one primary interaction. Here is actually a bug present in dd4hep.

Now that I have all this explained and wanted to have a cross-check on the code, I see that you have already implemented all this according to the ideas above :-).....

DDG4/src/Geant4GeneratorWrapper.cpp
```diff
@@ -77,11 +77,7 @@ void Geant4GeneratorWrapper::operator()(G4Event* event)  {
...
-  /// Add all the missing interactions (primary vertices) to the primary event record.
-  for(G4PrimaryVertex* gv=event->GetPrimaryVertex(); gv; gv=gv->GetNext())  {
-    if ( primaries.find(gv) == primaries.end() )   {
-      Geant4PrimaryInteraction* inter = createPrimary(m_mask, primaryMap, gv);
-      prim->add(m_mask, inter);
-    }
-  }
+  // Add all the missing interactions (primary vertices) to the primary event record.
+  Geant4PrimaryInteraction* inter = createPrimary(m_mask, primaryMap, primaries);
+  prim->add(m_mask, inter);
```
MarkusFrankATcernch, line 82: To be added:
``` 
for(G4PrimaryVertex* gv=event->GetPrimaryVertex(); gv; gv=gv->GetNext())  {
    Geant4Vertex* v = createPrimary(gv);
   inter->vertices.emplace(v, gv);  
   for (G4PrimaryParticle *gp = gv->GetPrimary(); gp; gp = gp->GetNext() )
     collectPrimaries(pm, interaction, v, gp);
}
```

MarkusFrankATcernch:
Plese see comments in https://github.com/AIDASoft/DD4hep/pull/1205/files#diff-8b487a2446c759b09966b8f525fc745122717fa1576905209fe0e1a3a9b7433dL76

DDG4/src/Geant4GeneratorWrapper.cpp
andresailer: This doesn't compile, but I think most of that is already done in createPrimary?

Though, we will discuss in person and see.

Issue #1210, opened 2024-01-12
Changes in ROOT master TGeo break DD4hep compilation


andresailer:
(I strongly suspect) these changes https://github.com/root-project/root/pull/14327

causes these errors

https://lcgapp-services.cern.ch/cdash/upload/982612cdabd898ccd6b79a141231dcc6f0f5ede7/DD4hep-master-build.log

```
/build/jenkins/workspace/lcg_nightly_pipeline/build/frameworks/DD4hep-master/src/DD4hep/master/DDCore/src/ShapeUtilities.cpp:437:36: error: no member named 'GetVertexIndex' in 'TGeoFacet'
        pars.emplace_back(double(f.GetVertexIndex(j)));
                                 ~ ^
1 error generated.
make[5]: *** [DDCore/CMakeFiles/DDCore.dir/build.make:1336: DDCore/CMakeFiles/DDCore.dir/src/ShapeUtilities.cpp.o] Error 1
[snip]
/build/jenkins/workspace/lcg_nightly_pipeline/build/frameworks/DD4hep-master/src/DD4hep/master/DDCore/src/Volumes.cpp:364:29: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
  : TGeoExtension(c), magic(move(c.magic)), refCount(0), volIDs() {
                            ^
                            std::
/build/jenkins/workspace/lcg_nightly_pipeline/build/frameworks/DD4hep-master/src/DD4hep/master/DDCore/src/Volumes.cpp:366:12: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
  volIDs = move(c.volIDs);
           ^
           std::
```

MarkusFrankATcernch:
Change is not backwards compatible. Do you know by chance the current/latest ROOT version number ?

andresailer:
The latest ROOT release is 6.30.02.

PR #1211, opened 2024-01-12
Adapt to ROOT 6.31. See issue https://github.com/AIDASoft/DD4hep/issues/1210

MarkusFrankATcernch:
BEGINRELEASENOTES
- Implement required chnges for new ROOT 6.31.X.
  See issue https://github.com/AIDASoft/DD4hep/issues/1210 for details.
ENDRELEASENOTES

### DIFF
--- a/DDCore/src/ShapeUtilities.cpp
+++ b/DDCore/src/ShapeUtilities.cpp
@@ -434,7 +434,7 @@ namespace dd4hep {
       const TGeoFacet& f = sh->GetFacet(i);
       pars.emplace_back(double(f.GetNvert()));
       for(int j=0, n=f.GetNvert(); j<n; ++j)   {
-#if ROOT_VERSION_CODE >= ROOT_VERSION(6,31,1)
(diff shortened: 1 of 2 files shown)


andresailer:
Please note that the ROOT in dev3 does not yet contain the changes, so build failures for dev3 are to be expected.

MarkusFrankATcernch:
@andresailer Yes, but this should be taken into account by the version number. No?
I though hope in https://lcgapp-services.cern.ch/cdash/upload/982612cdabd898ccd6b79a141231dcc6f0f5ede7/DD4hep-master-build.log the version number is higher as in the commit....

andresailer:
I think the version number for the development branch is only updated after the was a release, regardless of the interface changes along the way.

So this is not going to change from Yesterday to today.

https://github.com/root-project/root/blob/2c00a76c8fa58352151f7c83da936138a0bf2085/core/foundation/inc/ROOT/RVersion.hxx#L7

PR #1212, opened 2024-01-13
ASSIMP: adapt for changes in TGeoTesselated in ROOT

andresailer:
More issues https://lcgapp-services.cern.ch/cdash/upload/dc90a57b6947469d59a9ba12f3f24c333bce0c8a/DD4hep-master-build.log

Follow up for #1211 

BEGINRELEASENOTES
- Assimp: adapt to changes in TGeoTessellated for ROOT 6.32
- ShapeUtilities: adapt to changes in TGeoTessellated for ROOT 6.32

ENDRELEASENOTES

Also need https://github.com/root-project/root/pull/14339

PR #1215, opened 2024-01-17
Fix creation of tessellated shapes from CAD files.

MarkusFrankATcernch:
BEGINRELEASENOTES
- Fix creation of tessellated shapes from CAD files.
-  Properly hand quadri-linear facets from CAD files.
-  See also: https://github.com/root-project/root/pull/14327
-  See also: https://github.com/AIDASoft/DD4hep/pull/1212
ENDRELEASENOTES


MarkusFrankATcernch:
@andresailer FYI
- I had to go back to TGeo's `ComputeNormal` implementation to get rid of the errors.
- Also facets may only be added using the true vertices (double triplets). TGeoTessellated does then the correct thing and kills duplicated vertices from assimp.

PR #1218, opened 2024-01-25
feat: allow users of `ddsim` to set e.g. stepping action

wdconinc:
For some studies we use DDG4 stepping (etc) action plugins in `ddsim` runs. For example, we are using this for fluence calculations by keeping track of where steps are activated.

BEGINRELEASENOTES
- add command line options to pass stepping (etc) action plugins to ddsim

ENDRELEASENOTES


andresailer:
Thanks for this!

Is it likely or possible that one would like to configure properties for these actions?

wdconinc:
It probably should at least be possible, but because the plugins can be anything I don't know what the best way to pass these options would be. Is this best done with a dict?

andresailer:
For configuration something similar is done with the filters, if I remember correctly. Either just a name, like now, or a tuple of name and dict with properties.

wdconinc:
I'll implement this with parameters, otherwise it's pretty limited in its usefulness I'm realizing.

wdconinc:
I added the ability to specify parameters, and modified Geant4TestActions's Geant4TestBase to print out its properties on destructions to be able to test this. Now we can do:
```
ddsim --compactFile $DD4hepINSTALL/DDDetectors/compact/SiD.xml -G -N 10 --action.event '{ "name": "Geant4TestEventAction/EventAction0", "parameter" : {"Property_int": 10} }' --action.stack '{ "name": "Geant4TestStackAction/StackAction1", "parameter" : {"Property_int": 11} }' --printLevel VERBOSE
```
and it will print
```
...
EventAction0     VERB  Geant4Action: Deleting object EventAction0 of type dd4hep::sim::Test::Geant4TestEventAction Pointer:0x561856e3d560
Geant4TestEventAction VERB  properties at destruction: 10, 0.000000, 
...
StackAction1     VERB  Geant4Action: Deleting object StackAction1 of type dd4hep::sim::Test::Geant4TestStackAction Pointer:0x561856ef63d0
Geant4TestStackAction VERB  properties at destruction: 11, 0.000000, 
...
```

What I haven't been able to make work yet is something with two event actions, i.e. like this:
```
ddsim --compactFile $DD4hepINSTALL/DDDetectors/compact/SiD.xml -G -N 10 --action.event '{ "name": "Geant4TestEventAction/EventAction0", "parameter" : {"Property_int": 10} }' --action.event '{ "name": "Geant4TestEventAction/EventAction1", "parameter" : {"Property_int": 11} }' --printLevel VERBOSE
```
This just activates the last action only. I'm probably overlooking something obvious, but I'll look again later. For now, adding multiple event actions through a list is also possible:
```
ddsim --compactFile $DD4hepINSTALL/DDDetectors/compact/SiD.xml -G -N 10 --action.event '[ { "name": "Geant4TestEventAction/EventAction0", "parameter" : {"Property_int": 10} }, { "name": "Geant4TestEventAction/EventAction1", "parameter" : {"Property_int": 11} } ]' --printLevel VERBOSE
```
which does give
```
...
EventAction0     VERB  Geant4Action: Deleting object EventAction0 of type dd4hep::sim::Test::Geant4TestEventAction Pointer:0x55838d449fc0
Geant4TestEventAction VERB  properties at destruction: 10, 0.000000, 
...
EventAction1     VERB  Geant4Action: Deleting object EventAction1 of type dd4hep::sim::Test::Geant4TestEventAction Pointer:0x55838d4bd150
Geant4TestEventAction VERB  properties at destruction: 11, 0.000000, 
...
```

TODO:
- [ ] figure out how to make repeated actions be additive instead of overwriting,
- [ ] is there a benefit to put the JSON parsing into ConfigHelpers?

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -123,15 +123,14 @@ def calorimeterSDTypes(self):
...
-
   @staticmethod
   def makeListOfDictFromJSON(val):
     if isinstance(val, str):
       # assumes: valid JSON string or comma-separated list of names
       import json
       try:
         val = json.loads(val)
-      except:
+      except ValueError as e:
```
wdconinc, line 133:
```suggestion
      except ValueError:
```

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -108,3 +122,71 @@ def calorimeterSDTypes(self):
...
+
+  @staticmethod
+  def makeListOfDictFromJSON(val):
+    if isinstance(val, str):
+      # assumes: valid JSON string or comma-separated list of names
+      import json
+      try:
+        val = json.loads(val)
+      except ValueError:
+        val = tuple(val.split(","))
```
wdconinc, line 134: Upon reflection, this is still incorrect and should be:
```suggestion
        val = [dict(name=v) for v in val.split(",")]
```

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -347,6 +347,20 @@ def run(self):
...
+         (self.action.event, DDG4.EventAction, kernel.eventAction),
+         (self.action.track, DDG4.TrackingAction, kernel.trackingAction),
+         (self.action.step, DDG4.SteppingAction, kernel.steppingAction),
+         (self.action.stack, DDG4.StackingAction, kernel.stackingAction)]:
+      for action_dict in action_list:
+        action = DDG4_Action(kernel, action_dict["name"])
+        for parameter, value in action_dict['parameter'].items():
+          setattr(action, parameter, value)
+        kernel_Action().add(action)
+
```
wdconinc, line 363: We may want to `kernel.registerGlobalAction(action)` here to allow access by name later on(?).

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -30,6 +30,15 @@ class Action(ConfigHelper):
...
   >>> SIM.action.mapActions['ecal'] =( "CaloPreShowerSDAction", {"FirstLayerNumber": 1} )
 
+  Additional actions can be set as well with
+
+  >>> SIM = DD4hepSimulation()
+  >>> SIM.action.run = [ ("Geant4TestRunAction", {"Property_int": 10} ) ]
+  >>> SIM.action.event = [ ("Geant4TestEventAction", {"Property_int": 10} ) ]
+  >>> SIM.action.track = [ ("Geant4TestTrackAction", {"Property_int": 10} ) ]
+  >>> SIM.action.step = [ ("Geant4TestStepAction", {"Property_int": 10} ) ]
+  >>> SIM.action.stack = [ ("Geant4TestStackAction", {"Property_int": 10} ) ]
```
wdconinc, line 40:
```suggestion
  Additional actions can be set as well with the following syntax variations:

  >>> SIM = DD4hepSimulation()
  >>> SIM.action.run = "Geant4TestRunAction" # name only
  >>> SIM.action.event = "Geant4TestEventAction/EventAction0,Geant4TestEventAction/EventAction1" # comma-separated names
  >>> SIM.action.track = ( "Geant4TestTrackAction", {"Property_int": 10} ) # tuple of name and parameter dict
  >>> SIM.action.step = { "name": "Geant4TestStepAction", "parameter": {"Property_int": 10} } # dict of name and parameter dict
  >>> SIM.action.stack = [ { "name": "Geant4TestStackAction", "parameter": {"Property_int": 10} } ] # list of dict of name and parameter dict

On the command line, these can be specified as JSON strings.
```

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -30,6 +30,22 @@ class Action(ConfigHelper):
...
   >>> SIM.action.mapActions['ecal'] =( "CaloPreShowerSDAction", {"FirstLayerNumber": 1} )
 
+  Additional actions can be set as well with the following syntax variations:
+
+  >>> SIM = DD4hepSimulation()
+  # single action by name only:
+  >>> SIM.action.run = "Geant4TestRunAction"
+  # multiple actions with comma-separated names:
+  >>> SIM.action.event = "Geant4TestEventAction/Action0,Geant4TestEventAction/Action1"
+  # single actuon by tuple of name and parameter dict:
```
andresailer, line 40:
```suggestion
  # single action by tuple of name and parameter dict:
```

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -30,6 +30,22 @@ class Action(ConfigHelper):
   >>> SIM = DD4hepSimulation()
   >>> SIM.action.mapActions['ecal'] =( "CaloPreShowerSDAction", {"FirstLayerNumber": 1} )
 
+  Additional actions can be set as well with the following syntax variations:
+
+  >>> SIM = DD4hepSimulation()
+  # single action by name only:
+  >>> SIM.action.run = "Geant4TestRunAction"
+  # multiple actions with comma-separated names:
+  >>> SIM.action.event = "Geant4TestEventAction/Action0,Geant4TestEventAction/Action1"
```
andresailer, line 39: I think we should just add the `/Action` ourselves when we create the action, that unifies the setting of a single or multiple actions.

andresailer: We should probably also allow a list of strings `["MyFirstGeant4TestEventAction", "MySecondGeant4TestEventAction"]`

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -108,3 +129,71 @@ def calorimeterSDTypes(self):
...
+
+  @staticmethod
+  def makeListOfDictFromJSON(val):
+    if isinstance(val, str):
+      # assumes: valid JSON string or comma-separated list of names
+      import json
+      try:
+        val = json.loads(val)
+      except ValueError:
+        val = [dict(name=v) for v in val.split(",")]
```
andresailer, line 141:
```suggestion
        val = [dict(name=v, parameter={}) for v in val.split(",")]
```

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -347,6 +347,20 @@ def run(self):
...
+    # Configure run, event, track, step, and stack actions, if present
+    for action_list, DDG4_Action, kernel_Action in \
+        [(self.action.run, DDG4.RunAction, kernel.runAction),
+         (self.action.event, DDG4.EventAction, kernel.eventAction),
+         (self.action.track, DDG4.TrackingAction, kernel.trackingAction),
+         (self.action.step, DDG4.SteppingAction, kernel.steppingAction),
+         (self.action.stack, DDG4.StackingAction, kernel.stackingAction)]:
+      for action_dict in action_list:
+        action = DDG4_Action(kernel, action_dict["name"])
+        for parameter, value in action_dict['parameter'].items():
```
andresailer, line 360: We have to ensure that "parameter" is a valid key, so that means the json string _must_ include a parameter dict even if it that dict is empty, right?

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -30,6 +30,22 @@ class Action(ConfigHelper):
...
+  # multiple actions with comma-separated names:
+  >>> SIM.action.event = "Geant4TestEventAction/Action0,Geant4TestEventAction/Action1"
+  # single action by tuple of name and parameter dict:
+  >>> SIM.action.track = ( "Geant4TestTrackAction", {"Property_int": 10} )
+  # single action by dict of name and parameter dict:
+  >>> SIM.action.step = { "name": "Geant4TestStepAction", "parameter": {"Property_int": 10} }
+  # multiple actions by list of dict of name and parameter dict:
+  >>> SIM.action.stack = [ { "name": "Geant4TestStackAction", "parameter": {"Property_int": 10} } ]
+
+On the command line, these can be specified as JSON strings.
```
andresailer, line 47: An example JSON string would be great

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -108,3 +129,71 @@ def calorimeterSDTypes(self):
...
+      # assumes: valid JSON string or comma-separated list of names
+      import json
+      try:
+        val = json.loads(val)
+      except ValueError:
+        val = [dict(name=v) for v in val.split(",")]
+    if isinstance(val, tuple):
+      # assumes: ( "Geant4TestEventAction", {"Property_int": 10} )
+      # creates: { "name": "Geant4TestEventAction", "parameter": {"Property_int": 10} }
+      # note: not able to specified as json which only allows a list
```
andresailer, line 145:
```suggestion
      # note: not able to be specified as json which only allows a list
```

DDG4/python/DDSim/DD4hepSimulation.py
wdconinc: Yes, I have modified this locally already with an additional `if "parameter" in keys()` to avoid this. That avoids the need to define an empty parameter dict.

DDG4/python/DDSim/Helper/Action.py
wdconinc: The issue with a list of strings is that it interferes with the list of dicts if we just use `isinstance` checks... It would make `makeListOfDictFromJSON` more complicated: instead of
```python
    if isinstance(val, list):
      # assumes: [ { "name": "Geant4TestEventAction", "parameter": {"Property_int": 10} } ]
      return val
```
it would become
```
    if isinstance(val, list):
      if isinstance(val[0], str):
        # assumes: [ "Geant4TestEventAction", "Geant4TestEventAction" ]
        return [dict(name=v) for v in val]
      if isinstance(val[0], dict):
        # assumes: [ { "name": "Geant4TestEventAction", "parameter": {"Property_int": 10} } ]
        return val
```
and then we need to check that `val != []`... I guess not that bad.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -347,6 +347,21 @@ def run(self):
...
+        [(self.action.run, DDG4.RunAction, kernel.runAction),
+         (self.action.event, DDG4.EventAction, kernel.eventAction),
+         (self.action.track, DDG4.TrackingAction, kernel.trackingAction),
+         (self.action.step, DDG4.SteppingAction, kernel.steppingAction),
+         (self.action.stack, DDG4.StackingAction, kernel.stackingAction)]:
+      for action_dict in action_list:
+        action = DDG4_Action(kernel, action_dict["name"])
+        if 'parameter' in action_dict.keys():
+          for parameter, value in action_dict['parameter'].items():
+            setattr(action, parameter, value)
```
andresailer, line 362:
```suggestion
        for parameter, value in action_dict.get('parameter', {}).items():
          setattr(action, parameter, value)
```

Also
```python
if 'key' in myDict:
```
is sufficient, no need to call `keys()`

DDG4/python/DDSim/Helper/Action.py
```diff
@@ -108,3 +140,71 @@ def calorimeterSDTypes(self):
...
+      # creates: { "name": "Geant4TestEventAction", "parameter": {"Property_int": 10} }
+      # note: not able to be specified as json which only allows a list
+      val = dict(name=val[0], parameter=val[1])
+    if isinstance(val, dict):
+      # assumes: { "name": "Geant4TestEventAction", "parameter": {"Property_int": 10} }
+      # creates: [ { "name": "Geant4TestEventAction", "parameter": {"Property_int": 10} } ]
+      val = [val]
+    if isinstance(val, list):
+      # assumes: [ { "name": "Geant4TestEventAction", "parameter": {"Property_int": 10} } ]
+      return val
```
wdconinc, line 164:
```suggestion
    if isinstance(val, list):
      if not val:
        # empty list
        return []
      if isinstance(val[0], str):
        # assumes: [ "Geant4TestEventAction", "Geant4TestEventAction" ]
        return [dict(name=v) for v in val]
      if isinstance(val[0], dict):
        # assumes: [ { "name": "Geant4TestEventAction", "parameter": {"Property_int": 10} } ]
        return val
```

wdconinc: This addresses https://github.com/AIDASoft/DD4hep/pull/1218/files#r1478410643.

DDG4/python/DDSim/Helper/Action.py
wdconinc: Resolved in 62b7b14e6f2c764e17ee862be8dfc24b8f6cb773.

wdconinc:
Sorry for the slow progress on this (heavy teaching load this term).

This now passes the unit test, BUT
- CTest is unable to check for multiple required regex patterns in the output: `set_tests_properties(PASS_REGULAR_EXPRESSION ${list})` passes as soon as one match is found; since order is not guaranteed, I was going with a negative lookahead (i.e. `(?=Deleting object RunAction1)(?=Deleting object EventAction1)`, but apparently CMake implements its own regex parser without support for negative lookahead. No idea how to implement a unit test that asserts multiple distinct lines should be present in the output... So, the unit test is just testing whether it runs, not that it does all that's expected based on the command line arguments.
- Due to the above, the unit test doesn't catch the following:
  - an option specified in the steering file and NOT on the command line is duplicated (applies to `RunAction1` in the unit test); it gets added first in the `exec` of the steering file, and then again by `__parseAllHelper`.
  - an option specified multiple times on the command line only gets added for the last instance (applies to `StepActionCLI1`, which isn't added while `StepActionCLI3` is added).

wdconinc:
After https://github.com/AIDASoft/DD4hep/pull/1218/commits/c452e5ada237bcac56f7f010a2769f6af67768ed no more duplication of inputs, and no more missing arguments.

PR #1219, opened 2024-02-01
Make the renaming of TrackerHit -> TrackerHit3D transparent

tmadlener:
BEGINRELEASENOTES
- Make the upcoming renaming of `edm4hep::TrackerHit` to `edm4hep::TrackerHit3D` (https://github.com/key4hep/EDM4hep/pull/252) transparent as far as DD4hep is concerned.

ENDRELEASENOTES

I am using the existence of the new header file to detect which version of the TrackerHit to use and just typedef to the new type name in case it's not yet present. In my local testing this works for current EDM4hep as well as for the version with the above PR merged.


MarkusFrankATcernch:
@tmadlener Should I merge ?

PR #1225, opened 2024-02-11
fix(ci): checkout@v4, cvmfs@v4, upload-artifact@v4

wdconinc:
This PR updates the versions of CI actions to use node.js 20 and avoid warnings in the CI interface here on GitHub Actions. Upload-artifact does not support writing from multiple jobs to a single artifact (automatic merging) ~~but that is not used here as far as I can tell~~ and the g4units test artifact was renamed to avoid clashing.

BEGINRELEASENOTES
- upgrade CI to checkout@v4, cvmfs@v4, upload-artifact@v4

ENDRELEASENOTES


wdconinc:
The remaining warnings in LCG_100 through LCG_102 are due to the older version of cmake which does not support `--output-junit`.

PR #1228, opened 2024-02-13
Fix DD4hep_DetectorDump plugin to avoid crash. 

MarkusFrankATcernch:
BEGINRELEASENOTES
- Fix DD4hep_DetectorDump plugin to avoid crash if a DetElement placement is invalid,
   which is not allowed, but also should not end up in a crash.
- Add test to check if DetElements have proper placement set using MiniTel example
  (see examples/ClientTests/compact/MiniTel_err_place.xml) and
  test MiniTel_check_missing_placements
- The fake DD4hep object TwistedTube  did not correctly emulate the different constructors when converted
  to Geant4. This should be fixed with this PR.
ENDRELEASENOTES

### DIFF
--- a/DDCore/src/plugins/StandardPlugins.cpp
+++ b/DDCore/src/plugins/StandardPlugins.cpp
@@ -1480,15 +1480,17 @@ template <int flag> long dump_detelement_tree(Detector& description, int argc, c
       }
     }
     IDDescriptor get_id_descriptor(PlacedVolume pv)   {
-      Volume v = pv.volume();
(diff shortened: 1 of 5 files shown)


MarkusFrankATcernch:
@bcouturi 
This should fix the crash you got in the detector dump:
https://gitlab.cern.ch/lhcb/Detector/-/merge_requests/497
Now the path of the detector element should be printed insead like:
```
DetectorDump     ERROR 004         /MyLHCBdetector1/side_0/module_9 DetElement with INVALID PLACEMENT!
```

PR #1230, opened 2024-02-14
XML: add std:: when building with tinyXML

### LINKED ISSUE #1229: Extensions got broken with #1226
Originally reported in https://github.com/key4hep/k4geo/issues/319

For ILD there is a **long list* of the following:

```console
[...]
ObjectExtensions::extension ERROR The object has no extension of type FB10DC132C7B13D5.
ObjectExtensions::extension ERROR The object has no extension of type FB10DC132C7B13D5.
ObjectExtensions::extension ERROR The object has no extension of type FB10DC132C7B13D5.
[...]
```

I have tested locally and this appeared after #1226 but was not present before. Things seem to still be running, but I am not yet sure whether the results make sense or whether it's only not crashing.

andresailer:
BEGINRELEASENOTES
- XML: add missing std:: when building with TinyXML
- DetectorSurfaces: silence printout when creating SurfaceList, fixes #1229 

ENDRELEASENOTES

### DIFF
--- a/DDRec/src/DetectorSurfaces.cpp
+++ b/DDRec/src/DetectorSurfaces.cpp
@@ -35,8 +35,10 @@ namespace dd4hep {
       const VolSurfaceList* vsL = volSurfaceList(det) ;
 
       try {
-        _sL = det.extension< SurfaceList >() ;
-
+        _sL = det.extension< SurfaceList >(false) ;
+        if (not _sL) {
+          _sL = det.addExtension<SurfaceList >(  new SurfaceList( true )  ) ;
+        }
       } catch(const std::exception& e) { 
         _sL = det.addExtension<SurfaceList >(  new SurfaceList( true )  ) ; 
       }
--- a/DDCore/src/XML/DocumentHandler.cpp
+++ b/DDCore/src/XML/DocumentHandler.cpp
@@ -508,7 +508,7 @@ Document DocumentHandler::load(const std::string& fname, UriReader* reader) cons
       except("dd4hep:XML","++ Unknown error (TinyXML) while parsing:%s",fname.c_str());
     }
   }
-  catch(exception& e) {
+  catch(std::exception& e) {
     printout(ERROR,"DocumentHandler","+++ Exception (TinyXML): parse(path):%s",e.what());
   }
   if ( result ) {
--- a/DDCore/src/XML/XMLElements.cpp
+++ b/DDCore/src/XML/XMLElements.cpp
@@ -989,13 +989,13 @@ static unsigned int adler32(unsigned int adler, const XmlChar* xml_buff, size_t 
 typedef unsigned int (fcn_t)(unsigned int, const XmlChar*, size_t);
 unsigned int Handle_t::checksum(unsigned int param, fcn_t fcn) const {
 #ifdef DD4HEP_USE_TINYXML
-  typedef map<std::string, std::string> StringMap;
+  typedef std::map<std::string, std::string> StringMap;
   TiXmlNode* n = Xml(m_node).n;
   if ( n ) {
     if ( 0 == fcn ) fcn = adler32;
     switch (n->Type()) {
     case TiXmlNode::ELEMENT: {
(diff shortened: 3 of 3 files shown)


DDRec/src/DetectorSurfaces.cpp
```diff
@@ -35,8 +35,10 @@ namespace dd4hep {
       const VolSurfaceList* vsL = volSurfaceList(det) ;
 
       try {
-        _sL = det.extension< SurfaceList >() ;
-
+        _sL = det.extension< SurfaceList >(false) ;
+        if (not _sL) {
+          _sL = det.addExtension<SurfaceList >(  new SurfaceList( true )  ) ;
+        }
```
tmadlener, line 41: I still see error printouts, even with these changes. I haven't seen any runtime issues yet with ILD reco, I have to still check results.

andresailer: Which k4geo test are you running?

tmadlener: I am running the ILD standard reco, where this also pops up after #1226

andresailer: It might call a different `extension<>()`
Can you pinpoint via gdb?

tmadlener: In ILD reconstruction the issue is this extension here: https://github.com/iLCSoft/DDKalTest/blob/81e946b08c3d244c762714e707b640dae0b8152d/src/DDPlanarMeasLayer.cc#L110

That in turn is defined here:
https://github.com/AIDASoft/DD4hep/blob/b6f660a95bcd6270083d34443cd2fbbca4006e97/DDRec/include/DDRec/DetectorData.h#L505

I couldn't immediately find a similar hook as for the SurfaceList, but someone with a bit more knowledge about the internals probably has more success.

MarkusFrankATcernch:
Thomas,
I think this change in the object extensions should be reverted. 
If this is the only thing going wrong, then the change was not 
correct, because I though on failure always a message should be printed.
Should I revert ?

DDRec/src/DetectorSurfaces.cpp
andresailer:
```diff
- auto* ext = detElement.extension<dd4hep::rec::DoubleParameters>();
+ auto* ext = detElement.extension<dd4hep::rec::DoubleParameters>(false);
```
and then `if(ext)` instead of `catch`?

tmadlener: If I have a `false` in there, what happens with previous versions of DD4hep? IIUC, I still get an exception, right? So, I would only have to add an additional nullptr check to have it silent for newer versions?!

tmadlener:
I think Andres suggestion would make sense in this case. After all, the try-catch for this specific case is effectively there for a runtime check and not necessarily an error. 

On the other hand I think it might be used like this in a few other cases as well, so there might be more "error messages" popping up in other places that are not necessarily an error.

At least for me it would be OK to leave it in, but I have no idea on how many times this will still show up in other places.

DDRec/src/DetectorSurfaces.cpp
andresailer: It still works in previous versions of DD4hep with the `false` and nullptr check. The `if (!alert) return 0;` was already present before.

andresailer: PS: for ten years
image

MarkusFrankATcernch:
Let's see. If there are more problems popping up, I think we should reconsider reverting the change in the object extensions. As said: I thought the change is fully backwards compatible, which obviously is not the case.
Otherwise I would not have done it.

DDRec/src/DetectorSurfaces.cpp
tmadlener: Just for completeness: https://github.com/iLCSoft/DDKalTest/pull/16 fixes things for ILD reco

PR #1231, opened 2024-02-15
CMake: use the LCIO::lcio target instead of LCIO::LCIO

andresailer:
LCIO now exports LCIO::lcio, so we could use that if LCIO is new enough

BEGINRELEASENOTES
- CMake: use the LCIO::lcio target instead of LCIO::LCIO

ENDRELEASENOTES


smuzaffar:
ping .. just checking if github action does any thing to this already merged PR

PR #1240, opened 2024-03-11
Add multithreading support to ddsim

wdconinc:
This PR adds multithreading support to ddsim, through the `-j` flag.

Backwards-incompatible effects (as the corner cases that they are):
- Because some of the actions have to be place into `__setupWorker`, this means they may not be available in the same context as global actions. This means that during user physics list initialization one cannot query particle handlers which have not been created yet.

BEGINRELEASENOTES
- Multithreading support in ddsim

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDG4.py
+++ b/DDG4/python/DDG4.py
@@ -839,6 +839,7 @@ class Geant4:
     """
     # Only use shared=True in MT mode to avoid double-save in ST mode
     shared = self.master().NumberOfThreads > 1
+    
     evt_root = EventAction(self.kernel(), 'Geant4Output2ROOT/' + name, shared)
     evt_root.HandleMCTruth = mc_truth
     evt_root.Control = True
--- a/DDG4/python/DDG4.py
+++ b/DDG4/python/DDG4.py
@@ -837,7 +837,9 @@ class Geant4:
 
     \author  M.Frank
     """
-    evt_root = EventAction(self.kernel(), 'Geant4Output2ROOT/' + name, True)
+    # Only use shared=True in MT mode to avoid double-save in ST mode
+    shared = self.master().NumberOfThreads > 1
+    evt_root = EventAction(self.kernel(), 'Geant4Output2ROOT/' + name, shared)
     evt_root.HandleMCTruth = mc_truth
     evt_root.Control = True
     if not output.endswith('.root'):
@@ -853,7 +855,9 @@ class Geant4:
 
     \author  M.Frank
     """
-    evt_lcio = EventAction(self.kernel(), 'Geant4Output2LCIO/' + name, True)
+    # Only use shared=True in MT mode to avoid double-save in ST mode
+    shared = self.master().NumberOfThreads > 1
+    evt_lcio = EventAction(self.kernel(), 'Geant4Output2LCIO/' + name, shared)
     evt_lcio.Control = True
     evt_lcio.Output = output
     evt_lcio.enableUI()
@@ -862,7 +866,9 @@ class Geant4:
 
   def setupEDM4hepOutput(self, name, output):
     """Configure EDM4hep root output for the simulated events."""
-    evt_edm4hep = EventAction(self.kernel(), 'Geant4Output2EDM4hep/' + name, True)
+    # Only use shared=True in MT mode to avoid double-save in ST mode
+    shared = self.master().NumberOfThreads > 1
+    evt_edm4hep = EventAction(self.kernel(), 'Geant4Output2EDM4hep/' + name, shared)
     evt_edm4hep.Control = True
     evt_edm4hep.Output = output
     evt_edm4hep.enableUI()
--- a/DDG4/python/DDG4.py
+++ b/DDG4/python/DDG4.py
@@ -839,7 +839,7 @@ class Geant4:
     """
     # Only use shared=True in MT mode to avoid double-save in ST mode
     shared = self.master().NumberOfThreads > 1
-    
+
     evt_root = EventAction(self.kernel(), 'Geant4Output2ROOT/' + name, shared)
     evt_root.HandleMCTruth = mc_truth
     evt_root.Control = True
--- a/DDG4/python/DDG4.py
+++ b/DDG4/python/DDG4.py
@@ -921,7 +921,7 @@ class Geant4:
     \author  M.Frank
     """
     from ROOT import PyDDG4
-    PyDDG4.run(self.master().get())
+    PyDDG4.runAll(self.master().get())
     return self
 
 
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -436,6 +436,8 @@ class DD4hepSimulation(object):
     if actionList:
       generationInit = self._buildInputStage(geant4, actionList, output_level=self.output.inputStage,
                                              have_mctruth=self._enablePrimaryHandler())
+    # Store on self so run() can read numberOfEvents before terminate() destroys it
+    self._generationInit = generationInit
 
     # ================================================================================================
 
@@ -635,12 +637,10 @@ class DD4hepSimulation(object):
     if not PyDDG4.run(master):
       logger.error("Simulation failed!")
       exitCode += 1
-    if not PyDDG4.terminate(master):
-      exitCode += 1
-      logger.error("Termination failed!")
 
     totalTimeUser, totalTimeSys, _cuTime, _csTime, _elapsedTime = os.times()
     processedEvents = self.numberOfEvents
+    generationInit = getattr(self, '_generationInit', None)
     if generationInit:
       processedEvents = int(generationInit.numberOfEvents)
       if self.numberOfEvents < 0:
@@ -655,6 +655,11 @@ class DD4hepSimulation(object):
         perEventTime = eventTime / processedEvents
         logger.info("StartUp Time: %3.2f s, Processing and Init: %3.2f s (~%3.2f s/Event @ %d threads) "
                     % (startUpTime, eventTime, perEventTime, self.numberOfThreads))
+
+    if not PyDDG4.terminate(master):
+      exitCode += 1
+      logger.error("Termination failed!")
+
     return exitCode
 
   def __setMagneticFieldOptions(self, geant4):
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -616,6 +616,12 @@ class DD4hepSimulation(object):
     from ROOT import PyDDG4
     master = geant4.master().get()
 
+    # Preload reader libraries on the master thread before workers start.
+    if self.inputFiles and self.numberOfThreads > 1:
+      from ROOT import gSystem
+      if any(f.endswith(tuple(EDM4HEP_INPUT_EXTENSIONS)) for f in self.inputFiles):
+        gSystem.Load("libDDG4EDM4HEPReader")
+
     PyDDG4.configure(master)
     PyDDG4.initialize(master)
 
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -563,9 +563,9 @@ class DD4hepSimulation(object):
 
     logger.info("# Configure G4 sensitive detectors: python setup callback")
     seq, act = geant4.addDetectorConstruction(
-      "Geant4PythonDetectorConstruction/SetupSD",
-      sensitives=self.__setupSensitives,
-      sensitives_args=(geant4, detectorDescription,))
+        "Geant4PythonDetectorConstruction/SetupSD",
+        sensitives=self.__setupSensitives,
+        sensitives_args=(geant4, detectorDescription,))
     logger.info("# Configure G4 sensitive detectors: atach'em to the sensitive volumes")
     seq, act = geant4.addDetectorConstruction("Geant4DetectorSensitivesConstruction/ConstructSD")
 
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -610,6 +610,7 @@ class DD4hepSimulation(object):
 
     startUpTime, _sysTime, _cuTime, _csTime, _elapsedTime = os.times()
 
+    exitCode = 0
     if not PyDDG4.run(master):
       logger.error("Simulation failed!")
       exitCode += 1
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -326,6 +326,7 @@ class DD4hepSimulation(object):
     kernel.registerGlobalAction(run)
     kernel.runAction().add(run)
 
+    # ----------------------------------------------------------------------------------
     # Configure run, event, track, step, and stack actions, if present
     for action_list, DDG4_Action, kernel_Action in \
         [(self.action.run, DDG4.RunAction, kernel.runAction),
@@ -426,6 +427,8 @@ class DD4hepSimulation(object):
       generationInit = self._buildInputStage(geant4, actionList, output_level=self.output.inputStage,
                                              have_mctruth=self._enablePrimaryHandler())
 
+    # ================================================================================================
+
     # And handle the simulation particles.
     part = DDG4.GeneratorAction(kernel, "Geant4ParticleHandler/ParticleHandler")
     kernel.generatorAction().adopt(part)
@@ -446,6 +449,8 @@ class DD4hepSimulation(object):
 
     return 1
 
+    # =================================================================================
+
   def __setupSensitives(self, geant4, detectorDescription):
     kernel = geant4.kernel()
 
@@ -456,7 +461,9 @@ class DD4hepSimulation(object):
       logger.error("%s", e)
       return 1
 
+    # =================================================================================
     # get lists of trackers and calorimeters in detectorDescription
+
     trk, cal, unk = self.getDetectorLists(detectorDescription)
     for detectors, function, defFilter, defAction, abort in \
         [(trk, geant4.setupTracker, self.filter.tracker, self.action.tracker, False),
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -495,10 +495,12 @@ class DD4hepSimulation(object):
     kernel = geant4.kernel()
     logger.debug("Setting up actions")
     self.__setupActions(kernel)
+    import DDG4
+    logger.debug("Setting up Geant4Random for worker thread")
+    # Initialize a per-worker Geant4Random wrapping this thread's CLHEP engine
+    self.random.initializeWorker(DDG4, kernel, self.printLevel)
     logger.debug("Setting up EventSeeder for worker")
     # Setup EventSeeder for this worker
-    # Uses the shared Geant4Random instance from master
-    import DDG4
     self.random.setupEventSeeder(DDG4, kernel)
     logger.debug("Setting up generator actions")
     self.__setupGeneratorActions(kernel, geant4)
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -72,6 +72,7 @@ class DD4hepSimulation(object):
 
     self.numberOfEvents = 0
     self.skipNEvents = 0
+    self.numberOfThreads = 1
     self.physicsList = None  # deprecated use physics.list
     self.crossingAngleBoost = 0.0
     self.macroFile = ''
@@ -183,6 +184,9 @@ class DD4hepSimulation(object):
     parser.add_argument("--skipNEvents", action="store", dest="skipNEvents", default=self.skipNEvents, type=int,
                         help="Skip first N events when reading a file")
 
+    parser.add_argument("--numberOfThreads", "-j", action="store", dest="numberOfThreads", default=self.numberOfThreads,
+                        type=int, help="Number of threads for simulation")
+
     parser.add_argument("--physicsList", action="store", dest="physicsList", default=self.physicsList,
                         help="Physics list to use in simulation. Deprecated, use physics.list")
 
@@ -254,6 +258,7 @@ class DD4hepSimulation(object):
 
     self.numberOfEvents = parsed.numberOfEvents
     self.skipNEvents = parsed.skipNEvents
+    self.numberOfThreads = parsed.numberOfThreads
     self.physicsList = parsed.physicsList
     self.crossingAngleBoost = parsed.crossingAngleBoost
     self.macroFile = parsed.macroFile
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -79,6 +79,8 @@ class DD4hepSimulation(object):
     self.enableGun = False
     self.enableG4GPS = False
     self.enableG4Gun = False
+    self._g4gun = []
+    self._g4gps = []
     self.vertexSigma = [0.0, 0.0, 0.0, 0.0]
     self.vertexOffset = [0.0, 0.0, 0.0, 0.0]
     self.enableDetailedShowerMode = False
@@ -363,6 +365,7 @@ class DD4hepSimulation(object):
       g4gun.Mask = 2
       logger.info("++++ Adding Geant4 Particle Gun ++++")
       actionList.append(g4gun)
+      self._g4gun.append(g4gun)
 
     if self.enableG4GPS:
       # GPS Create something
@@ -371,6 +374,7 @@ class DD4hepSimulation(object):
       g4gps.Mask = 3
       logger.info("++++ Adding Geant4 General Particle Source ++++")
       actionList.append(g4gps)
+      self._g4gps.append(g4gps)
 
     start = 4
     for index, plugin in enumerate(self.inputConfig.userInputPlugin, start=start):
@@ -594,6 +598,15 @@ class DD4hepSimulation(object):
 
     PyDDG4.configure(master)
     PyDDG4.initialize(master)
+
+    # GPS
+    # FIXME this applies to only 1 thread, and Geant4GeneratorWrapper doesn't work as shared=True
+    if self._g4gun is not []:
+      for g4gun in self._g4gun:
+        g4gun.generator()
+    if self._g4gps is not []:
+      for g4gps in self._g4gps:
+        g4gps.generator()
 
     startUpTime, _sysTime, _cuTime, _csTime, _elapsedTime = os.times()
 
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -401,11 +401,11 @@ class DD4hepSimulation(object):
         gen.Input = "Geant4EventReaderHepEvtLong|" + inputFile
       elif inputFile.endswith(tuple([".hepmc"] + HEPMC3_SUPPORTED_EXTENSIONS)):
         if self.hepmc3.useHepMC3:
-          gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index)
+          gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index, shared=True)
           gen.Parameters = self.hepmc3.getParameters()
           gen.Input = "HEPMC3FileReader|" + inputFile
         else:
-          gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index)
+          gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/hepmc%d" % index, shared=True)
           gen.Input = "Geant4EventReaderHepMC|" + inputFile
       elif inputFile.endswith(".pairs"):
         gen = DDG4.GeneratorAction(kernel, "Geant4InputAction/GuineaPig%d" % index)
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -491,7 +491,6 @@ class DD4hepSimulation(object):
 
   def __setupMaster(self, geant4):
     logger.debug("Setting up master")
-    kernel = geant4.master()
     return 1
 
   def run(self):
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -347,6 +347,8 @@ class DD4hepSimulation(object):
   def __setupGeneratorActions(self, kernel, geant4):
     import DDG4
 
+    shared = (kernel.RunManagerType == "G4MTRunManager")
+
     actionList = []
 
     if self.enableGun:
@@ -360,7 +362,7 @@ class DD4hepSimulation(object):
 
     if self.enableG4Gun:
       # GPS Create something
-      g4gun = DDG4.GeneratorAction(kernel, "Geant4GeneratorWrapper/Gun")
+      g4gun = DDG4.GeneratorAction(kernel, "Geant4GeneratorWrapper/Gun", shared=shared)
       g4gun.Uses = 'G4ParticleGun'
       g4gun.Mask = 2
       logger.info("++++ Adding Geant4 Particle Gun ++++")
@@ -369,7 +371,7 @@ class DD4hepSimulation(object):
 
     if self.enableG4GPS:
       # GPS Create something
-      g4gps = DDG4.GeneratorAction(kernel, "Geant4GeneratorWrapper/GPS")
+      g4gps = DDG4.GeneratorAction(kernel, "Geant4GeneratorWrapper/GPS", shared=shared)
       g4gps.Uses = 'G4GeneralParticleSource'
       g4gps.Mask = 3
       logger.info("++++ Adding Geant4 General Particle Source ++++")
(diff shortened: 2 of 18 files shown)


DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -447,8 +393,6 @@ def run(self):
       self._buildInputStage(geant4, actionList, output_level=self.output.inputStage,
                             have_mctruth=self._enablePrimaryHandler())
 
-    # ================================================================================================
```
andresailer, line 450: Can you limit to necessary changes only, no gratuitous beautyfications please? 
I am not sure if things are re-arrange a lot or `diff` gets confused.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -467,31 +411,134 @@ def run(self):
...
-        self.__setupSensitiveDetectors(detectors, function, defFilter, defAction, abort)
+        sim.__setupSensitiveDetectors(detectors, function, defFilter, defAction, abort)
       except Exception as e:
         logger.error("Failed setting up sensitive detector %s", e)
         raise
 
-  # =================================================================================
+    return 1
+
+  def __setupWorker(geant4, sim):
```
andresailer, line 441: As is this is still an instance method.
```suggestion
  def __setupWorker(self, geant4)
```
and accordingly below? Or why would this not work?
Then you either need to move out of the class or add the `@staticmethod` decorator.
And is the return value required by the interface?

DDG4/python/DDSim/DD4hepSimulation.py
wdconinc: I was trying to :-)  The main reason for rearranging is that the worker actions needed to be separated from the master actions, as previously they were intermingled. That required some necessary rearranging.

DDG4/python/DDSim/DD4hepSimulation.py
wdconinc: I can (and should) likely make this work with `@staticmethod`. The return value appears to be unused. I based the structure on what is done in `SiDSim_MT.py`, to the extent it provides any prior art.

wdconinc: (I was unable to make this work with an instance method since I suspect that by the time `__setupWorker` is called, the DD4hepSimulation object is not available to the thread that calls `__setupWorker`.)

DDG4/python/DDSim/DD4hepSimulation.py
wdconinc: One way to help diff out could be to define helper functions in the order where things are in the file right now, and then use those helper functions in some new code that simply calls them (like `__setupWorker`). E.g. the UI setup can be put in a function at its current place in the file.

wdconinc:
This is still a work in progress, but let me give some more background here as to the plan (the "more later" from the top).
- The intention is, of course, full backwards compatibility so this runs in the single-threaded `G4RunManager` with command line options as given now (and therefore must pass all tests unchanged).
- This PR already works in multi-threaded mode for generators like the gun, but
  - Not for external (hepmc) input files: I would like to consider this beyond the scope of this PR since this will require some changes to the input file reader to make sure it does not read the same sequence of events in each file.
  - Adapting to MT has caused single-threaded running to break: As written above, this needs to get fixed, of course.
- There may be changes to the output files that are due to event reordering. These are likely unavoidable.

It is likely that I will be doing some rewrites of the last commit (5bf711cfc73b0b43151cab3e0bc4bfafb4418807) which is a bit too big for my taste and not as clear as one would like for non-squash merging.

DDG4/python/DDSim/DD4hepSimulation.py
andresailer: But according to the indentation, this still _is_ an instance method? Just because the first argument isn't `self` doesn't change that, does it?

wdconinc: Made this into an instance method in 1b409705f501fc3c106619b4e988956d2b29b845.

wdconinc:
The status of this PR after #1417 is that we can now run MT gun simulations, like so:
```
ddsim --compactFile DDDetectors/compact/SiD.xml -j8 -G -N160 --gun.distribution uniform --outputFile SiD.edm4hep.root
```
Main missing feature is ensuring `inputFiles` is serialized. Right now this just reads the same event sequence for all threads.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -71,14 +71,13 @@ def __init__(self):
...
     self.skipNEvents = 0
+    self.numberOfThreads = 1
     self.physicsList = None  # deprecated use physics.list
     self.crossingAngleBoost = 0.0
     self.macroFile = ''
     self.enableGun = False
     self.enableG4GPS = False
     self.enableG4Gun = False
-    self._g4gun = None
-    self._g4gps = None
```
wdconinc, line 81: TODO: Geant4 gun and GPS event generators are not initialized correctly due to this removal.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -558,13 +591,16 @@ def run(self):
...
+                    % (startUpTime, eventTime, perEventTime, self.numberOfThreads))
     return exitCode
 
   def __setMagneticFieldOptions(self, geant4):
     """ create and configure the magnetic tracking setup """
-    field = geant4.addConfig('Geant4FieldTrackingSetupAction/MagFieldTrackingSetup')
+    if self.numberOfThreads > 1:
+      seq, field = geant4.addDetectorConstruction("Geant4FieldTrackingConstruction/MagFieldTrackingSetup")
+    else:
+      field = geant4.addConfig('Geant4FieldTrackingSetupAction/MagFieldTrackingSetup')
```
wdconinc, line 603: TODO: This change is likely unnecessary.

DDG4/python/DDSim/DD4hepSimulation.py
wdconinc: Despite modified implementation, this remain a TODO since only one thread reads the `macroFile`... (as of da36a3e368bbe0651fdb921a67512cb287fa438e). And Geant4GeneratorWrapper doesn't allow setting properties when shared = True.

MarkusFrankATcernch:
Hi Wouter,

This construct is not simple. Let's see what is necessary before the hacking starts...

For `Geant4` the `GeneratorAction` belongs to the worker thread's execution flow, not the main thread.
This idea is also reflected in DDG4. If you look at [DD4hep/blob/master/DDG4/examples/SiDSim_MT.py](https://github.com/AIDASoft/DD4hep/blob/master/DDG4/examples/SiDSim_MT.py) you can clearly see this.

Now what one probably wants is somehow a structure that this worker thread bound `GeneratorAction` is sharable between all workers (share same input file, share common parametrization etc) and use this action from all threads. Secondly the underlying reader implementation should be `accessible` by the configuration (python with ddg4, ddsim, etc) to allow more fine grained customization than just setting some input file and defining a factory.

From what I see somehow the input stage needs then to be re-designed. In particular exchanging data between the actions building the queue of `GeneratorAction`s which define the input stage, cannot use the `Geant4Event` and its extension mechanism to pass these data. The `Geant4Event` is bound to the worker thread via the `Geant4Context`.

Having said all this, it looks necessary that the queue of `GeneratorAction`s needs to be cut in two:
- the first part being locked and single threaded doing the file reading (multiple inputs, spillover, multiple collisions per crossing, etc) and the data preparation with user exposure of the queue elements before execution starts.
- the second part is not locked and can then finalize the event generation (smearing, primary vertex movements, etc.) and passes all the initial elements to the `Geant4Event`, which is then passed to `Geant4` for the actual simulation work.

How to do something like this in a backwards compatible way is not so obvious to me and should be designed properly.
As I use to say: software is never difficult or complicated unless there is a lack of thinking upfront.
Maybe we should first collect all the goodies we want to see in such an implementation. The [parameter setup you mention here](https://github.com/AIDASoft/DD4hep/pull/1240#issuecomment-2749160918) is already a first step.

MarkusFrankATcernch:
So after coming back from the drawing board....
What you effectively want is a transition from the following object model as it is implemented now:

Geant4-Input


to an input stage which looks much rather like this (details are left out here):

Geant4-MT

Here you have 
- a single threaded simple input stage filling the `Geant4Event` structure. The `Geant4Event`s must be requested by the initializing action (`Geant4GenerationInit`) from the `Geant4EventQueue` object to limit the number of queued events (memory, etc). This compromises the first actions from the above model.
- these Geant4Event entities are then pushed into a queue from which these `Geant4Event`s are then requested by the multi-threaded simulation.
- the multi-thread action sequence then finalizes the primaruy generation and the rest continues as shown here: [DD4hep/blob/master/DDG4/examples/SiDSim_MT.py](https://github.com/AIDASoft/DD4hep/blob/master/DDG4/examples/SiDSim_MT.py)

In principle this is not too difficult, but still quite more difficult that you have imagined upfront.

In addition:
- the `Geant4InputAction` objects in the queue need their properties be exposed to the user
- the entire machinery must be embedded in ddsim. How this can be done elegantly has to be seen.

wdconinc:
While I agree that the diagram you outline (with an event queue) is optimal for throughput performance, I wonder if there is another less optimal but still functional approach (which may have some threads waiting unnecessarily for some of the time). What I had originally anticipated was a pattern where there are multiple Geant4InputActions sequences, one per thread as now, but the action that reads from disk would have a shared reader that is locked. This would mimic the approach of the Geant4OutputAction where each thread has an output event action, but they share and lock the output file.

MarkusFrankATcernch:
@wdconinc 
We should under no circumstances deviate too far from the processing model Geant4 uses.
Geant4 expects the `G4GeneratorAction` to be part of the worker thread and each worker thread is only provided with a unique seed for the random generator for this event. That's it.

Where one actually puts the Geant4EventQueue is sort of user-gusto here, but it must be located before the
G4Primary particles and vertices are created. If you want to work with input files a la HepMC, you have to go the complicated way I am afraid. Otherwise you would need one input per thread.

If we invest here, we should do it properly....even if it costs some development.

wdconinc:
> In addition:
> 
> * the `Geant4InputAction` objects in the queue need their properties be exposed to the user

See #1440, which exposes the properties correctly now.

murnanedaniel:
@wdconinc @MarkusFrankATcernch Hi Wouter and Markus, what is the status of including multithreading into ddsim with HEPMC3 file inputs? I am working on a large simulation, and having MT could potentially give a nice boost in throughput. If there are concrete places that I can contribute, please let me know. If the exposure of properties is handled, is this "simply" a matter of having ddsim handle file reads and writes correctly? Or are there more fundamental components to update?

Again, keen to help where it makes sense!

wdconinc:
@murnanedaniel Depending on your use case, this already works fine (multi-threaded).
- It works with DD4hep particle gun, input files (hepmc3, haven't tried evtgen and edm4hep inputs but should also work).
- It doesn't work with GPS and Geant4 particle gun since the setup macro is only passed to one thread.

I have been using this for a month or two (hepmc3 input files) typically 8 threads and it scales without loss of efficiency at that (small) thread count.

wdconinc:
> @wdconinc @murnanedaniel 
> So in other words, only the raw DDG4 python interface can currently be used to run multi-threaded.
> Correct Wouter ?

The scope of this PR has always been to make multi-threaded running work with ddsim. I don't know what other ways of running DDG4 people are using, but I'm not familiar with them.

murnanedaniel:
Okay, I made a couple of quick tweaks and it **does** seem to work, which is really promising. Firstly, the tweaks I made are here: https://github.com/murnanedaniel/DD4hep/commit/cbbe86e671293f15a705f5b296a652654d94f4fd. I'm not sure if you agree with them, or why they are not necessary in your setup?

Secondly, there is one final error appearing after the simulation that doesn't appear in the main DD4hep. You can see it at the end of the successful log file: [full_output_mt.log](https://github.com/user-attachments/files/20001134/full_output_mt.log)
Any idea where this is coming from?

murnanedaniel:
@wdconinc Just following up here, since we will soon be running a large set of simulations with your MT changes. I just want to ensure that the small error that appears, and the tweaks that I needed to make, are still consistent with the working version of ddsim as you see it. One that same point - have you validated that MT sim of hepmc3 inputs produces more or less identical outputs to DD4hep main?

BrieucF:
Is there any plan to have this integrated?

DDG4/python/DDG4.py
```diff
@@ -167,8 +167,6 @@ def _getKernelProperty(self, name):
     return _evalProperty(ret.data)
   elif hasattr(self.get(), name):
     return _evalProperty(getattr(self.get(), name))
-  elif hasattr(self, name):
```
andresailer: Is dropping this intentional?

DDG4/src/python/PyDDG4.cpp
```diff
@@ -33,6 +33,38 @@ int PyDDG4::run(Kernel& kernel)  {
   return ret;
 }
 
+int PyDDG4::configure(Kernel& kernel)  {
+  int ret;
+  dd4hep::DDPython::AllowThreads allow(0);
+  if ( 1 != (ret=kernel.configure()) )
+    return ret;
+  return ret;
```
andresailer, line 41:
```suggestion
  return kernel.configure();
```
This seems to be the same?

DDG4/src/python/PyDDG4.cpp
```diff
@@ -33,6 +33,38 @@ int PyDDG4::run(Kernel& kernel)  {
...
+}
+
+int PyDDG4::initialize(Kernel& kernel)  {
+  int ret;
+  dd4hep::DDPython::AllowThreads allow(0);
+  if ( 1 != (ret=kernel.initialize()) )
+    return ret;
+  return ret;
```
andresailer, line 49:
```suggestion
  return kernel.initialize();
```

DDG4/src/python/PyDDG4.cpp
```diff
@@ -33,6 +33,38 @@ int PyDDG4::run(Kernel& kernel)  {
...
+}
+
+int PyDDG4::run(Kernel& kernel)  {
+  int ret;
+  dd4hep::DDPython::AllowThreads allow(0);
+  if ( 1 != (ret=kernel.run()) )
+    return ret;
+  return ret;
```
andresailer, line 57:
```suggestion
  return kernel.run();
```

DDG4/src/python/PyDDG4.cpp
```diff
@@ -33,6 +33,38 @@ int PyDDG4::run(Kernel& kernel)  {
...
+}
+
+int PyDDG4::terminate(Kernel& kernel)  {
+  int ret;
+  dd4hep::DDPython::AllowThreads allow(0);
+  if ( 1 != (ret=kernel.terminate()) )
+    return ret;
+  return ret;
```
andresailer, line 65:
```suggestion
  return kernel.terminate();
```

jmcarcell:
With the permission of Wouter I have changed a few things to fix tests. All of them are passing now (MT generation and MT comparison, for example, between 2 and 4 threads) except comparing exactly the output of ST and MT.

First, there are some tests that don't get to run because batch mode is not passed and this triggers an error. Fixed in `DD4hepSimulation.py`, by allowing not passing `--runType=batch` with also G4Gun and G4GPS. I think this is a general fix, it could be taken out of this PR.

Then there a few crashes on CI, and the stack trace is pointing to
`TROOT::CloseFiles()`. This one seems to happen because lifetimes are preserved from python and in MT the file is never closed. `endRun` now is defined in `Geant4OutputROOT` to close the file when all the threads are done or immediately in ST. I'm not sure if it's the best fix. There is a mutex that maybe is not needed in `closeOutput()`.

Then, the comparison script needed a few setup changes, to load `libDDG4IO` and fix a few comparisons that were giving cppyy-related errors. 

The last change is to make `Geant4Random` thread-local, otherwise each thread will smash other thread's instance in `setMainInstance`. It follows that each thread has to use its own `m_rootRandom` instead of a global `gRandom`. And also to always fill G4EventID in TTrees so that it can be used in the comparison.

Next:
- Investigate whether it is possible to obtain the same results ST vs MT
- Fix the comparison when using HepMC input

wdconinc:
@jmcarcell I'm confused about e062d12. What target should produce the library `libDDG4IO`? I've reverted it.

With the latest commits (incl. some test reorganization), I now get:
```
100% tests passed, 0 tests failed out of 41
```

jmcarcell:
I'll check tomorrow but I have to push a few changes and then I get all test passing, I don't get how you can get them to pass without changes :thinking: CI is not happy.

wdconinc:
> I'll check tomorrow but I have to push a few changes and then I get all test passing, I don't get how you can get them to pass without changes 🤔 CI is not happy.

Yeah, I don't typically run locally against LCG (I probably only ever use LCG to debug DD4hep CI issues...), so I don't know what LCG is doing that breaks this.

wdconinc:
Ah, https://github.com/AIDASoft/DD4hep/actions/runs/27106031006/job/79995227785?pr=1240#step:4:1239:
```
  -- EDM4hepFileReader requires at least PODIO 1.0, disabled
```
So, the reader is not built (but writer is), so it's no wonder the tests fail to load.

wdconinc:
The only remaining issue here is now the G4EventID that isn't a 'proper' branch but a direct leaf, and it fails in DDDigi. @jmcarcell Do you want to take a look?

wdconinc:
>    18 files     18 suites   12h 14m 8s ⏱️
>   380 tests   380 ✅ 0 💤 0 ❌
> 3 310 runs  3 310 ✅ 0 💤 0 ❌

Passes all checks. Zarro boogs!

wdconinc:
Now, let's see what copilot thinks about all this...

DDTest/CMakeLists.txt
```diff
@@ -180,6 +180,272 @@ if (DD4HEP_USE_GEANT4)
...
+      ddsim --compactFile=${CMAKE_INSTALL_PREFIX}/DDDetectors/compact/SiD.xml
+            --runType=batch -N=10 -j 1 -G
+            --outputFile=gun_mt1.root
+            --random.seed=333333
+            --random.enableEventSeed
+            --gun.position \"0.0 0.0 1.0*cm\"
+            --gun.direction \"1.0 0.0 1.0\"
+            --gun.particle=mu-
+            --gun.energy=10*GeV
+            --part.userParticleHandler=)
```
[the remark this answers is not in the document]
wdconinc: This is in line with pre-existing uses, e.g. https://github.com/wdconinc/DD4hep/blob/89054db50e3e255e7adf1d8e6f8a74f854663dda/DDTest/CMakeLists.txt#L104

DDG4/src/Geant4Output2ROOT.cpp
```diff
@@ -81,11 +105,32 @@ TTree* Geant4Output2ROOT::section(const std::string& nam) {
...
+        if ( i != m_sections.end() )
+          m_sections.erase(i);
+        m_branches.clear();
+        for ( auto& [name, tree] : m_sections )  tree->Write();
+        m_sections.clear();
+        m_tree->Write();
+        m_file->Close();
+        m_tree = nullptr;
+        detail::deletePtr (m_file);
+      }
```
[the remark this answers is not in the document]
wdconinc: Extracted into `closeOutputUnlocked()`.

DDTest/python/compare_simulation_output.py
```diff
@@ -0,0 +1,367 @@
...
+    try:
+        import ROOT
+    except ImportError:
+        print("ERROR: ROOT/PyROOT not available")
+        return False
+
+    # Load the DDG4 dictionary so ROOT can read DD4hep-native output files.
+    # This is a no-op when only EDM4hep output is used.
+    ROOT.gSystem.Load("libDDG4")
+    ROOT.gSystem.Load("libDDG4Plugins")
```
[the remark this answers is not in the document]
wdconinc: Guarded and warning printed.

DDTest/python/compare_simulation_output.py
```diff
@@ -0,0 +1,367 @@
...
+                x1, x2 = _coord(v1, x_attr), _coord(v2, x_attr)
+                y1, y2 = _coord(v1, y_attr), _coord(v2, y_attr)
+                z1, z2 = _coord(v1, z_attr), _coord(v2, z_attr)
+                if abs(x1 - x2) > tolerance or abs(y1 - y2) > tolerance or abs(z1 - z2) > tolerance:
+                    print(f"  {description} mismatch: ({x1}, {y1}, {z1}) vs ({x2}, {y2}, {z2})")
+                    return False
+                return True
+            except Exception:
+                continue
+    return True
```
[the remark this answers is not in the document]
wdconinc: Print warning.

DDG4/src/Geant4Output2ROOT.cpp
```diff
@@ -51,15 +56,34 @@ Geant4Output2ROOT::~Geant4Output2ROOT() {
...
+  const int nThreads = context()->kernel().master().numThreads();
+  // ST: nThreads==0, close immediately
+  // MT: wait until every worker has called endRun before closing
+  int done = ++m_endRunCount;
+  if (nThreads == 0 || done >= nThreads) {
+    closeOutput();
+    m_endRunCount = 0;
+  }
+  Geant4OutputAction::endRun(run);
+}
```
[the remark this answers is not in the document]
wdconinc: Move to `beginRun`.

DDG4/src/python/PyDDG4.cpp
```diff
@@ -33,6 +33,38 @@ int PyDDG4::run(Kernel& kernel)  {
...
+}
+
 int PyDDG4::execute()  {
   Kernel& k = Kernel::instance(dd4hep::Detector::getInstance());
   return run(k);
```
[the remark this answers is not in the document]
wdconinc: This is intentional since we need to be able to call the phases in isolation, not only all together (with the new `runAll` which is the previous behavior.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -538,27 +613,36 @@ def run(self):
...
+    # In multi-threaded mode with shared=False, the actual generators are created
+    # per-worker and initialized automatically by Geant4, so we skip this step
+    if self.numberOfThreads == 1:
+      if self._g4gun is not []:
+        for g4gun in self._g4gun:
+          g4gun.generator()  # Initialize G4ParticleGun and register UI commands
+          logger.debug("Initialized G4Gun generator")
+      if self._g4gps is not []:
+        for g4gps in self._g4gps:
+          g4gps.generator()  # Initialize GPS and register UI commands
```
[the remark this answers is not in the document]
wdconinc: Replaced with better truthiness check.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -481,6 +436,8 @@ def run(self):
     if actionList:
       generationInit = self._buildInputStage(geant4, actionList, output_level=self.output.inputStage,
                                              have_mctruth=self._enablePrimaryHandler())
+    # Store on self so run() can read numberOfEvents before terminate() destroys it
+    self._generationInit = generationInit
```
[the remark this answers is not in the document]
wdconinc: Not sure what this one's on about...

wdconinc:
Now that the tests run multi-threaded, the issues like https://github.com/AIDASoft/DD4hep/actions/runs/27167948215/job/80199881284?pr=1240#step:4:6627 will pop up for older LCG releases due to Geant4's ongoing efforts at removing thread unsafe code (e.g. multiple threads writing simultaneously to a single static cache as in e.g. G4EmSaturation, or the ones listed in https://github.com/wdconinc/geant4/issues/43 which still apply to 11.4.1).

DDG4/python/DDSim/DD4hepSimulation.py
wdconinc: This was completed.

DDG4/python/DDSim/DD4hepSimulation.py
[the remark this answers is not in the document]
wdconinc: Resolving this as invalid. `generationInit` is set to `None` literally one line about this block, and that's been there for 2 years since the very first commit in this branch...

andresailer:
I ran some qq_ln events in SiD.
```
qq_01.txt:DDSim            INFO StartUp Time: 3.10 s, Processing and Init: 775.35 s (~24.23 s/Event @ 1 threads) 
qq_02.txt:DDSim            INFO StartUp Time: 4.26 s, Processing and Init: 824.39 s (~25.76 s/Event @ 2 threads) 
qq_04.txt:DDSim            INFO StartUp Time: 4.78 s, Processing and Init: 827.15 s (~25.85 s/Event @ 4 threads) 
qq_08.txt:DDSim            INFO StartUp Time: 4.40 s, Processing and Init: 834.77 s (~26.09 s/Event @ 8 threads) 
qq_12.txt:DDSim            INFO StartUp Time: 4.59 s, Processing and Init: 1130.63 s (~35.33 s/Event @ 12 threads) 
qq_16.txt:DDSim            INFO StartUp Time: 4.69 s, Processing and Init: 1366.86 s (~42.71 s/Event @ 16 threads)
```
Hyperthreading is enabled (8 -> 16), so the scaling is fine'ish until we run into hyperthreaded cores.

I also did some profiling with AMD uProf. I still need to understand how this and threading works in general. Tentatively, I would say if events are too short (particle gun muons), there is some lock contention from writing the event out, which does not happen when events are taking longer like the qq events I am using here.

I also found some issue with the "InstanceCount" helper just locking all over the place that we should probably disable for serious workloads where this is not needed.

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -526,7 +488,120 @@ def run(self):
...
+    logger.debug("Setting up EventSeeder for worker")
+    # Setup EventSeeder for this worker
+    self.random.setupEventSeeder(DDG4, kernel)
+    logger.debug("Setting up generator actions")
+    self.__setupGeneratorActions(kernel, geant4)
+    logger.debug("Setting up output")
+    self.outputConfig.initialize(dd4hepsimulation=self, geant4=geant4)
+    return 1
+
+  def __setupMaster(self, geant4):
```
andresailer, line 511: This seems like a dummy function? Probably deserves a comment

DDG4/src/python/PyDDG4.cpp
MarkusFrankATcernch: Is functionally the same, but if you need debugging, how do you set the break point?
Here the detail kicks in.....
As long as all works it is fine, but when there is trouble you first have to instrument the code again.

andresailer:
Hi @jmcarcell Could you rebase to fix the conflicts?

DDG4/python/DDG4.py
jmcarcell: This is https://github.com/AIDASoft/DD4hep/pull/1622/changes

DDTest/CMakeLists.txt
```diff
@@ -180,6 +180,290 @@ if (DD4HEP_USE_GEANT4)
...
+  set_tests_properties(t_ddsim_gun_mt1 PROPERTIES
+    FAIL_REGULAR_EXPRESSION "Exception;EXCEPTION;ERROR;Error"
+    PASS_REGULAR_EXPRESSION "@ 1 threads"
+    DISBALE ${disable_mt_tests}
```
andresailer:
```suggestion
    DISABLE ${disable_mt_tests}
```

DDTest/CMakeLists.txt
```diff
@@ -180,6 +180,290 @@ if (DD4HEP_USE_GEANT4)
...
+            --gun.position \"0.0 0.0 1.0*cm\"
+            --gun.direction \"1.0 0.0 1.0\"
+            --gun.particle=mu-
+            --gun.energy=10*GeV
+            --part.userParticleHandler=)
+  set_tests_properties(t_ddsim_gun_mt2 PROPERTIES
+    PROCESSORS 2
+    FAIL_REGULAR_EXPRESSION "Exception;EXCEPTION;ERROR;Error"
+    PASS_REGULAR_EXPRESSION "@ 2 threads"
+    DISBALE ${disable_mt_tests}
```
andresailer:
```suggestion
    DISABLE ${disable_mt_tests}
```

DDTest/CMakeLists.txt
```diff
@@ -180,6 +180,290 @@ if (DD4HEP_USE_GEANT4)
...
+            --gun.position \"0.0 0.0 1.0*cm\"
+            --gun.direction \"1.0 0.0 1.0\"
+            --gun.particle=mu-
+            --gun.energy=10*GeV
+            --part.userParticleHandler=)
+  set_tests_properties(t_ddsim_gun_mt4 PROPERTIES
+    PROCESSORS 4
+    FAIL_REGULAR_EXPRESSION "Exception;EXCEPTION;ERROR;Error"
+    PASS_REGULAR_EXPRESSION "@ 4 threads"
+    DISBALE ${disable_mt_tests}
```
andresailer:
```suggestion
    DISABLE ${disable_mt_tests}
```

andresailer (commented):
Me spel gud.

DDG4/python/DDSim/Helper/Geometry.py
```diff
@@ -79,14 +83,8 @@ def constructGeometry(self, kernel, geant4, geoPrintLevel=2, numberOfThreads=1):
...
-    sensitives = DetectorConstruction(kernel, str('Geant4DetectorSensitivesConstruction/ConstructSD'))
-    sensitives.enableUI()
-    seq.adopt(sensitives)
-
     for index, (detName, regexDetectors) in enumerate(sorted(self._regexSDDict.items())):
       seq, act = geant4.addDetectorConstruction(f'Geant4RegexSensitivesConstruction/ConstrSDRegEx_{index}_{detName}')
       act.Detector = detName
-      # this will set Match, and other properties if possible
       for key, value in regexDetectors.items():
         setattr(act, key, value)
```
andresailer, line 92: Shouldn't this move to where also constructSD is now called?

DDG4/python/DDSim/Helper/Random.py
```diff
@@ -3,9 +3,33 @@
...
+_random_counter = threading.local()
+
+
+def _get_eventseed_id():
+  """Get a unique ID for EventSeeder in this thread"""
+  if not hasattr(_eventseed_counter, 'value'):
+    _eventseed_counter.value = 0
+  else:
+    _eventseed_counter.value += 1
+  return _eventseed_counter.value
```
andresailer: This function is not used?

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -412,20 +363,24 @@ def run(self):
...
-      self._g4gun.Uses = 'G4ParticleGun'
-      self._g4gun.Mask = 2
-      logger.info("++++ Adding Geant4 Particle Gun ++++")
-      actionList.append(self._g4gun)
+      # G4Gun: Always use shared=False to allow macro configuration
+      # In MT mode, Geant4 will handle making the gun available to workers
+      g4gun = DDG4.GeneratorAction(kernel, "Geant4GeneratorWrapper/Gun", shared=False)
+      g4gun.Uses = 'G4ParticleGun'
+      g4gun.Mask = 2
+      logger.info("++++ Adding Geant4 Particle Gun (non-shared for macro support) ++++")
```
andresailer:
```suggestion
      logger.info("++++ Adding Geant4 Particle Gun ++++")
```
Since this is always non-shared, we don't have to print it?

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -412,20 +363,24 @@ def run(self):
...
-      self._g4gps.Uses = 'G4GeneralParticleSource'
-      self._g4gps.Mask = 3
-      logger.info("++++ Adding Geant4 General Particle Source ++++")
-      actionList.append(self._g4gps)
+      # GPS: Always use shared=False to allow macro configuration
+      # In MT mode, Geant4 will handle making GPS available to workers
+      g4gps = DDG4.GeneratorAction(kernel, "Geant4GeneratorWrapper/GPS", shared=False)
+      g4gps.Uses = 'G4GeneralParticleSource'
+      g4gps.Mask = 3
+      logger.info("++++ Adding Geant4 General Particle Source (non-shared for macro support) ++++")
```
andresailer:
```suggestion
      logger.info("++++ Adding Geant4 General Particle Source ++++")
```

DDG4/python/DDSim/Helper/Random.py
```diff
@@ -50,10 +79,68 @@ def initialize(self, DDG4, kernel, output):
...
+    We use the same initial seed as the master so that EventSeeder computes
+    the same per-event seeds across all workers.
+
+    :param DDG4: DDG4 module
+    :param kernel: Worker kernel
+    :param int output: output level
+    :returns: Geant4Random instance for this worker
+    """
+    # Create a Geant4Random wrapper for this worker
+    random_id = _get_random_id()
```
andresailer, line 115: I don't get how this works. Either because I don't understand how `threading.local()` works, or how the workers are set up. But this doesn't give a random ID, it will give 0 to each worker in a different thread?

andresailer:
Hi @wdconinc 

Will you have a bit of time looking at the recent issues commented on?

wdconinc:
> Will you have a bit of time looking at the recent issues commented on?

Yes, I have local responses (and fixes) to all of the remaining ones and will get to pushing them today. Was hoping it would get done by the meeting today, but alas...

DDG4/src/python/PyDDG4.cpp
wdconinc: See https://github.com/AIDASoft/DD4hep/pull/1240/changes#r2884825476.

DDG4/src/python/PyDDG4.cpp
wdconinc: See https://github.com/AIDASoft/DD4hep/pull/1240/changes#r2884825476.

DDG4/src/python/PyDDG4.cpp
wdconinc: See https://github.com/AIDASoft/DD4hep/pull/1240/changes#r2884825476.

DDG4/python/DDSim/Helper/Geometry.py
wdconinc: Done in b61a277352d8d5c7a76aec53314a68ea1c2a29b8.

DDG4/python/DDSim/Helper/Random.py
wdconinc: Removed in 8cfff419760b3729af428d7a8d1c758bfc87a0fa.

DDG4/python/DDSim/DD4hepSimulation.py
wdconinc: Added in 4dbba46a1c84fa47b32c16af3ed592c60c54b249.

DDG4/python/DDSim/Helper/Random.py
wdconinc: This was indeed a bug and resulted in 0 for each thread. In 18e5c182ac6d144bceb969d97adddb23c5161d0e the logic was changed to use a global random counter and lock.

andresailer (approved):
Thanks!

PR #1251, opened 2024-04-03
EDM4hepOutput: determine momentum vector type automagically

### LINKED ISSUE #1250: EDM4hep output: ensure float / double is consistent with EDM4hep version used
The version check has to be adapted because EDM4hep reverted the type.

https://github.com/AIDASoft/DD4hep/blob/0207c48272d9051e5c8f8cd8017bc40f1d3db6b6/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L358-L364

andresailer:
BEGINRELEASENOTES
- EDM4hepOutput: determine momentum vector type automagically. Make it work with any EDM4hep version correctly, fixes #1250 

ENDRELEASENOTES

### DIFF
--- a/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
+++ b/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
@@ -354,14 +354,11 @@ void Geant4Output2EDM4hep::saveParticles(Geant4ParticleMap* particles)    {
       const G4ParticleDefinition* def = p.definition();
       auto mcp = m_particles.create();
       mcp.setPDG(p->pdgID);
+      // Because EDM4hep is switching between vector3f[loat] and vector3d[ouble]
+      using MT = decltype(std::declval<edm4hep::MCParticle>().getMomentum().x);
+      mcp.setMomentum( {MT(p->psx/CLHEP::GeV),MT(p->psy/CLHEP::GeV),MT(p->psz/CLHEP::GeV)} );
+      mcp.setMomentumAtEndpoint( {MT(p->pex/CLHEP::GeV),MT(p->pey/CLHEP::GeV),MT(p->pez/CLHEP::GeV)} );
 
-#if edm4hep_VERSION < EDM4HEP_VERSION(0, 10, 3)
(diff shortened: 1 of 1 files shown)


DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -354,14 +354,11 @@ void Geant4Output2EDM4hep::saveParticles(Geant4ParticleMap* particles)    {
       const G4ParticleDefinition* def = p.definition();
       auto mcp = m_particles.create();
       mcp.setPDG(p->pdgID);
+      // Because EDM4hep is switching between vector3f[loat] and vector3d[ouble]
+      using MT = decltype(std::declval<decltype(std::declval<edm4hep::MCParticle>().getMomentum())>().x);
```
tmadlener, line 358:
```suggestion
      using MT = decltype(std::declval<edm4hep::MCParticle>().getMomentum().x);
```
I suppose you tried the version with `std::declval` less and it didn't work?

andresailer: I tried some version with a decl* less, but those didn't work.
Your proposal works.

tmadlener: Potentially, we should also throw a `std::remove_cv_t<decltype(...)>` in there to get rid of a potential `const` that might be there because `getMomentum` will return `a const edm4hep::Vector3x&`. (Mabye we also need a `std::remove_reference_t` as well.

Can you check what `MT` is deduced to be here?

```cpp
template<typename T>
struct TypeDebug;
```
and then
```cpp
TypeDebug mtType<MT>{};
```
should lead to a compilation error that tells us the type.

tmadlener: It should be safe without any of the `remove_cv` and `remove_reference` stuff: https://godbolt.org/z/o3zPcodd4

andresailer: Thanks for checking! Let's merge this then.

tmadlener:
If I had remembered this earlier, we could have also used this: https://github.com/key4hep/EDM4hep/blob/e27064f91cb1a8cdcbb840c6ff794dd1096f464a/utils/include/edm4hep/utils/vector_utils.h#L130-L132

PR #1252, opened 2024-04-10
Fix usage of podio pre-processor version checks

tmadlener:
BEGINRELEASENOTES
- DDDigiEDM4hep: Switch to the correct podio pre-processor version checks for switching to the non-deprecated readers / writers. Also simplify this such that choosing which type(name) to use is handled in one place rather than several.
- Geant4Output2EDM4hep: Switch to a non-deprecated setter for setting the MCParticle for a SimTrackerHit once EDM4hep ships with it.

ENDRELEASENOTES

### DIFF
--- a/DDDigi/io/DigiEdm4hepInput.cpp
+++ b/DDDigi/io/DigiEdm4hepInput.cpp
@@ -16,12 +16,16 @@
 #include "DigiIO.h"
 
 // podio/edm4hep include files
-#include <podio/Frame.h>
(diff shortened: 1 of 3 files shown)


tmadlener:
The failing tests seem unrelated to these changes(?)

andresailer:
Yes, t_ClientTests_sim_SiliconBlockGFlash_LONGTEST has been failing since some Geant4 update in the stacks. Only fails in G4 units

andresailer:
Good to merge?

tmadlener:
Just added back an include that I removed accidentally. Otherwise, yes.

PR #1254, opened 2024-04-19
CMake: drop possible use of PyROOT

andresailer:
We no longer test with ROOT 6.22, and in 6.32 PyROOT is coming back but subtly fails on macOS

BEGINRELEASENOTES
- CMake: drop possible use of ROOT::PyROOT target, always use ROOT::ROOTTPython
- Python: remove the use of (dd)six, fix issues pointed out by new version of flake8

ENDRELEASENOTES

Accidentally picking up PyROOT leads to
```
ERROR: failed to load libDDG4Python.1.28.dylib: dlopen(libDDG4Python.1.28.dylib, 0x0005): symbol not found in flat namespace '__ZN9TPyReturnD1Ev'
...
ERROR: failed to load libDDPythonPlugins.1.28.dylib: dlopen(libDDPythonPlugins.1.28.dylib, 0x0005): symbol not found in flat namespace '__ZN7TPython4EvalEPKc'
```


andresailer:
There is a new version of flake8 in the nightly stack giving errors in python-lint

PR #1258, opened 2024-05-03
DDSim: correct the number of events when running over all events (-1)

### LINKED ISSUE #1257: [ddsim] Get total number of events and time per event right
I realized while running on an externally generated file, with the option `-N -1`, that the total number of event and the time per event were wrong. 

There is one too many event in e.g. `m_evtTotal`: https://github.com/AIDASoft/DD4hep/blob/master/DDG4/src/Geant4GeneratorActionInit.cpp#L48 

```
Geant4Output2EDM4hep INFO  +++ Saving EDM4hep event 0 run 0.
GenerationInit   INFO  +++ Initializing event 2. Within run:0 event 2.
UI               INFO  ++ End of file reached, ending run...
GenerationInit   WARN  +++ Finished run 0 after 2 events (2 events in total)
```

while I have only one event.

And the time per event is reported as negative due to the -1 flag:

```
DDSim            INFO DDSim            INFO  StartUp Time: 8.60 s, Event Processing: 351.13 s (-351.13 s/Event)
```

Here is a reproducer [1], let me know if you need a more streamlined one.

[1] 
```
source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh
ddsim --compactFile $K4GEO/FCCee/IDEA/compact/IDEA_o1_v02/IDEA_o1_v02.xml --inputFiles /afs/cern.ch/user/b/brfranco/work/public/background_studies/k4geo/pairs_1.pairs --outputFile IDEA_01_v03_pairs_1.root -N -1
```

andresailer:
BEGINRELEASENOTES
- DDSim: correct the number of events when running over all events (-1), fixes #1257 
- GenerationActionInit: declare properties to access number of processed runs (numberOfRuns) and events (numberOfEVents)
- DDG4: decode or eval all properties out of str or cppyy.gbl.string types

ENDRELEASENOTES

### DIFF
--- a/DDG4/src/Geant4GeneratorActionInit.cpp
+++ b/DDG4/src/Geant4GeneratorActionInit.cpp
@@ -30,6 +30,8 @@ Geant4GeneratorActionInit::Geant4GeneratorActionInit(Geant4Context* ctxt, const 
   InstanceCount::increment(this);
   context()->kernel().runAction().callAtEnd(this,&Geant4GeneratorActionInit::end);
   context()->kernel().runAction().callAtBegin(this,&Geant4GeneratorActionInit::begin);
+  declareProperty("numberOfEvents", m_evtTotal);
+  declareProperty("numberOfRuns",  m_evtRun);
 }
 
 /// Default destructor
--- a/DDG4/python/DDG4.py
+++ b/DDG4/python/DDG4.py
@@ -304,11 +304,11 @@ def _get(self, name):
   a = Interface.toAction(self)
   ret = Interface.getProperty(a, name)
   if ret.status > 0:
-    return ret.data
+    return _evalProperty(ret.data)
   elif hasattr(self.action, name):
-    return getattr(self.action, name)
+    return _evalProperty(getattr(self.action, name))
   elif hasattr(a, name):
-    return getattr(a, name)
+    return _evalProperty(getattr(a, name))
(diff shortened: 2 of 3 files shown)


DDG4/src/Geant4GeneratorActionInit.cpp
```diff
@@ -30,6 +30,8 @@ Geant4GeneratorActionInit::Geant4GeneratorActionInit(Geant4Context* ctxt, const
   InstanceCount::increment(this);
   context()->kernel().runAction().callAtEnd(this,&Geant4GeneratorActionInit::end);
   context()->kernel().runAction().callAtBegin(this,&Geant4GeneratorActionInit::begin);
+  declareProperty("numberOfEvents", m_evtTotal);
```
andresailer, line 33: @MarkusFrankATcernch 
Is this the best way to let me read the number of events from the plugin, or is there something else that would be cleaner?

andresailer:
> @andresailer In principle there is a property Geant4Kernel::NumEvents, which defines the number of events to be processed. This property is then passed to Geant4Exec, where the G4 machinery is invoked. Does this property not work or is this not sufficient to do the computation?

But here the number of events to be processed is going to be `-1`, so I want to check the actual counter of events processed.

MarkusFrankATcernch:
@andresailer Correct. 
Then this approach looks correct.

andresailer:
Thanks @MarkusFrankATcernch

PR #1265, opened 2024-05-16
Fix various problems indicated by coverity

MarkusFrankATcernch:
BEGINRELEASENOTES
- Fix some coverity problems

ENDRELEASENOTES

### DIFF
--- a/DDG4/src/Geant4Kernel.cpp
+++ b/DDG4/src/Geant4Kernel.cpp
@@ -94,7 +94,7 @@ Geant4Kernel::Geant4Kernel(Detector& description_ref)
   declareProperty("OutputLevel",          m_outputLevel = DEBUG);
   declareProperty("NumEvents",            m_numEvent = 10);
   declareProperty("OutputLevels",         m_clientLevels);
-  declareProperty("NumberOfThreads",      m_numThreads);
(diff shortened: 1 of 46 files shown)


DDG4/src/Geant4Kernel.cpp
```diff
@@ -94,7 +94,7 @@ Geant4Kernel::Geant4Kernel(Detector& description_ref)
   declareProperty("OutputLevel",          m_outputLevel = DEBUG);
   declareProperty("NumEvents",            m_numEvent = 10);
   declareProperty("OutputLevels",         m_clientLevels);
-  declareProperty("NumberOfThreads",      m_numThreads);
+  declareProperty("NumberOfThreads",      m_numThreads = 0);
```
andresailer, line 97: Should this be 1, or does it not matter?

MarkusFrankATcernch: I think for single threading it does not matter. For multi threading it should be set properly.

MarkusFrankATcernch: It is set in the constructor explicitly to "0", so in principle the re-assignment in `declareProperty(...)` is redundant.

andresailer: Ok, thanks!

PR #1280, opened 2024-06-21
Allow finding newer versions of podio (for EDM4hep IO)

jmcarcell:
After podio changed to 1.0.0 (see https://github.com/AIDASoft/podio/commit/072c7dce699a2dc90f225cb0a75a4f2ce9677cf8), cmake won't run when asking for version 0.16.3. I think we can remove the requirement and let the build systems figure which versions are compatible with which.

The error when configuring is:

```
  >> 61    CMake Error at CMakeLists.txt:158 (find_package):
     62      Could not find a configuration file for package "podio" that is co
           mpatible
     63      with requested version "0.16.3".
     64    
     65      The following configuration files were considered but not accepted
           :
     66    
     67        /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2024-06-21/x86_64-a
           lmalinux9-gcc11.4.1-opt/podio/072c7dce699a2dc90f225cb0a75a4f2ce9677c
           f8_develop-52hwbt/lib64/cmake/podio/podioConfig.cmake, version: 1.0.
           0
```

BEGINRELEASENOTES
- CMake: Allow finding version 1.0 (and newer) for podio

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,7 +155,7 @@ endif()
 
 if(DD4HEP_USE_EDM4HEP)
   find_package(EDM4HEP REQUIRED)
-  find_package(podio 0.16.3 REQUIRED)
+  find_package(podio REQUIRED)
 #  DD4HEP_SETUP_EDM4HEP_TARGETS()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,7 +155,13 @@ endif()
 
 if(DD4HEP_USE_EDM4HEP)
   find_package(EDM4HEP REQUIRED)
-  find_package(podio REQUIRED)
+  # we need podio with Frame support (>=0.16.3)
+  # podio is "SameMajorVersion" compatible
+  find_package(podio 0.16.3)  # this will not find 1.0 and newer
+  if(NOT podio_FOUND)
+    # we try to find a newer version now
+    find_package(podio 1.0 REQUIRED)
+  endif()
 #  DD4HEP_SETUP_EDM4HEP_TARGETS()
 endif()
 


wdconinc:
I disagree with this approach of removing dependency version information. DD4hep requires a certain set of podio features, namely those provided by 0.16.3. Podio promises (through the SameMajorVersion policy) that any newer version with same major version can be expected to provide those features, but a different major version does not come with that promise. Failing on build configuration when you attempt to use podio 1.0.0 is exactly what should happen here.

Podio could change the policy to AnyNewerVersion (which then essentially means that you cannot use major version updates to *remove* API functionality).

CMakeLists.txt
```diff
@@ -155,7 +155,7 @@ endif()
 
 if(DD4HEP_USE_EDM4HEP)
   find_package(EDM4HEP REQUIRED)
-  find_package(podio 0.16.3 REQUIRED)
+  find_package(podio REQUIRED)
```
wdconinc, line 158: What about this approach? [1]
```suggestion
  find_package(podio 0.16.3)
  if(NOT podio_FOUND)
    find_package(podio 1.0 REQUIRED)
  endif()
```
[1]: https://github.com/eic/EDM4eic/pull/84

wdconinc:
> Can someone remind me again, why we search explicitly for podio and not just EDM4hep here?

This is because of the need for `podio::Frame` support (https://github.com/AIDASoft/DD4hep/commit/57a7bf2f20f38b27b854ecbc1554af047fc40b85), which EDM4hep separately doesn't care about.

CMakeLists.txt
andresailer:
```suggestion
    # we need podio with Frame support (>=0.16.3)
    # podio is "SameMajorVersion" compatible
    find_package(podio 0.16.3)  # this will not find 1.0 and newer
  if(NOT podio_FOUND)
    # we try to find a newer version now
    find_package(podio 1.0 REQUIRED)
  endif()
```

Issue #1292, opened 2024-07-08
k4geo CI failing, potentially caused by last DD4hep PR


atolosadelgado:
### Check duplicate issues.

- [X] Checked for duplicates

### Goal

Hi,

The k4geo repository stores the geometry of detectors for future colliders. I noticed that CI failed with nightlies over the weekend:
https://github.com/key4hep/k4geo/actions/runs/9834053300/job/27145226265

I suspect about one of the last PR, https://github.com/AIDASoft/DD4hep/pull/1288

Please, let us know if we can provide further information that could be useful to solve the issue.

### Operating System and Version

Alma9

### compiler

GCC 11.4

### ROOT Version

6.30/06

### DD4hep Version

master

### Reproducer

the following steps will reproduce the error, in Alma9 (or Ubuntu22)

```
source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh
git clone https://github.com/key4hep/k4geo.git
cd k4geo/
cmake -B build -S . -D CMAKE_INSTALL_PREFIX=install
cmake --build build -j 4 --target install
k4_local_repo
cd build
ctest -V -R t_test_steeringFile
```

The error message is the following:
```
13: Traceback (most recent call last):
13:   File "/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2024-07-08/x86_64-almalinux9-gcc11.4.1-opt/dd4hep/fe64884cac492f18574a8e2af37560667f7f3802_develop-6pb2jj/bin/ddsim", line 25, in <module>
13:     sys.exit(RUNNER.run())
13:   File "/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2024-07-08/x86_64-almalinux9-gcc11.4.1-opt/dd4hep/fe64884cac492f18574a8e2af37560667f7f3802_develop-6pb2jj/lib/python3.10/site-packages/DDSim/DD4hepSimulation.py", line 494, in run
13:     self.__setupSensitiveDetectors(detectors, function, defFilter, defAction, abort)
13:   File "/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2024-07-08/x86_64-almalinux9-gcc11.4.1-opt/dd4hep/fe64884cac492f18574a8e2af37560667f7f3802_develop-6pb2jj/lib/python3.10/site-packages/DDSim/DD4hepSimulation.py", line 652, in __setupSensitiveDetectors
13:     seq, act = setupFunction(det, action)
[...]

PR #1293, opened 2024-07-11
Fix a crash when an action has dictionaries and is not hashable

jmcarcell:
Fixes https://github.com/AIDASoft/DD4hep/issues/1292, introduced first in https://github.com/AIDASoft/DD4hep/pull/1288. The issue happens when passing an action that has a dictionary, like in https://github.com/key4hep/k4geo/blob/2e1bba216879562382ea5599b738c14c0938b540/example/steeringFile.py#L22.

BEGINRELEASENOTES
- Fix a crash when passing to ddsim an action has dictionaries and is not hashable.
- Refactor the functions in DDG4.py to setup a calorimeter and a tracker detector into a single one since the logic is the same. In this new function, use a try-except and if the type is not hashable use the default value.

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDG4.py
+++ b/DDG4/python/DDG4.py
@@ -678,15 +678,9 @@ class Geant4:
     retType = detType
     if detType is None:
       retType = self.sensitive_types['calorimeter' if isCalo else 'tracker']
-    elif detType is not None:
-      try:
-        retType = self.sensitive_types[detType]
-      # KeyError = not found in the dictionary
-      # TypeError = detType is not a hashable type
-      except (KeyError, TypeError):
-        pass
-      except Exception as X:
-        raise X
+    # detType is a tuple when an action with parameters in a dictionary is passed
+    elif not isinstance(detType, tuple) and detType in self.sensitive_types:
+      retType = self.sensitive_types[detType]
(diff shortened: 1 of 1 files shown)


andresailer:
Can you add some mention of `ddsim` into the release notes part?

> use a try-except and if the type is not hashable used the default value

I am not sure that should be done. Why is a non-hashable type given in the first place?
Probably, we shouldn't silently ignore this input, but either refuse to do something, or make it work as intended by the user.

jmcarcell:
So having another look at this, I updated the logic and the parameters are being set here: https://github.com/AIDASoft/DD4hep/blob/master/DDG4/python/DDG4.py#L664. I'm not sure how to test if a sensitive action is running though, the example in k4geo with the crash seems to create it but not actually run it.
In summary:
- Before #1288, if the type passed is None the default is used, otherwise it's passed
- After #1288, same as before but also if the type is found in the dictionary keys then the corresponding value is used - the lookup is done whenever the type is not None, this causes the crash
- After this PR, same as #1288 but only in the case when it's not a tuple (an action with parameters) we do the lookup in the dictionary

MarkusFrankATcernch:
Change in PR 1288 reverted. See PR 1301.

PR #1294, opened 2024-07-11
Adding CylindricalGridPhiZ files

ybedfer:
Segmentation , plugin and dictionary are modified to include "CylindricalGridPhiZ".
- "CylindricalGridPhiZ" is needed for my cylindrically shaped MPGD detector in ePIC at EIC.
- The class derives from the "CylindricalSegmentation" class.
- It reads its "gridSize" parameters the usual way. Optional "offset" parameters, the same.
- It has an extra type of parameter, viz. "radius", used by the "position" and "cellDimensions" methods, to transform from the phi,Z representation to X,Y.

BEGINRELEASENOTES
- Adding a "CylindricalGridPhiZ" segmentation class.
   - N.B.: the class has a "radius" data member, registered by the constructors.

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/CylindricalGridPhiZ.h
+++ b/DDCore/include/DD4hep/CylindricalGridPhiZ.h
@@ -12,8 +12,8 @@
 //  \version  1.0
 //
 //==========================================================================
-#ifndef DD4HEP_CARTESIANGRIDXYZ_H
-#define DD4HEP_CARTESIANGRIDXYZ_H 1
+#ifndef DD4HEP_CYLINDRICALGRIDPHIZ_H
+#define DD4HEP_CYLINDRICALGRIDPHIZ_H 1
 
 // Framework include files
 #include <DD4hep/Segmentations.h>
@@ -22,12 +22,12 @@
 namespace dd4hep {
 
   /// Namespace for base segmentations
-  namespace DDSegmentation  {    class CartesianGridXYZ;  }
-  
+  namespace DDSegmentation  {    class CylindricalGridPhiZ;  }
+    
   /// We need some abbreviation to make the code more readable.
-  typedef Handle<SegmentationWrapper<DDSegmentation::CartesianGridXYZ> > CartesianGridXYZHandle;
-  
-  /// Implementation class for the grid XYZ segmentation.
+  typedef Handle<SegmentationWrapper<DDSegmentation::CylindricalGridPhiZ> > CylindricalGridPhiZHandle;
+ 
+  /// Implementation class for the grid PhiZ segmentation.
   /**
    *  Concrete user handle to serve specific needs of client code
    *  which requires access to the base functionality not served
@@ -48,56 +48,52 @@ namespace dd4hep {
    *  \version 1.0
    *  \ingroup DD4HEP_CORE
    */
-  class CartesianGridXYZ : public CartesianGridXYZHandle {
+  class CylindricalGridPhiZ : public  CylindricalGridPhiZHandle {
   public:
     /// Default constructor
-    CartesianGridXYZ() = default;
+    CylindricalGridPhiZ() = default;
     /// Copy constructor
-    CartesianGridXYZ(const CartesianGridXYZ& e) = default;
+    CylindricalGridPhiZ(const CylindricalGridPhiZ& e) = default;
     /// Copy Constructor from segmentation base object
-    CartesianGridXYZ(const Segmentation& e) : Handle<Object>(e) {}
+    CylindricalGridPhiZ(const Segmentation& e) : Handle<Object>(e) {}
     /// Copy constructor from handle
-    CartesianGridXYZ(const Handle<Object>& e) : Handle<Object>(e) {}
+    CylindricalGridPhiZ(const Handle<Object>& e) : Handle<Object>(e) {}
     /// Copy constructor from other polymorph/equivalent handle
     template <typename Q>
-    CartesianGridXYZ(const Handle<Q>& e) : Handle<Object>(e) {}
+    CylindricalGridPhiZ(const Handle<Q>& e) : Handle<Object>(e) {}
     /// Assignment operator
-    CartesianGridXYZ& operator=(const CartesianGridXYZ& seg) = default;
+    CylindricalGridPhiZ& operator=(const CylindricalGridPhiZ& seg) = default;
     /// Equality operator
-    bool operator==(const CartesianGridXYZ& seg) const
-    {  return m_element == seg.m_element;    }
+    bool operator==(const CylindricalGridPhiZ& seg) const
+    {  return m_element == seg.m_element;           }
+
     /// determine the position based on the cell ID
     Position position(const CellID& cellID) const;
     /// determine the cell ID based on the position
     CellID cellID(const Position& local, const Position& global, const VolumeID& volID) const;
-    /// access the grid size in X
-    double gridSizeX() const;
-    /// access the grid size in Y
-    double gridSizeY() const;
+    /// access the grid size in phi
+    double gridSizePhi() const;
     /// access the grid size in Z
     double gridSizeZ() const;
-    /// set the grid size in X
-    void setGridSizeX(double cellSize) const;
-    /// set the grid size in Y
-    void setGridSizeY(double cellSize) const;
-    /// set the grid size in Z
-    void setGridSizeZ(double cellSize) const;
-    /// access the coordinate offset in X
-    double offsetX() const;
-    /// access the coordinate offset in Y
-    double offsetY() const;
+    /// access the radius
+    double radius() const;
+    /// access the coordinate offset in phi
+    double offsetPhi() const;
     /// access the coordinate offset in Z
     double offsetZ() const;
-    /// set the coordinate offset in X
-    void setOffsetX(double offset) const;
-    /// set the coordinate offset in Y
-    void setOffsetY(double offset) const;
+    
+    /// set the grid size in phi
+    void setGridSizePhi(double cellSize) const;
+    /// set the grid size in Z
+    void setGridSizeZ(double cellSize) const;
+    /// set the coordinate offset in phi
+    void setOffsetPhi(double offset) const;
     /// set the coordinate offset in Z
     void setOffsetZ(double offset) const;
-    /// access the field name used for X
-    const std::string& fieldNameX() const;
-    /// access the field name used for Y
-    const std::string& fieldNameY() const;
+    /// set the radius
+    void setRadius(double radius);
+    /// access the field name used for phi
+    const std::string& fieldNamePhi() const;
     /// access the field name used for Z
     const std::string& fieldNameZ() const;
     /** \brief Returns a vector<double> of the cellDimensions of the given cell ID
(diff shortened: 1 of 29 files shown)


DDCore/include/DD4hep/CylindricalGridPhiZ.h
```diff
@@ -0,0 +1,111 @@
...
+   *
+   *  Note:
+   *  We only check the validity of the underlying handle.
+   *  If for whatever reason the implementation object is not valid
+   *  This is not checked.
+   *  In principle this CANNOT happen unless some brain-dead has
+   *  fiddled with the handled object directly.....
+   *
+   *  Note:
+   *  The handle base corrsponding to this object in for
```
andresailer, line 44:
```suggestion
   *  The handle base corresponding to this object in for
```

DDCore/include/DD4hep/CylindricalGridPhiZ.h
```diff
@@ -0,0 +1,111 @@
...
+    const std::string& fieldNamePhi() const;
+    /// access the field name used for Z
+    const std::string& fieldNameZ() const;
+    /** \brief Returns a vector<double> of the cellDimensions of the given cell ID
+        in natural order of dimensions, e.g., dx/dy/dz, or dr/r*dPhi
+
+        Returns a vector of the cellDimensions of the given cell ID
+        \param cellID is ignored as all cells have the same dimension
+        \return std::vector<double> size 2:
+        -# size in x
```
andresailer, line 105: Does this need adaptation?

DDCore/include/DDSegmentation/CylindricalGridPhiZ.h
```diff
@@ -0,0 +1,127 @@
...
+      void setFieldNameZ(const std::string& fieldName) {
+        _zId = fieldName;
+      }
+      /** \brief Returns a vector<double> of the cellDimensions of the given cell ID
+          in natural order of dimensions, e.g., dx/dy/dz, or dr/r*dPhi
+
+          Returns a vector of the cellDimensions of the given cell ID
+          \param cellID is ignored as all cells have the same dimension
+          \return std::vector<double> size 2:
+          -# size in x
```
andresailer, line 102: Does this need adaptation?

DDCore/src/CylindricalGridPhiZ.cpp
```diff
@@ -0,0 +1,94 @@
...
+const std::string& CylindricalGridPhiZ::fieldNamePhi() const {
+  return access()->implementation->fieldNamePhi();
+}
+
+/// access the field name used for Z
+const std::string& CylindricalGridPhiZ::fieldNameZ() const {
+  return access()->implementation->fieldNameZ();
+}
+
+/** \brief Returns a vector<double> of the cellDimensions of the given cell ID
```
andresailer, line 83: Probably don't need this comment in the cpp file, too much repetition, other wise, does the return need adaptation?

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
```diff
@@ -0,0 +1,106 @@
...
+        _decoder->set( cID,_phiId, positionToBin(phi, _gridSizePhi, _offsetPhi) );
+	_decoder->set( cID,_zId,   positionToBin(Z,   _gridSizeZ,   _offsetZ) );
+
+	return cID ;
+}
+
+std::vector<double> CylindricalGridPhiZ::cellDimensions(const CellID&) const {
+#if __cplusplus >= 201103L
+  return {_gridSizePhi, _gridSizeZ};
+#else
```
andresailer, line 96: We don't need the else anymore.

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
```diff
@@ -0,0 +1,106 @@
...
+        CellID cID = vID ;
+        _decoder->set( cID,_phiId, positionToBin(phi, _gridSizePhi, _offsetPhi) );
+	_decoder->set( cID,_zId,   positionToBin(Z,   _gridSizeZ,   _offsetZ) );
+
+	return cID ;
+}
+
+std::vector<double> CylindricalGridPhiZ::cellDimensions(const CellID&) const {
+#if __cplusplus >= 201103L
+  return {_gridSizePhi, _gridSizeZ};
```
andresailer, line 95: And this means the function docstrings need to be adapted.

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
```diff
@@ -0,0 +1,106 @@
...
+#else
+  std::vector<double> cellDims(2,0.0);
+  cellDims[0] = _radius*_gridSizePhi;
```
andresailer, line 98: Note here is the radius, above it is not?

Is this wrong somewhere you copied it from as well?

DDCore/src/CylindricalGridPhiZ.cpp
ybedfer: Indeed, the comment is not useful. I could remove it. But does such a change belong to my intended commit? I could prepare a dedicated commit, where all such kind of cosmetics would be fixed at once: all of the "corrsponding" mentioned supra, and other such mistakes ('Y' instead of 'Z' in CartesianGridXZ, size = 2 instead of 3 in CartesianGridXYZ,, ...) and a simplification of comments.

DDCore/include/DD4hep/CylindricalGridPhiZ.h
ybedfer: No. This is copied from CartesianGridXZ. One could then expect "size in x" to have to changed to "size in phi", but what the calling program is expecting is a quantity of the dimension of a length, in the local coordinate system, to serve as an uncertainty for the hit. See, e.g., "EICrecon/src/algorithms/tracking/TrackerHitReconstruction.cc"

DDCore/include/DDSegmentation/CylindricalGridPhiZ.h
ybedfer: Same answer as for previous remark.

DDCore/src/CylindricalGridPhiZ.cpp
ybedfer: I didn't quite get the question concerning the "return". As already said, I copied all of my CylindricalGridPhiZ from CartesianGridXZ. I cannot see why something would have to be changed here...

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
ybedfer: I made a mistake here: line 94 should read "return {_radius*_gridSizePhi, _gridSizeZ};" so that the returned quantity be of the dimension of a length. I intend to commit that modification. Concerning the cancellation of the else, I would make the same remark as before: this kind of modification could instead be included in a dedicated commit that does the modification for all of the segmentation classes.

DDCore/include/DD4hep/CylindricalGridPhiZ.h
ybedfer: I committed your suggested change (or so I think: I'm not familiar with this answer-to-reviewer procedure) . Turns out that "grep -R corrsponding" returns 14 matches in "./DD4hep/DDCore".

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
ybedfer: Again, I don't quite get your question. By "doctring", are you alludiing to the remark about "const std::string& CylindricalGridPhiZ::fieldNameZ()" you make supra?

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
ybedfer: Yes, indeed. I made a mistake. Sorry. As I already said supra, I intend to commit the change
"return {_gridSizePhi, _gridSizeZ};" to "return {_radius*_gridSizePhi, _gridSizeZ};".

DDCore/include/DD4hep/CylindricalGridPhiZ.h
ybedfer: I have added a purely cosmetics commit, w/ all of your suggestions concerning comments.

DDCore/include/DD4hep/CylindricalGridPhiZ.h
ybedfer: I have added a purely cosmetics commit, w/ all of your suggestions concerning comments. In the present case, I changed the comment "size in x" for "size in x = R * size in phi"

DDCore/include/DDSegmentation/CylindricalGridPhiZ.h
ybedfer: I made the same modification as for the CylindricalGrid header of the DD4hep subdirectory. But in a distinct commit.

DDCore/src/CylindricalGridPhiZ.cpp
ybedfer: I have added a purely cosmetics commit, w/ this suggestion of yours implemented.

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
ybedfer: grep -R "__cplusplus >= 201103L" returns 21 matches from different corner of the DD4hep distribution. I then do not feel like making any modification here.

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
ybedfer: OK. Now I understand that the "docstring" you are referring is the comment. Then, in the cosmetics commit I am mentioning supra, I have updated it.

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
ybedfer: To the question: "Is this wrong somewhere you copied it from as well?". No, having to multiply by the radius is something specific to CylindricalGrid's. I simply made a mistake myself, forgetting to port the modification I had done for less modern C++ versions, to the "__cplusplus >= 201103L" case.

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
andresailer: We haven't done clean up in other places of the code. Doesn't mean we need to introduce new pieces that we just need to clean up later. DD4hep doesn't compile with anything less than c++14, 

https://github.com/AIDASoft/DD4hep/blob/fe64884cac492f18574a8e2af37560667f7f3802/CMakeLists.txt#L59-L60

and we only probably test with c++17 or newer.

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
andresailer: Sorry, `docstring` is apparently a pythonism. But given my other comments about the documentation of the return value for this (or similar) function I thought context was clear enough.

DDCore/src/segmentations/CylindricalGridPhiZ.cpp
ybedfer: I removed the #if,#else construct and committed.

Issue #1291, opened 2024-07-08
Increase in time consumption of the function `GeoHandler::i_collect` after PR #1277


SanghyunKo:
### Check duplicate issues.

- [X] Checked for duplicates

### Goal

To resolve a scalability issue caused by `std::find` for highly granular detectors

### Operating System and Version

Alma 9

### compiler

GCC 11

### ROOT Version

6.30/06

### DD4hep Version

nightly

### Reproducer

I used my own detector for testing, but I bet you can reproduce it by scaling up the `BoxOfStraws` example without SD

### Additional context

While moving from the old release to the nightly release, we realized that the overhead time of constructing geometry for a highly granular detector has significantly increased. After doing some tests, we tracked down PR #1277, which seems to be causing the issue. (the nightly build `2024-06-05` is fine, but the `2024-06-15` is too slow - and this is the only PR in DD4hep between the two)

We ran a `callgrind` profiler and figured out that `GeoHandler::i_collect` has significantly increased timing consumption. I have attached the results for two nightly builds below.
[0605_result_callgrind.out.3261198.txt](https://github.com/user-attachments/files/16124018/0605_result_callgrind.out.3261198.txt)
[0615_result_callgrind.out.1226121.txt](https://github.com/user-attachments/files/16124029/0615_result_callgrind.out.1226121.txt)

I assume that `std::find` [here](https://github.com/AIDASoft/DD4hep/pull/1277/files#diff-4ab4678be68e428e34209aaa4003030d0791273a55bc75c78569e2774fd1ddbaR154) is having a scalability issue for high granular detectors because `std::set::emplace` has O(log(N)) but `std::find` has O(N) time complexity.

But, to be honest, I'm not sure how to fix this without ruining #1271 again... so I'd be happy to hear ideas from the developers.

FYI @BrieucF @atolosadelgado @swkim95 @s6anloes

andresailer:
Fixed by @SanghyunKo

PR #1298, opened 2024-07-18
Improve lookup speed of GeoHandler (issue #1291)

SanghyunKo:
BEGINRELEASENOTES
- Improve lookup speed of `GeoHandler::i_collect` by using `std::set` instead of `std::find` (issue #1291)

ENDRELEASENOTES

Tried to add another container of `std::map<int, std::set<const TGeoNode*>>` for the lookup purpose as @andresailer suggested in #1291. The simulation now seems running fine (as before #1277) even with a highly granular detector, but I'd appreciate suggestions from core developers if there is any.

### DIFF
--- a/DDCore/src/GeoHandler.cpp
+++ b/DDCore/src/GeoHandler.cpp
@@ -55,12 +55,14 @@ namespace {
 /// Default constructor
 detail::GeoHandler::GeoHandler()  {
   m_data = new std::map<int, std::vector<const TGeoNode*> >();
+  m_set_data = new std::map<int, std::set<const TGeoNode*> >();
 }
 
 /// Initializing constructor
 detail::GeoHandler::GeoHandler(std::map<int, std::vector<const TGeoNode*> >* ptr,
-                               std::map<const TGeoNode*, std::vector<TGeoNode*> >* daus)
-  : m_data(ptr), m_daughters(daus)
+                std::map<int, std::set<const TGeoNode*> >* ptr_set,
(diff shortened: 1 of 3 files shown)


DDCore/src/GeoHandler.cpp
```diff
@@ -55,25 +55,35 @@ namespace {
...
+
   m_data = nullptr;
+  m_set_data = nullptr;
 }
 
 std::map<int, std::vector<const TGeoNode*> >* detail::GeoHandler::release() {
   std::map<int, std::vector<const TGeoNode*> >* d = m_data;
   m_data = nullptr;
+
+  delete m_set_data;
```
andresailer, line 84: This _must_ have a comment

DDCore/src/GeoHandler.cpp
```diff
@@ -150,8 +166,8 @@ detail::GeoHandler& detail::GeoHandler::i_collect(const TGeoNode* /* parent */,
     }
   }
   /// Collect the hierarchy of placements
-  auto& vec = (*m_data)[level];
-  if(std::find(vec.begin(), vec.end(), current) == vec.end()) {
+  /// perform lookup using std::set::emplace (faster than std::find for the large number of geometries)
```
andresailer, line 169:
```suggestion
  /// perform lookup using std::set::emplace (faster than std::find for very large number of volumes)
```

andresailer (approved):

PR #1300, opened 2024-07-22
Add a signal handler for SIGINT when building a detector

jmcarcell:
To cancel while building a geometry it seems I have to kill the process or the terminal where it's running. With this change it's possible to cancel by doing a Ctrl-C, and it seems to be working both in Python and C++:

``` python
import dd4hep
det = dd4hep.Detector.getInstance()
det.fromCompact(f'{os.environ["k4geo"]}/FCCee/CLD/compact/CLD_o2_v05/CLD_o2_v05.xml')
```
And then press Ctrl-C:
```
Compact          INFO  ++ Converted subdetector:LumiCalBackShield of type LumiCal_o1_v01 [calorimeter]
Calling Converter<Compact>::operator() with element: lccdd
Utilities        INFO  +++ setDetectorTypeFlags for detector: Vertex not set.
Compact          INFO  ++ Converted subdetector:Vertex of type DD4hep_SubdetectorAssembly
Compact          INFO  ++ Converted subdetector:VertexBarrel of type ZPlanarTracker [tracker]
XXX Vertex endcap layers: 6
Compact          INFO  ++ Converted subdetector:VertexEndcap of type VertexEndcap_o1_v06 [tracker]
Compact          INFO  ++ Converted subdetector:VertexVerticalCable of type TrackerEndcapSupport_o1_v02
VertexEndcap     INFO  +++ Processing SurfaceInstaller for subdetector: 'VertexEndcap'
DD4hep_GenericSurfaceInstallerPlugin: argument[1] = dimension = 2
DD4hep_GenericSurfaceInstallerPlugin: argument[2] = u_x = -1
DD4hep_GenericSurfaceInstallerPlugin: argument[3] = v_z = 1
DD4hep_GenericSurfaceInstallerPlugin: argument[4] = n_y = 1
DD4hep_GenericSurfaceInstallerPlugin: vectors: u( -1 , 0 , 0) v( 0 , 0 , 1) n( 0 , 1 , 0) o( 0 , 0 , 0)
Calling Converter<Compact>::operator() with element: lccdd
Utilities        INFO  +++ setDetectorTypeFlags for detector: InnerTrackers not set.
Compact          INFO  ++ Converted subdetector:InnerTrackers of type DD4hep_SubdetectorAssembly
^CCaught signal SIGINT, exiting...
```

BEGINRELEASENOTES
- Add a signal handler for SIGINT when building a detector, allowing to cancel with Ctrl-C

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/DetectorLoad.cpp
+++ b/DDCore/src/DetectorLoad.cpp
@@ -50,7 +50,6 @@ DetectorLoad::~DetectorLoad() {
 
 /// Process XML unit and adopt all data from source structure.
 void DetectorLoad::processXML(const std::string& xmlfile, xml::UriReader* entity_resolver) {
-
   try {
     xml::DocumentHolder doc(xml::DocumentHandler().load(xmlfile,entity_resolver));
     if ( doc )   {
--- a/DDCore/src/DetectorLoad.cpp
+++ b/DDCore/src/DetectorLoad.cpp
@@ -51,18 +51,6 @@ DetectorLoad::~DetectorLoad() {
 /// Process XML unit and adopt all data from source structure.
 void DetectorLoad::processXML(const std::string& xmlfile, xml::UriReader* entity_resolver) {
 
-  // Install signal handler for SIGINT (Ctrl-C)
-  struct sigaction sigIntHandler;
-
-  sigIntHandler.sa_handler = [](int) {
-    std::cerr << "Caught signal SIGINT, exiting..." << std::endl;
-    exit(1);
-  };
-  sigemptyset(&sigIntHandler.sa_mask);
-  sigIntHandler.sa_flags = 0;
-
-  sigaction(SIGINT, &sigIntHandler, NULL);
-
   try {
     xml::DocumentHolder doc(xml::DocumentHandler().load(xmlfile,entity_resolver));
     if ( doc )   {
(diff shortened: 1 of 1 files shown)


DDCore/src/DetectorLoad.cpp
```diff
@@ -21,6 +21,9 @@
 
 // C/C++ include files
 #include <stdexcept>
+#include <csignal>
```
andresailer, line 24: please sort alphabetically with the existing include

jmcarcell: Done

MarkusFrankATcernch:
I am not a big friend of this PR. In general the rule is always:
-- Generic signals (such as `SIGINT, SIGTERM,` etc) are handled by applications, not by libraries.
-- Only specific signals are handled by libraries e.g. `SIGIO` in an I/O library.
-- libraries should generally avoid to call `::exit()`. A library should only throw an exception or return a failure status.

In the event this behaviour is not wanted, users have to scan source code to find all signals handled internally.
This is not very convenient.

andresailer:
Closing this PR in view of the changes from #1336. But we need to enable the signal handler in ddsim now, I think.

MarkusFrankATcernch:
@andresailer Yes. The handler would have to be enabled in ddsim. The test show which line to call:
```
geant4 = DDG4.Geant4(kernel, tracker='Geant4TrackerCombineAction')
geant4.registerInterruptHandler() 

```

PR #1308, opened 2024-08-09
Allow speeding up RegexSD, add ddsim interface for it

andresailer:
BEGINRELEASENOTES
- Geant4RegexSensitivesConstruction: allow using regex that don't match the full volume path, reduce time by 40% or so
- DDSim: add interface for Geant4RegexSensitivesConstruction to geometry construction

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDSim/Helper/Geometry.py
+++ b/DDG4/python/DDSim/Helper/Geometry.py
@@ -35,7 +35,45 @@ class Geometry(ConfigHelper):
 
     self._dumpDGDML_EXTRA = {"help": "If not empty, filename to dump the Geometry as GDML"}
     self.dumpGDML = ""
+
+    self._regexSDDetectorList = []
+
     self._closeProperties()
+
+  @property
+  def regexSensitiveDetector(self):
+    """Configure a sensitive detector for a given detector matched by regular expression (regex).
+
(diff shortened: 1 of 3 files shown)


DDG4/python/DDSim/Helper/Geometry.py
```diff
@@ -35,8 +35,46 @@ def __init__(self):
...
+        self.__checkRegexKeys(value)
+        self._regexSDDetectorList.append(value)
+    raise RuntimeError(f"Unsupported type for regexSensitiveDetector: {val!r}")
+
+  @staticmethod
+  def __checkRegexKeys(val):
+    """Check the regex SD arguments for required keys."""
+    requiredKeys = ('Detector', 'Match')
+    if not all(key in val for key in requiredKeys):
+      raise RuntimeError(f"RegexSD configuration {val} is missing at least one mandatory key:", requiredKeys)
```
andresailer, line 76:
```suggestion
      raise RuntimeError(f"RegexSD configuration {val} is missing at least one mandatory key: {requiredKeys}")
```

andresailer:
dev4 is also broken in master for this test `t_AlignDet_AlephTPC_load (Failed)`, but not in the tests in the LCG builds https://lcgapp-services.cern.ch/cdash3/tests/6434290

andresailer:
Key4hep stack is missing https://github.com/root-project/root/issues/15986

PR #1314, opened 2024-08-14
Cylinder segment surface

armin-ilg:
BEGINRELEASENOTES
- Adding the capability to VolCylinder to be only a segment of a cylinder >2π needed for ultra-light curved vertex concept.
ENDRELEASENOTES

Hi everyone

I'm trying to make the ultra-light vertex detector concept working in full simulation, using[ k4geo from this PR](https://github.com/key4hep/k4geo/pull/363). As a first step I added the functionality to VolCylinder to feature less than 2π circumference, changing the getLines function. Is there something else needed? Have in the past cylindrical sensitive surfaces actually been used in reconstruction?

Thank you,
Armin

### DIFF
--- a/DDRec/src/Surface.cpp
+++ b/DDRec/src/Surface.cpp
@@ -75,6 +75,8 @@ namespace dd4hep {
     double VolSurfaceBase::innerThickness() const { return _th_i ; }
     double VolSurfaceBase::outerThickness() const { return _th_o ; }
     
+    double VolSurfaceBase::phiTot() const { return _phiTot ; }
+    double VolSurfaceBase::phi0() const { return _phi0 ; }
 
     double VolSurfaceBase::length_along_u() const {
       
@@ -235,7 +237,7 @@ namespace dd4hep {
     }
 
 
-    std::vector< std::pair<Vector3D, Vector3D> > VolSurfaceBase::getLines(unsigned ) {
+    std::vector< std::pair<Vector3D, Vector3D> > VolSurfaceBase::getLines(unsigned) {
       // dummy implementation returning empty set
       std::vector< std::pair<Vector3D, Vector3D> >  lines ;
       return lines ;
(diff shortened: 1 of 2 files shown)


DDRec/src/Surface.cpp
```diff
@@ -1182,12 +1187,14 @@ namespace dd4hep {
 
 
           unsigned n = nMax / 4 ;
-          double dPhi = 2.* ROOT::Math::Pi() / double( n ) ; 
+          double dPhi = phiTot() / double( n ) ; 
 
-          for( unsigned i = 0 ; i < n ; ++i ) {
+          std::cout << "Using phiTot=" << phiTot() << " and phi0=" << phi0() << std::endl;
```
andresailer, line 1192:
```suggestion
```

andresailer:
I converted this to a draft, until the whole chain is confirmed to be fully working.

gaede:
Hi Armin,
I believe that adopting the getLines() as you have done should be all that is needed for Cylinders <2pi as the 'boundary' of the surface is handled by the volume.  We use cylindrical surfaces for tracking in the ILD TPC, see e.g.:
https://github.com/iLCSoft/DDKalTest/blob/master/src/DDCylinderMeasLayer.cc
https://github.com/key4hep/k4geo/blob/main/detector/tracker/TPC10_geo.cpp
Cheers, Frank.

armin-ilg:
> @armin-ilg Is this PR still relevant ? Or can it be closed without merge otherwise?

Hi @MarkusFrankATcernch . No it can be closed without merge. Even with cylinder surfaces, conformal tracking would still need to be adapted to work with that. Instead, I've come up with the solution of representing curved sensors by trapezoidal volumes with flat surfaces, which works with conformal tracking (see https://github.com/key4hep/k4geo/pull/557).

Sorry for not coming back to this earlier.

PR #1315, opened 2024-08-15
Fix constant exception from  Geometry.regexSensitiveDetector

SanghyunKo:
BEGINRELEASENOTES
- DDSim: prevent Geometry.regexSensitiveDetector from always throwing exception when called with argument.

ENDRELEASENOTES

Fixing a (likely) typo from PR #1308, otherwise the `DDSim` interface always throws a Runtime error.
(sorry that I didn't catch this earlier)


andresailer:
Thanks! Bit more than a typo :/

andresailer (approved):

PR #1317, opened 2024-08-23
RegexSD: change ddsim interface to use a dictionary instead of a list, fix infinite loop 

andresailer:
Too much hassle to ensure that the list was not duplicating the entries in the list

The original implementation also caused infinite loop by appending to the list that was iterated over causing ddsim to use up all the memory and not do anything

Added a test with an example ddsim steering file

@SanghyunKo @lopezzot: please note, but this time actually tested to run

BEGINRELEASENOTES
- DDSim: regexSD fix infinite loop, but change the interface from assignment to dictionary entries. Have to ensure only a single entry of a detector is given to the RegexSD

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDSim/Helper/Geometry.py
+++ b/DDG4/python/DDSim/Helper/Geometry.py
@@ -36,45 +36,29 @@ class Geometry(ConfigHelper):
     self._dumpDGDML_EXTRA = {"help": "If not empty, filename to dump the Geometry as GDML"}
     self.dumpGDML = ""
 
-    self._regexSDDetectorList = []
+    self._regexSDDict = {}
 
     self._closeProperties()
 
   @property
   def regexSensitiveDetector(self):
-    """Configure a sensitive detector for a given detector matched by regular expression (regex).
+    """ The map key is the name of the Detector, and 'Match' is a mandatory elements of the dictionary, other Keys are
+        assigned as property to the object. OutputLevel _sets_ the outputlevel of the plugin, so lower numbers mean more
+        output from the plugin.
 
-    'Detector' and 'Match' are mandatory elements of the dictionary, other Keys are assigned as property to the object.
-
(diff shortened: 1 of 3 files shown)


DDG4/python/DDSim/Helper/Geometry.py
```diff
@@ -36,45 +36,29 @@ def __init__(self):
...
+    return self._regexSDDict
 
   @regexSensitiveDetector.setter
   def regexSensitiveDetector(self, val):
-    if not val:
-      return
     if isinstance(val, dict):
-      self.__checkRegexKeys(val)
-      self._regexSDDetectorList.append(val)
-    elif isinstance(val, list):
```
andresailer, line 65: When the ddsim steering file was parsed and values given to the main DD4hepSimulation instance, this caused the list we iterate over to be also appended to, which caused a a growth of the list until memory exhausted. Making sure that this doesn't happen was just causing code bloat, so I decided to change the interface of this. Should be OK since the feature was not yet in any tag.

andresailer:
There is an issue in the LCG nightly builds, hopefully fixed over the weekend.
I want to do some more tests, because the BoxOfStraws isn't behaving as I expected it to behave.

examples/ClientTests/CMakeLists.txt
```diff
@@ -632,4 +632,18 @@ if (DD4HEP_USE_GEANT4)
       REGEX_PASS "ResourcesAfterConstruction       ConstructSD:   VmRSS"
       REGEX_FAIL "Error;ERROR; Exception"
   )
+
+  dd4hep_add_test_reg(ClientTests_ddsim_setup_BoxOfStraws
```
andresailer, line 636: The test should only run when DDG4 and LCIO are used for building.
Also currently gets no hits. Not sure if shooting in a bad direction, or SD not working.

andresailer:
Will merge this, so it shows up in the nightlies, and fix the things that still need fixing in next PR.

Issue #1322, opened 2024-09-05
materialScan on CAD-imported geometry : strange results...


danieljeans:
### Check duplicate issues.

- [X] Checked for duplicates

### Goal

list materials along a line in detector model which including CAD imports, using the materialScan utility.

### Operating System and Version

ubuntu22

### compiler

gcc11.4.0

### ROOT Version

6.32.04

### DD4hep Version

/cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh -r 2024-09-04

### Reproducer

Using the model defined in 
https://github.com/aciarma/k4geo/tree/commonMDI/FCCee/MDI/compact/MDI_o1_v01/MDI_standalone_o1_v01.xml
which includes a beampipe from a CAD geometry. (n.b. You need to download the stl geometry files from [https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/MDI/MDI_o1_v01/](https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/MDI/MDI_o1_v01/) )

For simplicity I commented out all detector components except  
`<include ref="Beampipe_CADimport_o1_v02.xml"/>`
from the file MDI_standalone_o1_v01.xml

first I scan the geometry using a geantino, using ddsim:

```
ddsim --compactFile FCCee/MDI/compact/MDI_o1_v01/MDI_standalone_o1_v01.xml --enableG4Gun --runType shell
Idle> /gun/particle geantino
Idle> /tracking/verbose 1   
Idle> /gun/number 1         
Idle> /run/beamOn  
```
I see something reasonable: a beampipe made of various layers, with inner radius ~10mm:

```
*********************************************************************************************************
* G4Track Information:   Particle = geantino,   Track ID = 1,   Parent ID = 0
*********************************************************************************************************

Step#    X(mm)    Y(mm)    Z(mm) KinE(MeV)  dE(MeV) StepLeng TrackLeng  NextVolume ProcName
    0        0        0        0     1e+03        0        0         0 world_volume_1 initStep
    1     9.99        0        0     1e+03        0     9.99      9.99 AV_9!Gold_STL_2#2!"Solid1<stl_0#0 Transportation
    2       10        0        0     1e+03        0    0.005        10 AV_7!Beampipe_STL_0#0!"Componente1"_0#0 Transportation
    3     10.4        0        0     1e+03        0     0.45      10.4 AV_10!Paraffin_STL_3#3!"Solid2<stl_0#0 Transportation
[...]

PR #1324, opened 2024-09-17
Add warning if TGeo materialScan hits TGeoTessellated volumes. See gi…

MarkusFrankATcernch:
BEGINRELEASENOTES
Add warning if TGeo materialScan hits TGeoTessellated volumes. 

The DDRec MaterialManager and hence the corresponding MaterialScan use TGeo tracking through volumes.
Tracking in TGeo is not implemented for tessellated shapes as they result from CAD shapes. Though TGeo tracking is used
in the material scanner.... In TGeo tessellated surfaces behave like their bounding box. Hence they have no real shape 
and for this reason you only see a "box". 
TGeo tracking for tessellated shapes will only be implemented once TGeo moves to VecGeom. This will still take some time.

See github issue https://github.com/AIDASoft/DD4hep/issues/1322 for details about the discussion.
ENDRELEASENOTES

### DIFF
--- a/DDRec/src/MaterialScan.cpp
+++ b/DDRec/src/MaterialScan.cpp
@@ -22,6 +22,7 @@
 #include <DD4hep/DD4hepUnits.h>
 #include <DD4hep/Detector.h>
 #include <DD4hep/Printout.h>
+#include <DD4hep/Shapes.h>
 
 /// C/C++ include files
 #include <cstdio>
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
@andresailer Hi Andre, is there some issue with /dev3 ?
It cannot be me honestly....

andresailer:
> @andresailer Hi Andre, is there some issue with /dev3 ? It cannot be me honestly....

Yes @MarkusFrankATcernch this looks like an issue with dev3

PR #1326, opened 2024-09-30
Fix xercesc parsing of environment variables in the xml path

dudarboh:
BEGINRELEASENOTES

- Fix xercesc parsing of environment variables in the xml path https://github.com/iLCSoft/Marlin/issues/60

ENDRELEASENOTES

See https://github.com/iLCSoft/Marlin/issues/60 for the discussion.

This hard fixes it for me. But there is probably a cleaner way of doing it...
I am not an expert, so please cross-check.

### DIFF
--- a/DDCore/src/XML/DocumentHandler.cpp
+++ b/DDCore/src/XML/DocumentHandler.cpp
@@ -328,13 +328,20 @@ Document DocumentHandler::load(Handle_t base, const XMLCh* fname, UriReader* rea
 
 /// Load XML file and parse it using URI resolver to read data.
 Document DocumentHandler::load(const std::string& fname, UriReader* reader) const   {
+  auto _clean_fname = [](const std::string& s) -> std::string {
+    const std::string& temp = getEnviron(s);
+    std::string temp2 = undressed_file_name(temp.empty() ? s : temp);
+    if ( strncmp(temp2.c_str(),"file:",5) == 0 ) return temp2.substr(5);
+    return temp2;
+  };
+  auto fname_clean = _clean_fname(fname);
   std::string path;
-  printout(DEBUG,"DocumentHandler","+++ Loading document URI: %s",fname.c_str());
+  printout(DEBUG,"DocumentHandler","+++ Loading document URI: %s",fname_clean.c_str());
   try  {
-    size_t idx = fname.find(':');
-    size_t idq = fname.find('/');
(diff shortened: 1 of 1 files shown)


DDCore/src/XML/DocumentHandler.cpp
```diff
@@ -328,13 +328,20 @@ Document DocumentHandler::load(Handle_t base, const XMLCh* fname, UriReader* rea
 
 /// Load XML file and parse it using URI resolver to read data.
 Document DocumentHandler::load(const std::string& fname, UriReader* reader) const   {
+  auto _clean_fname = [](const std::string& s) -> std::string {
+    const std::string& temp = getEnviron(s);
+    std::string temp2 = undressed_file_name(temp.empty() ? s : temp);
+    if ( strncmp(temp2.c_str(),"file:",5) == 0 ) return temp2.substr(5);
+    return temp2;
+  };
+  auto fname_clean = _clean_fname(fname);
```
tmadlener, line 337:
```suggestion
  const auto fname_clean = [](const std::string& s) -> std::string {
    const std::string& temp = getEnviron(s);
    std::string temp2 = undressed_file_name(temp.empty() ? s : temp);
    if ( strncmp(temp2.c_str(),"file:",5) == 0 ) return temp2.substr(5);
    return temp2;
  }();
```
We don't need to give the lambda a name in this case, we can also invoke it immediately. The whole thing could probably do with a brief comment describing in english what it does / is supposed to do.

andresailer: just use the _clean_fname function that already exists in this file?

dudarboh:
I moved the function to the top of the file... Seems to work...

andresailer:
I just want to have @MarkusFrankATcernch comment on this, in case there is some side-effect from resolving environment variables here that could clash with other variables we resolve.

MarkusFrankATcernch:
Hmm..... This environment resolution mechanism is always a bit tricky and very often one runs into contradicting requirements.
Formally this all looks OK to me.
If all tests work I think this is fine. If people start to complain we can have a look at real use cases.

MarkusFrankATcernch (approved):
The environment resolution mechanism is always a bit tricky and very often one runs into contradicting requirements.
Formally this all looks OK to me.
If all tests work I think this is fine. If people start to complain we can have a look at real use cases.

PR #1144, opened 2023-07-14
DDSim: Change Default Output config

andresailer:
BEGINRELEASENOTES
- DDSim: change the default output file name on how DD4hep was built. Fixes #1101 
   - Add outputConfig.forceLCIO, outputConfig.forceEDM4HEP, outputConfig.forceDD4HEP switches to force a particular output plugin
   - If LCIO was used, slcio is still the default
   - If EDM4hep was used but not LCIO, EDM4hep is the default and used for all root output files. 
   - If neither LCIO or EDM4hep was used, then DD4hep's root output is used. Use the outputConfig.forceDD4HEP flag to use DD4hep's root output if EDM4hep is available as well

- DDSim: remove python2 legacy shims
- Geant4UIManager: fail when pre or post run commands are not successfully executed.

ENDRELEASENOTES

PR #1329, opened 2024-10-11
Edm4hep format as default when both edm4hep and lcio enabled

BrieucF:
BEGINRELEASENOTES
- ddsim: Set `edm4hep` the default output format in case DD4hep is compiled with both `DDD4HEP_USE_LCIO` `DD4HEP_USE_EDM4HEP` ON
- ddsim: Change the name of the default output file from dummyOutput to ddsimOutput

ENDRELEASENOTES

### DIFF
--- a/DDG4/python/DDSim/Helper/OutputConfig.py
+++ b/DDG4/python/DDSim/Helper/OutputConfig.py
@@ -12,9 +12,9 @@ DD4HEP_USE_EDM4HEP = "@DD4HEP_USE_EDM4HEP@" != "OFF"
 
 
 def defaultOutputFile():
-  if DD4HEP_USE_LCIO:
-    return "dummyOutput.slcio"
-  return "dummyOutput.root"
+  if DD4HEP_USE_LCIO and not DD4HEP_USE_EDM4HEP:
+    return "ddsimOutput.slcio"
+  return "ddsimOutput.root"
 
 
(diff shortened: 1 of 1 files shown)


tmadlener:
Will this actually yield EDM4hep files and not the DD4hep format without the `edm4hep.root` suffix?

@gaede anything speaking against this? I think all of the worklows, tutorials, etc. get an LCIO output in any case, because we specify an output file name with `.slcio` in any case.

BrieucF:
> Will this actually yield EDM4hep files and not the DD4hep format without the `edm4hep.root` suffix?
> 
> @gaede anything speaking against this? I think all of the worklows, tutorials, etc. get an LCIO output in any case, because we specify an output file name with `.slcio` in any case.

This PR does not change the behavior in case both the `lcio` and `edm4hep` flags are OFF (the dd4hep format is the default for rootfiles). If `edm4hep` is ON, rootfile will be edm4hep (without the need of edm4hep.root in the name), but this was introduced in https://github.com/AIDASoft/DD4hep/pull/1144.

tmadlener (approved):
> If `edm4hep` is ON, rootfiles will be edm4hep 

Thanks for the clarification. I have either missed that entirely or forgotten again in the meatime.

All good from my side then.

BrieucF:
I could not relate any of the failing test to this PR...

PR #1331, opened 2024-10-16
conditions: use static unsigned long long int because ROOTs latest ll…

andresailer:
…vm otherwise causes assertion in debug mode about not being able to represent in 64 bit. 
(Not sure if bug in enum parsing???)

Otherwise this error shows up when ROOT is in debug mode 
```
python -c "import DDG4"
python: /build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/interpreter/llvm-project/llvm/include/llvm/ADT/APSInt.h:99: int64_t llvm::APSInt::getExtValue() const: Assertion `isRepresentableByInt64() && "Too many bits for int64_t"' failed.
```

BEGINRELEASENOTES
- Conditions: change FIRST_KEY, LAST_KEY to constexpr instead of anonymous enum

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/Conditions.h
+++ b/DDCore/include/DD4hep/Conditions.h
@@ -101,10 +101,7 @@ namespace dd4hep {
       LAST_DET_KEY        = ~0x0U
     };
     /// Flags to indicate global conditions ranges
-    enum {
-      FIRST_KEY           =  0x0ULL,
-      LAST_KEY            = ~0x0ULL        
-    };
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch (approved):
This change from the C++ side should not be noticeable, since the enum is anonymous.
Go along with it.

Maybe we should start to replace the enums consistently with "static constexpr ...."?
I do not know.

andresailer:
Thanks for the review @MarkusFrankATcernch !

I think, only change things if we have to!

PR #1332, opened 2024-10-17
test_import: change error to fehler to avoid triggering test failure …

andresailer:
…because we get a warning from gosam

BEGINRELEASENOTES
- ImportTest: change 'error' to 'fehler' to avoid triggering test failures for spurious warning from gosam package in lcg stacks

ENDRELEASENOTES


andresailer:
At least a couple tests less failed, and dev3 is green.

PR #1333, opened 2024-10-17
Use edm4hep::labels::CellIDEncoding instead of "CellIDEncoding"

jmcarcell:
BEGINRELEASENOTES
- Use `edm4hep::labels::CellIDEncoding` instead of "CellIDEncoding"
- Use `podio::collMetadataParamName` instead of hardcoding the convention of adding `__`

ENDRELEASENOTES

See https://github.com/key4hep/EDM4hep/blob/22e85d79eca00bbd4eb479a536f90b86e76ea726/include/edm4hep/Constants.h#L29. This is the only place I could find that is related to EDM4hep.

### DIFF
--- a/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
+++ b/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
@@ -25,6 +25,7 @@
 #include <edm4hep/CaloHitContributionCollection.h>
 #include <edm4hep/SimCalorimeterHitCollection.h>
 #include <edm4hep/EDM4hepVersion.h>
+#include <edm4hep/Constants.h>
 /// podio include files
 #include <podio/CollectionBase.h>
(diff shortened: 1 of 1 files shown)


andresailer (approved):

wdconinc:
EIC would like to request that EDM4hep [features](https://github.com/key4hep/EDM4hep/commit/f1b2d827fcea6546173f3e9295395262b815f3c9) that are not in an actual release (only in a pre-release 0.99 and 0.99.1) do not get merged into DD4hep master without backwards compatibility in place. Too many headaches are coming our way from key4hep, where it seems the criterion for merging backwards-incompatible changes seems to be that it's ok as long as someone can get the right cherry-picks to work in a key4hep stack.

MarkusFrankATcernch:
@wdconinc Please see issue https://github.com/AIDASoft/DD4hep/issues/1379

PR #1334, opened 2024-10-17
Ci fixes

andresailer:
BEGINRELEASENOTES
- checkOverlaps: move from DDG4 to DDCore as this does not depend on Geant4, no impact on users
- checkGeometry: move from DDG4 to DDCore as this does not depend on Geant4, no impact on users
- CI: Key4hep: set CMAKE_INSTALL_LIBDIR=lib because Qt6 sets GNUInstallDir

ENDRELEASENOTES


andresailer:
Alright, down to one CAD Duck test error, which seems to come from a new version of assimp in key4hep.

PR #1337, opened 2024-10-18
CMake: explicitly ask for ROOTEve component

andresailer:
ROOT no longer finds nlohmann_json unless we ask for the ROOTEve component, so we ask for the ROOTEve component if the ROOTEve component exists.  DDEve and one executable in UtilityApps depend on this

cf. https://github.com/root-project/root/pull/16703

BEGINRELEASENOTES
- CMake: explicitly ask for ROOTEve component if it exists so we potentially find the non-builtin nlohmann_json

ENDRELEASENOTES


andresailer:
PS: this is PR 1337 :tada:

PR #1349, opened 2024-11-07
PythonPlugin: replace deprecated/removed Eval with Exec

andresailer:
It doesn't matter because we don't care about the result

BEGINRELEASENOTES
- PythonPlugin: replace deprecated Eval with Exec (root 6.36 need)

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/python/PythonPlugin.cpp
+++ b/DDCore/src/python/PythonPlugin.cpp
@@ -87,7 +87,8 @@ namespace  {
           result = TPython::Exec(c.second.c_str());
           break;
         case 'c':
-          TPython::Eval(c.second.c_str());
+          // we do not care about the result
+          TPython::Exec(c.second.c_str());
           result = kTRUE;
           break;
         case 'p':


MarkusFrankATcernch (commented):
Hi Andre, conditional compilation is probably better:
```
#include <RVersion.h>
....
#if ROOT_VERSION_CODE < ROOT_VERSION(6,35,1)
          TPython::Eval(c.second.c_str());
#else
          TPython::Exec(c.second.c_str());
#endif

```

andresailer:
The Exec exists and we use it in the different case, so this will always compile.

MarkusFrankATcernch:
I was doing the same changes yesterday.
Normally there is a subtle difference between Eval and Exec:
- `Eval` is in the same variable space (locals, globals)
- `Exec` normally is in a fresh variable space and only the global variables are presented.
Though, at least for me yesterday all tests worked, but I wonder whether on the ROOT side 
this was ever though of....

andresailer:
Admittedly I just took the advice in the ROOT doc at face value to just use Exec instead of eval. But since we didn't look at the returned value before, it all boils down to

https://github.com/root-project/root/blob/v6-34-00-patches/bindings/tpython/src/TPython.cxx#L465

```
  PyObject *pyObjectResult =
      PyRun_String(const_cast<char *>(command.str().c_str()), Py_file_input, gMainDict, gMainDict);
```
and

```
PyObject *result = PyRun_String(const_cast<char *>(expr), Py_eval_input, gMainDict, gMainDict);
```

Which is equivalent. The rest of the functions are just treating the output, are they not?

And if we don't compile conditionally we make this change dependent on the DD4hep version, and not the ROOT version behind it. Which is maybe better to spot errors if things turn out to be different in the end?

MarkusFrankATcernch:
I think it is OK. There is anyhow no other choice.

andresailer:
Hi @smuzzafar,

Any preference?

smuzaffar:
I have no preference. If `TPython::Exec` works for root 6.30 and above then I am fine with `TPython::Exec`

PR #1371, opened 2024-12-13
Read edm4hep

### LINKED ISSUE #1369: Create EDM4hep reader for DDG4

andresailer:
BEGINRELEASENOTES
- DDSim: add a reader for EDM4hep files, fixes #1369 
  - DDSim can now read EDM4hep files in root format.

ENDRELEASENOTES

- [x] Actually ingest parameters
   - ~~[x] Also ingest GeneratorInfo~~ (to be done later #1373)
- [x] Test with complicated events


DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,320 @@
...
+      } catch(std::invalid_argument&) {
+        //we ignore if we do not have runs information
+      }
+      try {
+        podio::Frame metaFrame = m_reader.readEntry("metadata", 0);
+        parameters->ingestParameters(metaFrame.getParameters());
+      } catch (std::runtime_error& e) {
+        // we ignore if we do not have metadata information
+      } catch(std::invalid_argument&) {
+        //we ignore if we do not have runs information
```
andresailer, line 146:
```suggestion
        //we ignore if we do not have metadata information
```

DDG4/lcio/LCIOEventReader.cpp
```diff
@@ -121,7 +121,10 @@ LCIOEventReader::readParticles(int event_number,
     int genStatus = mcp->getGeneratorStatus();
     // Copy raw generator status
     p->genStatus = genStatus&G4PARTICLE_GEN_STATUS_MASK;
-    m_inputAction->setGeneratorStatus(genStatus, status);
+    if(m_inputAction) {
+      // in some events we do not set up the inputAction
```
andresailer, line 125:
```suggestion
      // in some circumstances we do not set up the inputAction
```

andresailer:
That GeneratorInfo is not yet written to EDM4hep files?

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    // build collection of MCParticles
+    for(int i=0; i<NHEP; ++i )   {
+      const auto& mcp = primaries.at(i);
+      Geant4ParticleHandle p(new Particle(i));
+      const auto mom   = mcp.getMomentum();
+      const auto vsx   = mcp.getVertex();
+      const auto vex   = mcp.getEndpoint();
+      const auto spin  = mcp.getSpin();
+      const auto color = mcp.getColorFlow();
```
andresailer, line 235: Slated for removal. https://github.com/key4hep/EDM4hep/pull/389

tmadlener: I think since we plan to remove it in any case from EDM4hep, it doesn't make too much sense to add it here in the first place.

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,320 @@
...
+#include <podio/Frame.h>
+#include <podio/ROOTReader.h>
+
+typedef dd4hep::detail::ReferenceBitMask<int> PropertyMask;
+
+
+namespace dd4hep::sim {
+
+  // we use the index of the objectID to identify particles
+  // we will not support MCParticles from different collections
```
tmadlener, line 43: I suppose the same constraint is also in place for LCIO (or in general for various readers)? I think it's pretty reasonable, but want to make sure that this is documented somwhere (maybe it already is).

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,320 @@
...
+  using MCPARTICLE_MAP=std::map<int, int>;
+
+  /// get the parameters from the GenericParameters of the input EDM4hep frame and store them in the EventParameters
+  /// extension
+  template <class T=podio::GenericParameters> void EventParameters::ingestParameters(T const& source) {
+    auto const& intKeys = source.template getKeys<int>();
+    for(auto const& key: intKeys) {
```
tmadlener, line 50:
```suggestion
    for(auto const& key: source.template getKeys<int>()) {
```
Lifetime extension should work in this context, I think. (Similar for other types below)

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+  /// Class to read EDM4hep files
+  /**
+   *  \version 1.0
+   *  \ingroup DD4HEP_SIMULATION
+   */
+  class EDM4hepFileReader : public Geant4EventReader {
+  protected:
+    /// Reference to reader object
+    podio::ROOTReader m_reader {};
```
tmadlener, line 98:
```suggestion
    podio::Reader m_reader {};
```
Would immediately enable support for more backends (most importantly RNTuple, to a lesser extent SIO), without having to do that later.

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,320 @@
...
+  };
+
+  /// Initializing constructor
+  dd4hep::sim::EDM4hepFileReader::EDM4hepFileReader(const std::string& nam)
+    : Geant4EventReader(nam)
+    , m_collectionName("MCParticles")
+    , m_eventHeaderCollectionName("EventHeader")
+  {
+    printout(INFO,"EDM4hepFileReader","Created file reader. Try to open input %s",nam.c_str());
+    m_reader.openFile(nam);
```
tmadlener, line 125:
```suggestion
    m_reader = podio::makeReader(nam);
```
to complement using the generic reader interface.

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: removed colorflow.

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: Since once can only set one MCParticle collection name at the moment, this is implicitly the case. It would be a feature request to read more than one MCParticle collection.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -460,14 +460,24 @@ void Geant4Output2EDM4hep::saveEvent(OutputContext<G4Event>& ctxt)  {
...
   header.setWeight(eventWeight);
   //not implemented in EDM4hep ?  header.setDetectorName(context()->detectorDescription().header().name());
-  header.setTimeStamp( std::time(nullptr) ) ;
-  m_frame.put( std::move(header_collection), "EventHeader");
+  header.setTimeStamp(std::time(nullptr));
+
+  // extract event header, in case we come from edm4hep input
+  auto* meh = context()->event().extension<edm4hep::MutableEventHeader>(false);
+  if(meh) {
+    header.setTimeStamp(meh->getTimeStamp());
```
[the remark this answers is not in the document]
andresailer: There is just the nullptr timestamp in line 469

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -460,14 +460,24 @@ void Geant4Output2EDM4hep::saveEvent(OutputContext<G4Event>& ctxt)  {
...
+    for (auto const& weight: meh->getWeights()) {
+      header.addToWeights(weight);
+    }
```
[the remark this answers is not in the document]
andresailer: we just copy over.

DDG4/edm4hep/EDM4hepFileReader.cpp
tmadlener: Ah right, in that case I think this is "documented" enough.

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,320 @@
...
+    m_directAccess = true;
+  }
+
+  void EDM4hepFileReader::registerRunParameters() {
```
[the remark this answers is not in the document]
andresailer: Indeed it would do that.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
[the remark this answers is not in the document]
tmadlener: :+1: I misunderstood this whole thing here. We are copying over the contents from the input to the EventHeader that we are creating here. 

Just to be sure, I understood the rest correctly: The event & run number as well as the nominal event weight are "copied" by ingesting them in the reading side and then setting them again here on the output side.

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: This comment here was really for implementation reasons, since podio::ObjectID cannot be used as a map key due to missing operator< (<=?) so I went the lazy route.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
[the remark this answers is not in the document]
tmadlener: thanks for the explanation. Makes sense to me.

I think the EventHeader should not acquire much more information. It might be that we would need to follow such an approach with whatever we come up with for the generator info. But we can do that when we solve #1373

DDG4/edm4hep/EDM4hepFileReader.cpp
tmadlener: Today is your lucky day https://github.com/AIDASoft/podio/releases/tag/v01-02 ;)

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: What is the file extension convention for sio files?

tmadlener: `"sio"` (as per [this check](https://github.com/AIDASoft/podio/blob/d3f03813d7bf50f87c3aa4f32e619cebeb060113/src/Reader.cc#L72)

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: Done (to be pushed soon).

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: I implemented the switch to Reader, which compiles except for a linking issue, which I hope will be solved in the new podio version. I reverted the commit with the implementation I have and can re-apply it when podio was updated in dev3 and it really works.

tmadlener: https://github.com/AIDASoft/DD4hep/blob/abb85cbacd3909ea54733f2defd8053eaee63d12/DDG4/CMakeLists.txt#L108

`podio::podioRootIO` -> `podio::podioIO` might fix the linking issue (the `Reader` is not part of the root I/O for podio).

DDG4/edm4hep/EDM4hepFileReader.cpp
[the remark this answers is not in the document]
tmadlener: Is there a way to keep them separated? E.g. by attaching  the *metadata* frame to some entity as an extension and then writing it out again as a whole in the output if it is present?

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: Added sio extension

DDG4/edm4hep/EDM4hepFileReader.cpp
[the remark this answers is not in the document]
andresailer: Separated in the latest commits.

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,328 @@
...
+  {
+    printout(INFO,"EDM4hepFileReader","Created file reader. Try to open input %s",nam.c_str());
+    m_directAccess = true;
+  }
+
+  void EDM4hepFileReader::registerRunParameters() {
+    try {
+      auto *parameters = new RunParameters();
+      try {
+        podio::Frame runFrame = m_reader.readFrame("runs", 0);
```
tmadlener, line 139: I don't know if it's worth it here (because that would imply a podio version bump or some more pre-processor version handling), but in principle we have the `podio::Category::{Run,Event,Metadata}` constants in podio (available from 0.17 onwards) that would allow to replace some of the string constants in various places.

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,328 @@
...
+        podio::Frame metaFrame = m_reader.readFrame("metadata", 0);
+        fileParameters->ingestParameters(metaFrame.getParameters());
+      } catch (std::runtime_error& e) {
+        // we ignore if we do not have metadata information
+      } catch(std::invalid_argument&) {
+        // we ignore if we do not have metadata information
+      }
+      context()->run().addExtension<FileParameters>(fileParameters);
+    } catch(std::exception &e) {
+      printout(ERROR,"EDM4hepFileReader::registerRunParameters","Failed to register run parameters: %s", e.what());
```
tmadlener, line 159:
```suggestion
      printout(ERROR,"EDM4hepFileReader::registerRunParameters","Failed to register run or file parameters: %s", e.what());
```
Unless *run* is referring to the DD4hep run and not the EDM4hep / podio run(?).

DDG4/python/DDSim/DD4hepSimulation.py
```diff
@@ -45,11 +45,17 @@
     ".hepmc3", ".hepmc3.gz", ".hepmc3.xz", ".hepmc3.bz2",
     ".hepmc3.tree.root",
     ]
+EDM4HEP_INPUT_EXTENSIONS = [
+    ".root",
```
tmadlener, line 49: Wondering whether we should add a comment here, that these have to be checked **after** the HepMC3 extensions to make sure that we do not actually try to open a HepMC3 root file with the EDM4hep reader, if someone has the idea to refactor the `DD4hepSimulation` below. Or at least we should make sure that the tests catch that (which might be the case already).

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: Not worth it IMNSHO.

DDG4/python/DDSim/DD4hepSimulation.py
andresailer: Not here, but when we try to figure out the inputAction, I think?

andresailer: Comment coming in the next commit.

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: I split the catch between file and run parameters and updated the error message.

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,332 @@
...
+      p->vsz          = vsx[2]*CLHEP::mm;
+      p->vex          = vex[0]*CLHEP::mm;
+      p->vey          = vex[1]*CLHEP::mm;
+      p->vez          = vex[2]*CLHEP::mm;
+      p->process      = 0;
+      p->spin[0]      = spin[0];
+      p->spin[1]      = spin[1];
+      p->spin[2]      = spin[2];
+      p->colorFlow[0] = 0;
+      p->colorFlow[1] = 0;
```
tmadlener, line 267: We need to drop the `colorFlow` here as https://github.com/key4hep/EDM4hep/pull/389 has been merged. Additionally `spin` will be changed as well, but we can get to that once I have had time to do that.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -157,6 +158,27 @@ namespace dd4hep {
...
+        printout(DEBUG, "Geant4OutputEDM4hep", "Saving meta parameter: %s", p.first.c_str());
+        frame.putParameter(p.first, p.second);
+      }
+#if PODIO_BUILD_VERSION > PODIO_VERSION(0, 16, 2)
+      // This functionality is only present in podio > 0.16.2
+      for (auto const& p: this->dblParameters()) {
+        printout(DEBUG, "Geant4OutputEDM4hep", "Saving meta parameter: %s", p.first.c_str());
+        frame.putParameter(p.first, p.second);
+      }
 #endif
```
tmadlener, line 182: This has become unnecessary, since

https://github.com/AIDASoft/DD4hep/blob/9b9d43a0f2b1ec78e8c15d964fcae863c210921f/CMakeLists.txt#L166-L170

DDG4/edm4hep/EDM4hepFileReader.cpp
tmadlener: Right, too many things with the same members flying around here. Sorry got confused.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
tmadlener: Yes, exactly. I was referring to the pre-processor check here not the actual code.

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+  /// Class to read EDM4hep files
+  /**
+   *  \version 1.0
+   *  \ingroup DD4HEP_SIMULATION
+   */
+  class EDM4hepFileReader : public Geant4EventReader {
+  protected:
+    /// Reference to reader object
+    podio::Reader m_reader;
```
tmadlener, line 104: Looking at CI results, suggesting to use the `Reader` interface here was not entirely sensible, as that would mean we need at least podio 1.0. Sorry, I didn't think that one entirely through.

I am not sure whether we should simply bump the minimum podio version to 1.0. I suppose that is a bit easier to accommodate for EIC than a bump in EDM4hep minimum version @wdconinc ?

Otherwise we would either need to go to quite some pre-processor gymnastics or we could dial this back to only read the ROOT format for now (which would require only a few changes, as far as I can judge right now).

wdconinc: I guess I'll be pinged on every backwards compatibility decision now :-)

We're already using podio-1.1 now (but edm4hep-0.10.5), so we'd be able to accommodated a minimum podio version of 1.0.

tmadlener: Well, at least the ones where I am aware of a potential issue ;) Also once we are in post 1.0 land a breaking change has a much clearer association with a version number than it does below that, so I hope this will be among the last times we have to ask.

Thanks for the confirmation.

andresailer: Lets enable the reader, when podio>1 ?

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -467,14 +485,24 @@ void Geant4Output2EDM4hep::saveEvent(OutputContext<G4Event>& ctxt)  {
...
-  m_frame.put( std::move(header_collection), "EventHeader");
+  header.setTimeStamp(std::time(nullptr));
+
+  // extract event header, in case we come from edm4hep input
+  auto* meh = context()->event().extension<edm4hep::MutableEventHeader>(false);
+  if(meh) {
+    header.setTimeStamp(meh->getTimeStamp());
+    for (auto const& weight: meh->getWeights()) {
+      header.addToWeights(weight);
+    }
```
tmadlener, line 502:
```suggestion
#if EDM4HEP_BUILD_VERSION >= EDM4HEP_VERSION(0, 99, 0)
    for (auto const& weight: meh->getWeights()) {
      header.addToWeights(weight);
    }
#endif
```

To not undo the efforts from #1380 again immediately ;)

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,309 @@
...
+      }
+      return (*ip).second;
+    }
+  }
+
+  /// Read an event and fill a vector of MCParticles
+  EDM4hepFileReader::EventReaderStatus
+  EDM4hepFileReader::readParticles(int event_number, Vertices& vertices, std::vector<Particle*>& particles) {
+    m_currEvent = event_number;
+    podio::Frame frame = m_reader.readFrame("events", event_number);
```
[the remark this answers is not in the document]
andresailer: I prefer to keep readFrame, because it rhymes better with the frame reading of the other categories.

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,332 @@
...
+    int NHEP = primaries.size();
+    // check if there is at least one particle
+    if ( NHEP == 0 ) {
+      printout(WARNING,"EDM4hepFileReader", "We have no particles");
+      return EVENT_READER_NO_PRIMARIES;
+    }
+
+    MCPARTICLE_MAP mcparts;
+    std::vector<int>  mcpcoll;
+    mcpcoll.resize(NHEP);
```
jmcarcell, line 233:
```suggestion
    mcpcoll.reserve(NHEP);
```
`resize` will make all the elements be 0, which is not needed because all of them are written to below

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,332 @@
...
+    /// Name of the MCParticle collection to read
+    std::string m_collectionName;
+    /// Name of the EventHeader collection to read
+    std::string m_eventHeaderCollectionName;
+
+  public:
+    /// Initializing constructor
+    EDM4hepFileReader(const std::string& nam);
+
+    /// Read parameters set for this action
```
jmcarcell, line 114:
```suggestion
    /// Set parameters for this action
```

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -0,0 +1,332 @@
...
+      return EVENT_READER_NO_PRIMARIES;
+    }
+
+    MCPARTICLE_MAP mcparts;
+    std::vector<int>  mcpcoll;
+    mcpcoll.resize(NHEP);
+    for(int i=0; i<NHEP; ++i ) {
+      edm4hep::MCParticle p = primaries.at(i);
+      mcparts[p.getObjectID().index] = i;
+      mcpcoll[i] = p.getObjectID().index;
```
jmcarcell, line 237:
```suggestion
      mcpcoll.push_back(p.getObjectID().index);
```
With the comment above this has to change

andresailer: This looks more clear to what is happening, and certainly isn't the bottle neck.

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: Will keep as is. (could consider creating the vector at correct size...)

DDG4/CMakeLists.txt
```diff
@@ -104,12 +104,19 @@ ENDIF()
...
   set_target_properties(DDG4EDM4HEP PROPERTIES VERSION ${DD4hep_VERSION} SOVERSION ${DD4hep_SOVERSION})
-
+  if(TARGET(podio::podioIO))
+      dd4hep_add_plugin(DDG4EDM4HEPReader
+        SOURCES edm4hep/EDM4hepFileReader.cpp
+        USES    DD4hep::DDG4 EDM4HEP::edm4hep EDM4HEP::edm4hepDict podio::podio podio::podioDict podio::podioIO
+      )
+      install(TARGETS DDG4EDM4HEPReader EXPORT DD4hep LIBRARY DESTINATION lib)
+      set_target_properties(DDG4EDM4HEPReader PROPERTIES VERSION ${DD4hep_VERSION} SOVERSION ${DD4hep_SOVERSION})
+    endif()
```
tmadlener, line 119: I think a warning or at least a status message would be nice here too, to explain why the EDM4hep input reader is not built. That might avoid some confusion and could also be used to provide the solution along side it.

andresailer: Thanks! Also the IF check was wrong.

DDG4/CMakeLists.txt
```diff
@@ -104,12 +104,22 @@ ENDIF()
...
+  if(TARGET podio::podioIO)
+      dd4hep_add_plugin(DDG4EDM4HEPReader
+        SOURCES edm4hep/EDM4hepFileReader.cpp
+        USES    DD4hep::DDG4 EDM4HEP::edm4hep EDM4HEP::edm4hepDict podio::podio podio::podioDict podio::podioIO
+      )
+      install(TARGETS DDG4EDM4HEPReader EXPORT DD4hep LIBRARY DESTINATION lib)
+      set_target_properties(DDG4EDM4HEPReader PROPERTIES VERSION ${DD4hep_VERSION} SOVERSION ${DD4hep_SOVERSION})
+      MESSAGE(STATUS "EDM4hepFileReader enabled")
+  else()
+      MESSAGE(STATUS "EDM4hepFileReader requires a new version of EDM4hep, disabled")
```
tmadlener, line 121:
```suggestion
      MESSAGE(STATUS "EDM4hepFileReader requires at least PODIO 1.0, disabled")
```
as the `Reader` is part of podio and unrelated to EDM4hep.

tmadlener (approved):
Looks good from my side, now.

PR #1377, opened 2025-01-06
Use canonical variables for the installation locations of libraries and binaries

andriish:
Use canonical variables for the installation locations of libraries and binaries

BEGINRELEASENOTES
- Use CMAKE_INSTALL_LIBDIR and CMAKE_INSTALL_BINDIR to install libraries and binaries

ENDRELEASENOTES


MarkusFrankATcernch:
Happy New Year to all!

@andresailer 
Do you agree with these changes ?

MarkusFrankATcernch:
@andriish 
Thanks for fixing. Let's see with Andre tomorrow, when he is back.

andresailer:
>  Do you agree with these changes ?

I won't answer until I see that all the tests are passing. But this has the potential to also break other peoples installation procedures.

Issue #1379, opened 2025-01-07
Discussion: Evolution of edm4hep and dd4hep


MarkusFrankATcernch:
This issue is a result from a valid discussion item brought up by Wouter in merge request https://github.com/AIDASoft/DD4hep/pull/1333

@jmcarcell @tmadlener @andresailer @wdconinc 
The request mentioned by Wouter in https://github.com/AIDASoft/DD4hep/pull/1333 is a valid request. 
However, we then need to define some upgrade path:

- Either dd4hep disconnects from edm4hep - edm4hep is only an add-on to dd4hep or
- Edm4hep developers can sustain backwards compatibility long-term or
- We package the dd4hep-edm4hep modules as a separate package so that versions of (edm4hep, dd4hep) can evolve more freely.

Points up for discussion

tmadlener:
I would like to add another potential "solution" to the discussion and also something that should have been done with #1333 already: We need to make sure that we properly "communicate" the version of EDM4hep that we need. Currently there is only a `find_package(EDM4HEP REQUIRED)`, which arguably should be `find_package(EDM4HEP 0.99 REQUIRED)`.

Arguably this only shifts the discussion a bit, but 

> where it seems the criterion for merging backwards-incompatible changes seems to be that it's ok as long as someone can get the right cherry-picks to work in a key4hep stack.

seems a bit harsh, given that this was likely a mistake in (not) updating version dependencies (combined with the fact that DD4heps CI only builds on top of rather recent EDM4hep versions).

For the actual discussion: Personally, I would be in favor of

> - Edm4hep developers can sustain backwards compatibility long-term or

We will try to make sure that we don't break things for EDM4hep 0.10.X for now (and also make sure that is reflected in the cmake config). For that we would probably need to setup a dedicated build that ensures that via CI. I am not entirely sure how to handle things once there is an EDM4hep 1.0. I personally, would have a preference of making that a hard requirement of DD4hep and removing any special casing for pre-release versions. However, this is not the strongest opinion and as long as we have some CI setup that checks compatibility with the minimum EDM4hep version, I don't have a problem with keeping some pre-processor handling for that.

wdconinc:
I think the underlying issue here has multiple components:
[...]

PR #1380, opened 2025-01-07
Restore compatibility with EDM4hep 0.10

jmcarcell:
BEGINRELEASENOTES
- Revert some of the changes of https://github.com/AIDASoft/DD4hep/pull/1375 to build a vector of floats instead of doubles since the necessary constructor doesn't exist in 0.10.X.
- Use `edm4hep::labels::CellIDEncoding` only when they are available (starting at version 0.99)

ENDRELEASENOTES

See https://github.com/AIDASoft/DD4hep/pull/1333#issuecomment-2573680487

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -13,6 +13,7 @@ jobs:
               "LCG_102/x86_64-centos9-gcc11-opt",    # g4 11
               "LCG_104/x86_64-el9-gcc13-opt",         # g4 11
               "LCG_105/x86_64-ubuntu2204-gcc11-opt",  # g4 11
+              "LCG_106/x86_64-el9-gcc13-opt",  # g4 11
               "dev3/x86_64-el9-gcc13-opt",
               "dev3/x86_64-el9-clang16-opt",          # g4 11
               "dev4/x86_64-el9-gcc13-opt"]
@@ -66,6 +67,10 @@ jobs:
           if [[ ${{ matrix.LCG }} =~ dev3 ]]; then
             echo "::group::CMakeConfig 2"
             cmake -DDD4HEP_HEPMC3_COMPRESSION_SUPPORT=ON ..
+          fi
+          if [[ ${{ matrix.LCG }} =~ 106 ]]; then
+            echo "::group::CMakeConfig Enable EDM4hep"
+            cmake -DDD4HEP_USE_EDM4HEP=ON ..
           fi
           echo "::group::Compile"
           ninja install
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -68,6 +68,7 @@ jobs:
             echo "::group::CMakeConfig 2"
             cmake -DDD4HEP_HEPMC3_COMPRESSION_SUPPORT=ON ..
           fi
+          # Make sure DD4hep can be built with an older version of EDM4hep (0.10.5 for LCG_106)
           if [[ ${{ matrix.LCG }} =~ 106 ]]; then
             echo "::group::CMakeConfig Enable EDM4hep"
             cmake -DDD4HEP_USE_EDM4HEP=ON ..
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -69,7 +69,7 @@ jobs:
             cmake -DDD4HEP_HEPMC3_COMPRESSION_SUPPORT=ON ..
           fi
           # Make sure DD4hep can be built with an older version of EDM4hep (0.10.5 for LCG_106)
-          if [[ ${{ matrix.LCG }} =~ 106 ]]; then
+          if [[ ${{ matrix.LCG }} =~ 106 ||${{ matrix.LCG }} =~ 104 ]]; then
             echo "::group::CMakeConfig Enable EDM4hep"
             cmake -DDD4HEP_USE_EDM4HEP=ON ..
           fi
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -69,7 +69,7 @@ jobs:
             cmake -DDD4HEP_HEPMC3_COMPRESSION_SUPPORT=ON ..
           fi
           # Make sure DD4hep can be built with an older version of EDM4hep (0.10.5 for LCG_106)
-          if [[ ${{ matrix.LCG }} =~ 106 ||${{ matrix.LCG }} =~ 104 ]]; then
+          if [[ ${{ matrix.LCG }} =~ 106 || ${{ matrix.LCG }} =~ 104 ]]; then
             echo "::group::CMakeConfig Enable EDM4hep"
             cmake -DDD4HEP_USE_EDM4HEP=ON ..
           fi
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -68,9 +68,8 @@ jobs:
             echo "::group::CMakeConfig 2"
             cmake -DDD4HEP_HEPMC3_COMPRESSION_SUPPORT=ON ..
           fi
-          # Make sure DD4hep can be built with an older version of EDM4hep,
-          # 0.10.5 for LCG_106 and 0.10.0 for LCG_104
-          if [[ ${{ matrix.LCG }} =~ 106 || ${{ matrix.LCG }} =~ 104 ]]; then
+          # Make sure DD4hep can be built with an older version of EDM4hep, 0.10.5 for LCG_106
+          if [[ ${{ matrix.LCG }} =~ 106 ]]; then
             echo "::group::CMakeConfig Enable EDM4hep"
             cmake -DDD4HEP_USE_EDM4HEP=ON ..
           fi
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -68,7 +68,8 @@ jobs:
             echo "::group::CMakeConfig 2"
             cmake -DDD4HEP_HEPMC3_COMPRESSION_SUPPORT=ON ..
           fi
-          # Make sure DD4hep can be built with an older version of EDM4hep (0.10.5 for LCG_106)
+          # Make sure DD4hep can be built with an older version of EDM4hep,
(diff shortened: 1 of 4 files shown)


DDDigi/io/DigiIO.cpp
```diff
@@ -171,8 +171,8 @@ namespace dd4hep {
       mcp.setCharge(3.0*p.charge);
       mcp.setVertex( _toVectorD(p.start_position) );
       mcp.setEndpoint( _toVectorD(p.end_position) );
-      mcp.setMomentum( _toVectorD(p.momentum) );
-      mcp.setMomentumAtEndpoint( _toVectorD(p.momentum) );
+      mcp.setMomentum( _toVectorF(p.momentum) );
+      mcp.setMomentumAtEndpoint( _toVectorF(p.momentum) );
```
tmadlener, line 175: I would assume that we run into [this deprecation warning](https://github.com/key4hep/EDM4hep/blob/285a6f04a47937b8769dd4495100f7c5507b5dcd/edm4hep.yaml#L71-L72) here with this construct? Similar below.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -26,6 +26,11 @@
 #include <edm4hep/SimCalorimeterHitCollection.h>
 #include <edm4hep/EDM4hepVersion.h>
 #include <edm4hep/Constants.h>
+#if EDM4HEP_VERSION_MAJOR == 0 && EDM4HEP_VERSION_MINOR < 99
```
tmadlener, line 29:
```suggestion
#if EDM4HEP_BUILD_VERSION < EDM4HEP_VERSION(0, 99, 0)
```

tmadlener:
Given the discussion in #1379. How hard would it be to setup a workflow that builds an older version of EDM4hep on the fly to make sure things still actually work?

andresailer:
> related to the deprecation of a constructor 

Can you add why these are reverted please!

andresailer:
> Given the discussion in #1379. How hard would it be to setup a workflow that builds an older version of EDM4hep on the fly to make sure things still actually work?

Which old version of EDM4hep should we test against? Maybe it exists in one of the LCG releases.

tmadlener:
> Which old version of EDM4hep should we test against?

Letting @wdconinc confirm, but I think 0.10.5 is what we should be aiming for.

tmadlener:
Can we also put a the `0.10.5` (or whatever we settle on for CI) into the `find_package(EDM4HEP)` call in the appropriate `CMakeLists.txt`?

wdconinc:
> Can we also put a the `0.10.5` (or whatever we settle on for CI) into the `find_package(EDM4HEP)` call in the appropriate `CMakeLists.txt`?

Or, in the interest of forward compatibility, the same structure that's used for podio to allow both 0.99 and 1.0?

DDDigi/io/DigiIO.cpp
jmcarcell: Yes

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
jmcarcell: Done

.github/workflows/linux.yml
```diff
@@ -13,6 +13,7 @@ jobs:
               "LCG_102/x86_64-centos9-gcc11-opt",    # g4 11
               "LCG_104/x86_64-el9-gcc13-opt",         # g4 11
               "LCG_105/x86_64-ubuntu2204-gcc11-opt",  # g4 11
+              "LCG_106/x86_64-el9-gcc13-opt",  # g4 11
```
tmadlener, line 16:
```suggestion
              "LCG_104/x86_64-el9-gcc13-opt",  # g4 11
```
There we would actually check EDM4hep 0.10 instead of 0.10.5

DDDigi/io/DigiIO.cpp
tmadlener: If we want to avoid pre-processor special casing we could change the definition of `_toVectorF` to

```cpp
template<typename T = edm4hep::Vector3f>
T _toVectorF(const dd4hep::Position& ep) {
      using FT = decltype(T::x);
      return { FT(ep.x()), FT(ep.y()), FT(ep.z()) };
} 
```
and then use it here as
```cpp
mcp.setMomentum( _toVectorF<(decltype(mcp.getMomentum())>(p.momentum) );
```

But a pre-processor `#if` definitely sticks out a bit more for a potential future cleanup.

.github/workflows/linux.yml
andresailer: But EIC is using 0.10.5 so that makes more sense? I understood Wouter's comment to build whatever latest patch version of 0.10.X if we were to build EDM4hep on the fly.

Otherwise 104 is already in the list above, and we could use that and just change the IF to enable EDM4hep.

tmadlener: > However, it should probably be 0.10, and not rely on a patch release.

I was coming from this angle and the fact that we have `find_package(EDM4HEP 0.10.0 REQUIRED)` in cmake. I think the main point is that the `find_package` and at least one CI check should be consistent for the actual minimal version.

andresailer: But I am coming from the same angle? :confused: 

It is anyway good to add 106 based CI. We can additionally enable EDM4hep for 104.

DDDigi/io/DigiIO.cpp
jmcarcell: Went for the preprocessor check to be able to grep `EDM4HEP_VERSION`

.github/workflows/linux.yml
```diff
@@ -68,6 +68,7 @@ jobs:
             echo "::group::CMakeConfig 2"
             cmake -DDD4HEP_HEPMC3_COMPRESSION_SUPPORT=ON ..
           fi
+          # Make sure DD4hep can be built with an older version of EDM4hep (0.10.5 for LCG_106)
           if [[ ${{ matrix.LCG }} =~ 106 ]]; then
```
tmadlener, line 72: We should also enable this for `104` (as discussed in https://github.com/AIDASoft/DD4hep/pull/1380#discussion_r1906861531)

tmadlener:
It looks like DDDigi is also built with LCG_102 and LCG_101, but that doesn't yet have a recent enough version of EDM4hep.

.github/workflows/linux.yml
jmcarcell: It doesn't work due to the version of `podio` being too old in 104. However I built locally with EDM4hep 0.10.0 and it is possible. I reverted the addition of LCG_104. If it's not possible to easily test against 0.10.0 then 0.10.5 is the best we can do :shrug:

jmcarcell: LCG_105 has 0.10.2 if we want to try that one

tmadlener: In that case we also have to adjust the minimal version of podio in the CMakelists:

https://github.com/AIDASoft/DD4hep/blob/c532581bc637fcedb32e1ecb71b6390328772f3b/CMakeLists.txt#L166

LCG_104 comes with 0.16.6 which is larger than the 0.16.3 we currently require.

jmcarcell:
This should be ready now.


> > Can we also put a the `0.10.5` (or whatever we settle on for CI) into the `find_package(EDM4HEP)` call in the appropriate `CMakeLists.txt`?
> 
> Or, in the interest of forward compatibility, the same structure that's used for podio to allow both 0.99 and 1.0?

I do not mind this, but maybe we can do it once we get closer to EDM4hep 1.0 since I'm not sure when it's going to happen.

andresailer:
Something in ROOT master broke

```
   	 62 - t_ClientTests_Check_Shape_PseudoTrap (Subprocess aborted)
  	 63 - t_ClientTests_Check_Shape_PseudoTrap1 (Subprocess aborted)
  	 64 - t_ClientTests_Check_Shape_PseudoTrap2 (Subprocess aborted)
```

andresailer (approved):
Since test failures are unrelated

CMakeLists.txt
```diff
@@ -160,10 +160,10 @@ if(DD4HEP_USE_LCIO)
 endif()
 
 if(DD4HEP_USE_EDM4HEP)
-  find_package(EDM4HEP REQUIRED)
+  find_package(EDM4HEP 0.10.0 REQUIRED)
```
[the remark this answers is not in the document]
andresailer: Then we should put 0.10.5.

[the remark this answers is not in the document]
jmcarcell: Done

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -25,7 +25,15 @@
 #include <edm4hep/CaloHitContributionCollection.h>
 #include <edm4hep/SimCalorimeterHitCollection.h>
 #include <edm4hep/EDM4hepVersion.h>
-#include <edm4hep/Constants.h>
+#if EDM4HEP_BUILD_VERSION <= EDM4HEP_VERSION(0, 10, 2)
+  constexpr const char* CellIDEncoding = "CellIDEncoding";
+#elif EDM4HEP_BUILD_VERSION < EDM4HEP_VERSION(0, 99, 0)
```
tmadlener, line 30:
```suggestion
#if EDM4HEP_BUILD_VERSION < EDM4HEP_VERSION(0, 99, 0)
```

As we don't need to support down to 0.10.2 now.

tmadlener (commented):
Minor nitpick from my side. We can also leave it in, it doesn't really hurt at this point.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
andresailer: I can take it out in my PR.

andresailer:
I would like to merge this after it is done on LCG_104, so that I can test the reader on LCG_104 as well.

PR #1383, opened 2025-01-10
Fix CI failures.

scott-snyder:
Trying to fix the observed CI failures for this repo.
There are several independent problems.  I could split this up if desired
(but then the individual ones would show CI failures).

  - For newer ROOT versions, we need to explicitly require ROOTEve.
    Otherwise, we get failed dependencies on nlohman_json.

  - TPython::Eval has been removed from newer versions of ROOT.
    (Use TPython::Exec instead.)

  - Recent versions of Gaudi provide a pytest plugin that adds DartMeasurement
    xml entities to the output.  This includes the source of the test function.
    But in this case, the function source contains a print statement
    containing the string ERROR, causing the end-of-test filtering
    to report the test as failed.

    This is arguably a Gaudi bug, as it really shouldn't be changing the
    behavior of tests in other packages.  Will try to follow up there.
    But for now, this behavior can be disabled with the environment
    setting DISBABLE_CTEST_MEASUREMENTS=1.

  - The RICH test_number_of_hits script requires that hte average
    number of hits be within +- 15 of the nominal value.  But with one
    of the platform variations, it is actually about 15.6 off.
    Increased the threshold to 16.

  - In the nightly key4hep build, including the Geant4 configuration
    implicitly (via including the qt6 configuration) sets
    CMAKE_INSTALL_LIBDIR to lib64.  This results in some libraries
    being installed to lib and some to lib64.  But the setup scripts
    put only the lib directory on the path, so we have tests failing
    due to not finding plugins that ended up in lib64.

  - assimp 5.2.0 included a change that altered the output for the
    duck test, so logic was added to switch to a new reference file version
    for newer versions of assump.  However, as of assump 5.4.3 (in the
    key4hep nightly), that change has been reverted.  Update the logic
    to go back to the original reference for assimp versions newer than 5.4.3.

BEGINRELEASENOTES
- Fix all observed CI test failures.
ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ dd4hep_configure_output( OUTPUT "${PROJECT_BINARY_DIR}" INSTALL "${CMAKE_INSTALL
 ########################
 
 # Configure ROOT
-find_package (ROOT 6.08 REQUIRED CONFIG)
+find_package (ROOT 6.08 COMPONENTS ROOTEve REQUIRED CONFIG)
 DD4HEP_SETUP_ROOT_TARGETS()
 
 # Configure BOOST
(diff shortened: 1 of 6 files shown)


scott-snyder:
Need to resolve conflicts.

PR #1384, opened 2025-01-11
Handle ganged segmentations.

scott-snyder:
Some segmentations can gang together regions from multiple volumes (such as the Allegro ECal).  In the calorimeter implementations of Geant4SensitiveAction, we were finding the hit position by first getting the local position from the segmentation, then converting to a global position using the transform obtained from the step. But if multiple volumes are ganged, this may not be correct: the volume which actually contains the hit may not be the same one which the segmentation used for its local coordinate system. In this case, we need to instead find the volID to use from the segmentation and get the transformation by looking it up in the volume manager.

However, not all segmentations properly implement the volumeID method needed for this, so this has to be opt-in.  We add an isGanged() method to the Segmentation interface.  If this is overridden with a method that returns true, then we proceed as described above. Otherwise, we preserve the present behavior.

Also factor out some duplicated code among the calorimeter implementations.

BEGINRELEASENOTES
- DD4hep::Segmentations: add cellsSpanVolumes function to mark segmentations that that can have cells that span multiple volumes, such as the Allegro ECal.

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DDSegmentation/Segmentation.h
+++ b/DDCore/include/DDSegmentation/Segmentation.h
@@ -127,9 +127,11 @@ namespace dd4hep {
           \return vector<double> in natural order of dimensions, e.g., dx/dy/dz, or dr/r*dPhi
       */
       virtual std::vector<double> cellDimensions(const CellID& cellID) const;
-      /// Return true if this segmentation can gang together regions
-      /// from multiple volumes.
-      virtual bool isGanged() const
+      /// Return true if this segmentation can have cells that span multiple
+      /// volumes.  That is, points from multiple distinct volumes may
+      /// be assigned to the same cell.
+      /// In that case, a working volumeID() implementation is required.
+      virtual bool cellsSpanVolumes() const
       {
         return false;
       }
--- a/DDCore/include/DDSegmentation/Segmentation.h
+++ b/DDCore/include/DDSegmentation/Segmentation.h
@@ -127,6 +127,12 @@ namespace dd4hep {
           \return vector<double> in natural order of dimensions, e.g., dx/dy/dz, or dr/r*dPhi
       */
       virtual std::vector<double> cellDimensions(const CellID& cellID) const;
+      /// Return true if this segmentation can gang together regions
+      /// from multiple volumes.
+      virtual bool isGanged() const
+      {
+        return false;
+      }
 
     protected:
       /// Default constructor used by derived classes passing the encoding string
--- a/DDG4/plugins/Geant4SDActions.cpp
+++ b/DDG4/plugins/Geant4SDActions.cpp
@@ -504,7 +504,7 @@ namespace dd4hep {
         std::cout << out.str();
         return true;
       }
-      handleCalorimeterHit (cell, contrib, *coll, h, *this, m_segmentation);
+      handleCalorimeterHit(cell, contrib, *coll, h, *this, m_segmentation);
       mark(h.track);
       return true;
     }
--- a/DDG4/plugins/Geant4SDActions.cpp
+++ b/DDG4/plugins/Geant4SDActions.cpp
@@ -45,7 +45,7 @@ namespace dd4hep {
 
           // Convert the position relative to the local readout volume
           // to a global position.
-          if (!segmentation.isGanged()) {
+          if (!segmentation.cellsSpanVolumes()) {
             global = h.localToGlobal(pos);
           }
           else {
--- a/DDG4/plugins/Geant4SDActions.cpp
+++ b/DDG4/plugins/Geant4SDActions.cpp
@@ -300,7 +300,7 @@ namespace dd4hep {
         return true;
       }
 
-      handleCalorimeterHit (cell, contrib, *coll, h, *this, m_segmentation);
+      handleCalorimeterHit(cell, contrib, *coll, h, *this, m_segmentation);
       mark(h.track);
       return true;
     }
--- a/DDG4/plugins/Geant4SDActions.cpp
+++ b/DDG4/plugins/Geant4SDActions.cpp
@@ -27,7 +27,57 @@ namespace dd4hep {
 
     namespace {
       struct Geant4VoidSensitive {};
+
+      /// Common code to handle the creation of a calorimeter hit.
+      template <class HANDLER>
+      void handleCalorimeterHit (VolumeID cell,
+                                 const HitContribution& contrib,
+                                 Geant4HitCollection& coll,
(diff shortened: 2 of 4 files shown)


andresailer:
I think "gang" sounds too much like slang (pun intended) to be clearly understood

DDG4/plugins/Geant4SDActions.cpp
```diff
@@ -27,7 +27,57 @@ namespace dd4hep {
...
+            // In this case, we can't use the transformation we get from
+            // the step --- the volume that actually contains the hit
+            // may not be the same volume that the segmentation uses
+            // for the local coordinate system.  We need to get the
+            // actual volID used from the segmentation and then look
+            // it up the volume manager to get the proper transformation.
+            VolumeID volID = segmentation.volumeID(cell);
+            VolumeManager vman = VolumeManager::getVolumeManager(sd.detectorDescription());
+            VolumeManagerContext* vc = vman.lookupContext(volID);
+            global = vc->localToWorld(Position(pos)) / dd4hep::mm;
```
andresailer, line 62: Why is the unit transformation needed here?

DDCore/include/DDSegmentation/Segmentation.h
```diff
@@ -127,6 +127,14 @@ namespace dd4hep {
           \return vector<double> in natural order of dimensions, e.g., dx/dy/dz, or dr/r*dPhi
       */
       virtual std::vector<double> cellDimensions(const CellID& cellID) const;
+      /// Return true if this segmentation can have cells that span multiple
+      /// volumes.  That is, points from multiple distinct volumes may
+      /// be assigned to the same cell.
+      /// In that case, a working volumeID() implementation is required.
+      virtual bool cellsSpanVolumes() const
```
andresailer, line 134: Thanks! That is a better name, and also a docstring that doesn't just use self reference to explain itself :)

DDG4/plugins/Geant4SDActions.cpp
```diff
@@ -250,25 +300,7 @@ namespace dd4hep {
...
-                c_name(),contrib.deposit,pos.X,pos.Y,pos.Z,handler.path().c_str(),
-                coll->GetName().c_str());
-        if ( 0 == hit->cellID )  { // for debugging only!
-          hit->cellID = cellID(step);
-          except("+++ Invalid CELL ID for hit!");
-        }
-      }
-      hit->truth.emplace_back(contrib);
-      hit->energyDeposit += contrib.deposit;
+      handleCalorimeterHit (cell, contrib, *coll, h, *this, m_segmentation);
```
andresailer, line 303:
```suggestion
      handleCalorimeterHit(cell, contrib, *coll, h, *this, m_segmentation);
```

DDG4/plugins/Geant4SDActions.cpp
```diff
@@ -462,23 +477,7 @@ namespace dd4hep {
...
-        printM2("CREATE hit with deposit:%e MeV  Pos:%8.2f %8.2f %8.2f  %s",
-                contrib.deposit,pos.X,pos.Y,pos.Z,handler.path().c_str());
-        if ( 0 == hit->cellID )  { // for debugging only!
-          hit->cellID = cellID(step);
-          except("+++ Invalid CELL ID for hit!");
-        }
-      }
-      hit->truth.emplace_back(contrib);
-      hit->energyDeposit += contrib.deposit;
+      handleCalorimeterHit (cell, contrib, *coll, h, *this, m_segmentation);
```
andresailer, line 480:
```suggestion
      handleCalorimeterHit(cell, contrib, *coll, h, *this, m_segmentation);
```

DDG4/plugins/Geant4SDActions.cpp
```diff
@@ -505,23 +504,7 @@ namespace dd4hep {
...
-        printM2("CREATE hit with deposit:%e MeV  Pos:%8.2f %8.2f %8.2f  %s",
-                contrib.deposit,pos.X,pos.Y,pos.Z,handler.path().c_str());
-        if ( 0 == hit->cellID )  { // for debugging only!
-          hit->cellID = cellID(h.touchable(), h.avgPositionG4());
-          except("+++ Invalid CELL ID for hit!");
-        }
-      }
-      hit->truth.emplace_back(contrib);
-      hit->energyDeposit += contrib.deposit;
+      handleCalorimeterHit (cell, contrib, *coll, h, *this, m_segmentation);
```
andresailer, line 507:
```suggestion
      handleCalorimeterHit(cell, contrib, *coll, h, *this, m_segmentation);
```

DDG4/plugins/Geant4SDActions.cpp
scott-snyder: The localToWorld() call returns something in dd4hep coordinates, but global is used to initialize a Geant4Calorimeter::Hit, which uses Geant4 /CLHEP units (cf Geant4OUtput2EDM4Hep.cpp, which divides the position by CLHEP::mm).

In the existing branch of the code

```
            global = h.localToGlobal(pos);
```

the method used here is

```
dd4hep::Position Geant4HitHandler::localToGlobal(const DDSegmentation::Vector3D& local)  const   {
  return localToGlobal(G4ThreeVector(local.X / dd4hep::mm,local.Y / dd4hep::mm,local.Z / dd4hep::mm));
}
```

which does the unit conversion.

andresailer: Thanks! Let's add a comment for this in the code, please!
```suggestion
            // explicit unit conversion; h.localToGlobal does it internally already
            global = vc->localToWorld(Position(pos)) / dd4hep::mm;
```

DDG4/plugins/Geant4SDActions.cpp
```diff
@@ -294,24 +327,7 @@ namespace dd4hep {
...
-                c_name(),contrib.deposit,pos.X,pos.Y,pos.Z,handler.path().c_str(),
-                coll->GetName().c_str());
-        if ( 0 == hit->cellID )  { // for debugging only!
-          hit->cellID = cellID(h.touchable(), h.avgPositionG4());
-          except("+++ Invalid CELL ID for hit!");
-        }
-      }
-      hit->truth.emplace_back(contrib);
-      hit->energyDeposit += contrib.deposit;
+      handleCalorimeterHit (cell, contrib, *coll, h, *this, m_segmentation);
```
andresailer, line 330:
```suggestion
      handleCalorimeterHit(cell, contrib, *coll, h, *this, m_segmentation);
```

PR #1386, opened 2025-01-12
Fix potential use of dangling temporary.

scott-snyder:
In

```
  const TGeoHMatrix& m = par.nominal().worldTransformation();
```
nominal() returns a temporary Alignment object by value, and worldTransformation() returns a reference within that Alignment object. So we shoud make a copy.

BEGINRELEASENOTES
- Fix potential use of dangling temporary in exampe.
ENDRELEASENOTES


MarkusFrankATcernch:
@scott-snyder Yes. You are completely correct. I guess the compiler generated a temporary copy from the stack rather than a reference (yes, compiler tend to be that clever!). That's why there were no SEGVs.
Thanks for spotting!

PR #1390, opened 2025-01-25
Optimize Geant4VolumeManager

MarkusFrankATcernch:
BEGINRELEASENOTES
-  Optimize the Geant4VolumeManager. Instead of mapping the Geant4 pathes as vectors of pointers,
    hash64 the pointers of the path and lookup pathes from the touchable history using this key.
-  Relevant changes are in the files DDG4/Geant4GeometryInfo.h and src/Geant4VolumeManager.cpp.
    Everything else is only cosmetics.

ENDRELEASENOTES

### DIFF
--- a/DDG4/include/DDG4/Geant4GeometryInfo.h
+++ b/DDG4/include/DDG4/Geant4GeometryInfo.h
@@ -19,10 +19,13 @@
 #include <DD4hep/GeoHandler.h>
 #include <DD4hep/PropertyTable.h>
 #include <DDG4/Geant4Primitives.h>
+#include <DDG4/Geant4TouchableHandler.h>
+
 
 // C/C++ include files
 #include <map>
 #include <vector>
+#include <cstdint>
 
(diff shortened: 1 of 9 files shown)


MarkusFrankATcernch:
This MR should optimize the memory usage of the `Geany4VolumeManager` in 
up to some degree should overcome the need to use https://github.com/AIDASoft/DD4hep/pull/1288.
The memory consumption of the `Geany4VolumeManager` should be largely reduced, because 
instead of performing lookups with vectors of pointers as key only 64bit hash values are used.

ianna:
Hi @MarkusFrankATcernch . Has this PR been included in a release? Version? Thanks!

andresailer:
Hi @ianna 
It is in 1.31 https://github.com/AIDASoft/DD4hep/blob/8545739348aa117de3ca020a70fabc4bc5ca60aa/doc/ReleaseNotes.md?plain=1#L6

Issue #1407, opened 2025-02-27
`ddsim` executes `DDG4.kernel` from a foreign installation


ybedfer:
Hello, I installed my private `DD4hep` in an environment where another installation already exists (in some `/opt/software` link).
The DDG4 kernel that is executed is that from that `/opt/software` instead of mine. This, only recently: in the past, where as far as I can tell, I was proceeding exactly the same, everything was fine.

What I am doing (and was also doing in the past):
 - `git clone` from my fork, compile and source the `thisdd4hep.sh` script.
 - My private `DD4hep` installation is then linked into  env. variables `PATH`, `PYTHONPATH`, `LD_LIBRARY_PATH` and `DD4HEP_LIBRARY_PATH`.
 - Execute `ddsim`. The program runs correctly and produces a valid output TTree. But it does not make use of my private `DD4hep`. I can tell that from the fact that I cannot see any of my printouts.
 - Also, when I execute `ddsim` w/in `gdb`, I can see that the DDG4 kernel that is executed is that from `/opt/software`. For example, I get the following trace:
```
#15 0x00007fffdb9c39c5 in dd4hep::sim::Geant4Exec::run(dd4hep::sim::Geant4Kernel&) ()
   from /opt/software/linux-debian12-x86_64_v2/gcc-12.2.0/dd4hep-1.30-cyaqyw2wpezcww3x5rxtqgzqpweevtx5/lib/libDDG4.so.1.30
#16 0x00007fffdb9f1cb9 in dd4hep::sim::Geant4Kernel::run() ()
   from /opt/software/linux-debian12-x86_64_v2/gcc-12.2.0/dd4hep-1.30-cyaqyw2wpezcww3x5rxtqgzqpweevtx5/lib/libDDG4.so.1.30
```
I also added a printout in `DD4hepSimulation.py` to output the path to the imported module (`os.path.dirname(DDG4.__file__)`. => It's correct and does point to my own installation.

What can be the reason? What do I miss?

andresailer:
Either your environments are not completely updated with respect to the second installation of dd4hep, or you are using some software that is still linked to the first installation of DD4hep, or both.

andresailer:
Ok, then I will assume that there is nothing else we can do on our side.

PR #1408, opened 2025-03-03
Update ROOT_LIBRARY_PATH in thisdd4hep.sh

veprbl:
BEGINRELEASENOTES
- `ROOT_LIBRARY_PATH` will now be updated by thisdd4hep.sh.
ENDRELEASENOTES

This is needed to instruct Cling to prefer a particular dd4hep path.


ybedfer:
I can add that this commit fixes my issue #1407. See "https://github.com/AIDASoft/DD4hep/issues/1407".

andresailer:
If we move the setting of ROOT_LIBRARY_PATH to
https://github.com/AIDASoft/DD4hep/blob/42956b19149612a335ffd85ba62345e53d953764/cmake/thisdd4hep.sh#L43-L53

also the packages depending on DD4hep benefit from this change directly

veprbl:
> When is this needed for cling? This would not affect the creation of the rootmaps during the compilation of DD4hep.

cppyy is using cling

andresailer (approved):

PR #1414, opened 2025-03-07
Fix the result of `innerMaterial()` and `outerMaterial()` depending on if the other is called first

### LINKED ISSUE #1413: The result of `innerMaterial` and `outerMaterial` depends on if the other function is called before
### Check duplicate issues.

- [x] Checked for duplicates

### Goal

When calling `outerMaterial()` after `innerMaterial()` has been called, the result is different for some surfaces than when `outerMaterial()` is called first. This is happening because the TGeoManager keeps some internal state that changes after `innerMaterial()` is called, leading to it not being the same when calling `outerMaterial()`. For CLD_o2_v07, it seems about 20 surfaces are affected, so it seems it's not a very large effect.

AlmaLinux 9

### compiler

GCC 14

### ROOT Version

6.32.08 and 6.34.04

### DD4hep Version

master

### Reproducer

``` bash
source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh
python issue.py
python issue.py --call-inner-material
```
If we now diff the two output files
```
diff surfaces_good.txt surfaces_bad.txt
```
We will see that in one file some surfaces are Silicon and in the other Air.

The file issue.py:
``` python
import os
import argparse
import dd4hep
import DDRec

argparser = argparse.ArgumentParser()
argparser.add_argument('--call-inner-material', action='store_true', default=False)

det = dd4hep.Detector.getInstance()
det.fromCompact(f'{os.getenv("K4GEO")}/FCCee/CLD/compact/CLD_o2_v07/CLD_o2_v07.xml')

args = argparser.parse_args()

sm = det.extension[DDRec.SurfaceManager]()
multimap = sm.map('world')
d = {}
for k, v in multimap:
    if args.call_inner_material:
        innermat = v.innerMaterial()
    outermat = v.outerMaterial()
    d[k] = outermat.name()

with open(f'surfaces_{"bad" if args.call_inner_material else "good"}.txt', 'w') as f:
    for k, v in d.items():
        f.write(f'{k} {v}\n')
```

### Additional context

_No response_

jmcarcell:
BEGINRELEASENOTES
- Fix the result of `innerMaterial()` and `outerMaterial()` depending on if the other is called first
- Add a test that fails without this fix and passes with the fix

ENDRELEASENOTES

Fix #1413.

I don't know exactly why this is happening for a few surfaces. What happens is that when getting the next boundary after `innerMaterial()` has been called is not the same as if it wasn't called (in this call https://github.com/AIDASoft/DD4hep/blob/24fcc3ebb73bbf77227a5a283fbc6268bbee7937/DDRec/src/MaterialManager.cpp#L77). Making a backup of the navigator and restoring it solves this problem.

### DIFF
--- a/DDRec/src/MaterialManager.cpp
+++ b/DDRec/src/MaterialManager.cpp
@@ -39,6 +39,9 @@ namespace dd4hep {
 
     const MaterialVec& MaterialManager::materialsBetween(const Vector3D& p0, const Vector3D& p1 , double epsilon) {
       if( ( p0 != _p0 ) || ( p1 != _p1 ) ) {	
+        // A backup is needed to restore the state of the navigator after the track is done
+        // see https://github.com/AIDASoft/DD4hep/issues/1413
+        _tgeoMgr->DoBackupState();
         //---------------------------------------	
         _mV.clear() ;
         _placeV.clear();
@@ -152,6 +155,8 @@ namespace dd4hep {
         _tgeoMgr->ClearTracks();
 
         _tgeoMgr->CleanGarbage();
+
+        _tgeoMgr->DoRestoreState();
 	
         //---------------------------------------	
 	
--- a/examples/ClientTests/CMakeLists.txt
+++ b/examples/ClientTests/CMakeLists.txt
@@ -668,5 +668,10 @@ if (DD4HEP_USE_GEANT4)
     REGEX_FAIL "Exception;EXCEPTION;ERROR;Error;FATAL"
   )
   #
-  #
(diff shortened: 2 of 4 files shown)


andresailer:
Thanks! That was a nice big piece of debugging!

andresailer:
@MarkusFrankATcernch 

We have a failing seed for the box of straws test here
```
  EventSeeder1     INFO  At beginEvent: eventID=0, runID=0 initialSeed=1960713805, newSeed=3837709546
```

gaede:
Excellent, well done.   This was around for years, I guess...

Issue #1418, opened 2025-03-13
Possible to build without Eve dependency


cgleggett:
Is it possible to build dd4hep without the libEve dependency from ROOT? Don't need to do any visualization, and would like to be able to build without requiring X11 libraries.

MarkusFrankATcernch:
Hi Charles,
The answer is: Most likely yes.
If you could even supply a PR with the open to disable DDeve it would even be more welcome!

Cheers, Markus

cgleggett:
hi Markus -

  I have something that does this now. Question for you - do you prefer an explicit build option like `-DBUILD_DDEVE=OFF` or just a check for ROOT::Eve, and if it's not found, will not build DDEve?

MarkusFrankATcernch:
Hi Charles,
I think best is both:
- If it is explicitly switched off:   -DBUILD_DDEVE=OFF, do not even try to build it.
- If nothing is switched off and a check for ROOT::Eve failed, print a warning an do not build it.
Does this sound reasonable ?

Cheers,
Markus

andresailer:
https://github.com/AIDASoft/DD4hep/blob/6ce741bf632ed250ddf424e9086c662bfbfa310e/CMakeLists.txt#L86

One should be able to remove DDEve from this list to disable building DDEve.

MarkusFrankATcernch:
@cgleggett Charles, please see here: https://github.com/AIDASoft/DD4hep/pull/1420

bedanar:
Hello! I am sorry. maybe I am writing somwhere wrong, but I have an error whem building DD4hep and I think it is connected to using Eve. Nearly a week ago I had no such problem when building DD4hep. ROOT  was coorectly built from source as it is stated in the instruction: https://root.cern/install/build_from_source/
Here how I build DD4hep:
```
git clone https://github.com/AIDASoft/DD4hep.git
cd DD4hep
mkdir build && cd build
cmake -DDD4HEP_USE_GEANT4=ON -DBoost_NO_BOOST_CMAKE=ON -DDD4HEP_USE_LCIO=ON -DBUILD_TESTING=ON -DDD4HEP_BUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release ..
make -j20
make install
cd ..
source bin/thisdd4hep.sh
```
The error happens at the 4th step:
```
CMake Error at DDEve/CMakeLists.txt:21 (target_link_libraries):
  The link interface of target "DDEve_Interface" contains:

[...]

PR #1420, opened 2025-03-17
Allow to selectively disable DD4HEP sub-packages

MarkusFrankATcernch:
BEGINRELEASENOTES
- Fix for issue https://github.com/AIDASoft/DD4hep/issues/1418.
  Individual DD4HEP sub-packages can be disable from the cmake command line:
  ```
  $> cmake <other options> -DDD4HEP_DISABLE_PACKAGES="DDRec DDEve DDDigi"  ../DD4hep
  ```
  This will disable the sub-packages DDRec,DDEve,DDDigi:
  ```
  $> cmake -DDD4HEP_DISABLE_PACKAGES="DDRec DDEve DDDigi"  ../DD4hep
       -- Could NOT find LATEX (missing: LATEX_COMPILER) 
       -- No LaTeX/Biber found, cannot compile user manual.
       -- Disable package: DDEve
       -- Disable package: DDDigi
       -- Disable package: DDRec
       -- Will be building these packages: DDDetectors;DDCond;DDAlign;DDCAD;DDG4;UtilityApps
       -- Including DD4hepBuild.cmake
  ```
   Clearly no prevision can be done concerning package dependencies. This is up to the user.
   For example `DDG4` cannot be disabled if `DDDigi` should be kept.
ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,28 +78,28 @@ option(DD4HEP_USE_EDM4HEP       "Build edm4hep extensions"                      
 option(DD4HEP_USE_HEPMC3        "Build hepmc3 extensions"                       OFF)
 option(DD4HEP_USE_TBB           "Build features that require TBB"               OFF)
 option(DD4HEP_LOAD_ASSIMP       "Download and build ASSIMP from github"         OFF)
-option(DD4HEP_DISABLE           "Selectively disable DD4HEP sub-packages"       OFF)
+option(DD4HEP_DISABLE_PACKAGES  "Selectively disable DD4HEP sub-packages [Separate multiples by space]"  OFF)
 option(BUILD_TESTING            "Enable and build tests"                        ON)
 option(BUILD_SHARED_LIBS        "If OFF build STATIC Libraries"                 ON)
 option(DD4HEP_SET_RPATH         "Link libraries with built-in RPATH (run-time search path)" ON)
 option(DD4HEP_RELAX_PYVER       "Do not require exact python version match with ROOT" OFF)
 #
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
> We use a space separated list for DD4HEP_BUILD_PACKAGES, can we use the same for DD4HEP_DISABLE to make these consistent? Can we rename to DD4HEP_DISABLE_PACKAGES?

Spaces are difficult: for cmake they need to appear as one single command/option for proper processing. The rest is simple.

andresailer:
Spaces just need to be quoted, no: `-DDD4HEP_DISABLE="DDEve DDigi DDRec"` ?

MarkusFrankATcernch:
> Spaces just need to be quoted, no: `-DDD4HEP_DISABLE="DDEve DDigi DDRec"` ?
I do not really care, but does this really look better ?
The spaces are a weak argument: in principle you want a list.....

andresailer:
I would like consistency with the existing option.

cgleggett:
hi Markus -

  You beat me to it ;-)

  However, the reason I wanted to not build DDEve is that it depends on ROOT::Eve, which depends on X11 libraries. Some minimal installations (especially on headless servers and HPCs) don't have the X11 libraries installed, and thus require a root build without X11 dependencies, which prevents the building of ROOT:Eve. I think we should add a check in UtilityApps for the presence of ROOT::Eve, and not build geoWebDisplay and teveDisplay of it's not found. If that make sense I'll submit a PR.

cgleggett:
My local tests don't show that there is a problem. Maybe this is from an older version of ROOT? I'll create a new PR and see if the CI finds an issue.

PR #1422, opened 2025-03-17
Add LANGUAGES CXX to the project command not to depend on a C compiler

jmcarcell:
Currently DD4hep depends on the C compiler when it's not needed. The changes in this PR disable the checks for a C compiler since `LANGUAGES NONE` or `LANGUAGES` defaults to both C and C++.

BEGINRELEASENOTES
- Remove checks for the C compiler when configuring
- Remove `ENABLE_LANGUAGE` which is redundant when called after `PROJECT`

ENDRELEASENOTES

What's different now? You won't see messages like
``` 
-- The C compiler identification is GNU 14.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
```
when running CMake.


MarkusFrankATcernch:
@jmcarcell Hi Juan Miguel,
Should I merge this one ?

CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
-PROJECT( DD4hep LANGUAGES NONE)
+PROJECT( DD4hep LANGUAGES CXX)
```
andresailer, line 2: The NONE is needed when we just want to build documentation?

jmcarcell: That is correct. It is later overwritten in https://github.com/AIDASoft/DD4hep/blob/f9ef7188119b272a5da02695a28bac7efecaa966/CMakeLists.txt#L120. Now it is back to NONE.

jmcarcell:
It should be ready now. I think the call of `ENABLE_CXX` has to be there because even though `project(... LANGUAGES CXX)` is called when setting the DD4hep version it happens in a subdirectory instead of at the top level and that doesn't seem to have the same effect. Locally, removing `ENABLE_CXX` works for me, but when I try with the Key4hep stack it doesn't.

PR #1423, opened 2025-03-17
[WIP] Check for ROOT::Eve before building geoWebDisplay and teveDislplay

cgleggett:
BEGINRELEASENOTES
For systems that don't have X11 libs, ROOT::Eve may not be available, so check before building utility apps that depend on it
ENDRELEASENOTES


andresailer:
Hi @cgleggett 

Could you please keep the RELEASENOTES format for your PR message?

andresailer:
Also note that `ROOT::ROOTEve` and `ROOT::Eve` are two different targets.

CMakeLists.txt
```diff
@@ -136,12 +136,6 @@ dd4hep_configure_output( OUTPUT "${PROJECT_BINARY_DIR}" INSTALL "${CMAKE_INSTALL
 # Configure ROOT
 find_package (ROOT 6.08 REQUIRED CONFIG)
 DD4HEP_SETUP_ROOT_TARGETS()
-# ROOT no longer finds nlohmann_json unless we ask for the ROOTEve component, so we ask
-# for the ROOTEve component if the ROOTEve component exists.
-# DDEve and one executable in UtilityApps depend on this
-if(TARGET ROOT::ROOTEve)
```
andresailer, line 142: Please just leave this here.

In your build of ROOT there shouldn't be any ROOT::ROOTEve, so it won't hurt your use case.

UtilityApps/CMakeLists.txt
```diff
@@ -42,10 +42,13 @@ if(TARGET Geant4::Interface)
 endif()
 
 # the web display tool depends on the ROOT7 version of DDEve
-if(TARGET ROOT::ROOTEve)
+find_package (ROOT COMPONENTS ROOTEve)
+if(ROOTEve_FOUND)
```
andresailer, line 46: I don't see ROOTEve_FOUND being defined in the ROOTConfig.cmake.
Just rely on the existence of the TARGET please!

UtilityApps/CMakeLists.txt
```diff
@@ -81,8 +84,13 @@ if (DD4HEP_USE_LCIO)
 endif()
 
 # #-----------------------------------------------------------------------------------
-add_executable(teveDisplay src/teve_display.cpp src/next_event_dummy.cpp G__eve.cxx)
-target_link_libraries(teveDisplay DD4hep::DDRec ROOT::Core ROOT::Eve ROOT::Gui ROOT::Graf3d ROOT::RGL )
+if(ROOTEve_FOUND)
```
andresailer, line 87:
```suggestion
if(TARGET ROOT::Eve)
```

andresailer (changes requested):

cgleggett:
> @cgleggett You ran exactly into the problem which was noted in the main CMakeLists.txt:
> 
> ```
>   -- 
>   -- Configuring done (22.5s)
>   CMake Error at /cvmfs/sft-nightlies.cern.ch/lcg/views/dev3/Tue/x86_64-el9-clang16-opt/cmake/ROOTConfig-targets.cmake:805 (set_target_properties):
>     The link interface of target "ROOT::ROOTEve" contains:
>   
>       nlohmann_json::nlohmann_json

This is very interesting - I was building locally against cmvfs 106c (root 6.32.10) and my own root 6.34.04, and it works fine for both of those. dev3/Tue uses 6.35.01, so it's an issue with very recent versions of root. I'll ask the root team about this.

UtilityApps/CMakeLists.txt
cgleggett: This is my bad - the usual CMake style should have defined ROOTEve_FOUND after a find_package(ROOTEve) call, but looks like it's not working here.

andresailer:
> I'll ask the root team about this.

As noted in https://github.com/AIDASoft/DD4hep/pull/1337 this comes from https://github.com/root-project/root/pull/16703

MarkusFrankATcernch:
@andresailer 
Yes. Please ask.  
nlohmann_json::nlohmann_json is very useful- actually one of the best json parsers.
I would not like to loose it.
Getting it from ROOT for DD4hep is very convenient, since ROOT is now everywhere.

cgleggett:
I've somehow managed to mess up my repo after a rebase. I'm going to close this PR and open a new one.

andresailer:
>  Getting it from ROOT for DD4hep is very convenient, since ROOT is now everywhere.

We are (usually) not getting it from ROOT though, because we don't use the ROOT-builtin nlohmann-json in LCG stacks.

PR #1426, opened 2025-03-18
Only build components that need ROOT::Eve if it's available

cgleggett:
If ROOT is built without X11 support, ROOT::Eve is not built. This PR prevents the building of UtilityApp executable that depend on Root::Eve if it can't be found. Note that ROOT::ROOTEve does not require X11 libs.

BEGINRELEASENOTES
- Suppress building of UtilityApp components that require ROOT::Eve if it's not available (eg if ROOT was built without X11 support)
ENDRELEASENOTES

### DIFF
--- a/UtilityApps/CMakeLists.txt
+++ b/UtilityApps/CMakeLists.txt
@@ -69,7 +69,7 @@ dd4hep_add_dictionary( G__eve
   )
 
 # #-----------------------------------------------------------------------------------
-if (DD4HEP_USE_LCIO)
+if (DD4HEP_USE_LCIO AND TARGET ROOT:Eve)
   dd4hep_add_dictionary( G__eve1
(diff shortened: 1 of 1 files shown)


UtilityApps/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ dd4hep_add_dictionary( G__eve
   )
 
 # #-----------------------------------------------------------------------------------
-if (DD4HEP_USE_LCIO)
+if (DD4HEP_USE_LCIO AND TARGET ROOT:Eve)
```
jmcarcell, line 72:
```suggestion
if (DD4HEP_USE_LCIO AND TARGET ROOT::Eve)
```

cgleggett: thanks for noticing this! fixed.

PR #1427, opened 2025-03-19
Remove C++ preprocessor checks for a standard before C++14

jmcarcell:
BEGINRELEASENOTES
- Remove C++ preprocessor checks for a standard before C++14

ENDRELEASENOTES

Since at least 14 is required in the top level CMakeLists.txt


MarkusFrankATcernch:
Are there really clients still using C++11 ?

jmcarcell:
Probably not since it is not allowed: https://github.com/AIDASoft/DD4hep/blob/6c3e19e0157c6e5be2785e5452d53f95fd37dbe3/CMakeLists.txt#L58

andresailer (approved):
Looks good to me!

PR #1438, opened 2025-03-27
make timeout configurable in g4GraphicalScan

### LINKED ISSUE #1435: Make g4GraphicalScan internal timeouts configurable
Hi

I was running the new tool g4GraphicalScan with a new detector and I noticed that the pilot scan to check consistency of geometry has an internal timeout of 90s. The material scan runs have a time out of 600s. 

I was wondering if it would be possible to make the timeout the same in both cases (pilot scan and actual scan runs) and configurable. 

The detector I am trying to scan takes more than 90s to initialize when the run starts, and a pilot run cannot be executed. 

Thank you for your time.

atolosadelgado:
BEGINRELEASENOTES
- g4GraphicalScan: add option `-t` `--timeout` to make timeout configurable, fixes #1435 

ENDRELEASENOTES

I have marked this as WIP because I have not tested it yet.

### DIFF
--- a/DDG4/python/bin/g4GraphicalScan.py
+++ b/DDG4/python/bin/g4GraphicalScan.py
@@ -280,7 +280,7 @@ if not noPilot:
     for cc in cmd:
         print(cc, end=' ')
     print('\n')
-    pilotresult = subprocess.run(cmd, capture_output=True, text=True, timeout=int(opt.timeOutValue))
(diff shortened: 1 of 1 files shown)


atolosadelgado:
it is ready for review :)

andresailer:
Looks good to me! Thanks @atolosadelgado 
I let @danieljeans comment as well.

danieljeans:
a good idea! thanks @atolosadelgado

atolosadelgado:
the straw tube test seems to fail randomly :)

PR #1439, opened 2025-04-01
Fix, Optical physics option invokeSD require poststep info only

atolosadelgado:
BEGINRELEASENOTES
- Modification of `Geant4Sensitive::cellID` base class function to handle properly the cases when the step corresponds to an optical photon and the option InvokeSD is enable, so only post-step information is used to calculate the volume/cell ID.

ENDRELEASENOTES

Hi,

I have tested it with ARC, and now it seems to work without errors. Is there any optical physics test that may cover the changes introduced in this PR? Also, is there any test that performs some kind of benchmarking? I am concerned about worsening the overall performance

Please let me know if you think this fix can be done in a better way.

If you think the fix is too specific for optical physics, alternatively we can patch the optical-related sensitive detector actions. I have tried with the optical tracker SD action, and it works.

### DIFF
--- a/DDG4/src/Geant4SensDetAction.cpp
+++ b/DDG4/src/Geant4SensDetAction.cpp
@@ -31,6 +31,9 @@
 // C/C++ include files
 #include <stdexcept>
 
+#include "G4OpticalParameters.hh"
+#include "G4OpticalPhoton.hh"
+
 #ifdef DD4HEP_USE_GEANT4_UNITS
 #define MM_2_CM 1.0
 #else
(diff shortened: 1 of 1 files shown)


DDG4/src/Geant4SensDetAction.cpp
```diff
@@ -260,11 +263,16 @@ long long int Geant4Sensitive::volumeID(const G4VTouchable* touchable) {
 long long int Geant4Sensitive::cellID(const G4Step* step) {
   Geant4StepHandler h(step);
   Geant4VolumeManager volMgr = Geant4Mapping::instance().volumeManager();
-  VolumeID volID = volMgr.volumeID(h.preTouchable());
+  bool OpticsInvokeSD = G4OpticalParameters::Instance()?
+                        G4OpticalParameters::Instance()->GetBoundaryInvokeSD() : false;
+  bool IsOpticalPhoton = step->GetTrack()->GetDefinition() == G4OpticalPhoton::Definition();
+  bool UsePostStepOnly = IsOpticalPhoton && OpticsInvokeSD;
```
andresailer, line 269:
```suggestion
  bool UsePostStepOnly = G4OpticalParameters::Instance() && G4OpticalParameters::Instance()->GetBoundaryInvokeSD() && (step->GetTrack()->GetDefinition() == G4OpticalPhoton::Definition());
```
If you are worried about performance?

MarkusFrankATcernch:
@atolosadelgado Is this OK? Should I merge?

atolosadelgado:
It is ready from my side, I do not know if you have further comments. Maybe it is worth to mention this feature in the documentation as well?

MarkusFrankATcernch:
@atolosadelgado I think there is no real documentation for this sensitive detector.

atolosadelgado:
@MarkusFrankATcernch but this modifies the base class method that calculates the cell ID and the volume ID, it is common to all sensitive detectors :)

PR #1441, opened 2025-04-04
DDCore: drop RootHistDraw, no longer exists

andresailer:
Target was dropped https://github.com/root-project/root/pull/18191
I want to see if we need it for older versions of ROOT.

BEGINRELEASENOTES
- CMake: drop use of ROOT:ROOTHistDraw

ENDRELEASENOTES


andresailer:
We don't need it at all it seems.

PR #1444, opened 2025-04-15
Add ccache to CI workflows

tmadlener:
BEGINRELEASENOTES
- CI: Switch to `run-lcg-view` action `v5` and inject `ccache` as compiler launcher via cmake to speed up CI builds.

ENDRELEASENOTES


andresailer:
Thanks!

PR #1445, opened 2025-04-27
feat: allow event readers with pre-existing (Event,Run)Parameters

wdconinc:
In EIC we use a run action plugin to write additional information to the RunParameters. Due to the sequence in which the plugins are loaded, this only works when the event reader is loaded first since the current event reader fails when the (Event,Run)Parameters already exist. 

This PR makes the event readers work whether a (Event,Run)Parameters already exists or not. If the (Event,Run)Parameters already exist, they are added to.

BEGINRELEASENOTES
- DDG4: Allow event readers to work with pre-existing (Event,Run)Parameters, written by other plugins.

ENDRELEASENOTES

### DIFF
--- a/DDG4/edm4hep/EDM4hepFileReader.cpp
+++ b/DDG4/edm4hep/EDM4hepFileReader.cpp
@@ -148,7 +148,6 @@ namespace dd4hep::sim {
       } catch(std::invalid_argument&) {
         // we ignore if we do not have runs information
       }
-      context()->run().addExtension<RunParameters>(parameters);
     } catch(std::exception &e) {
       printout(ERROR,"EDM4hepFileReader::registerRunParameters","Failed to register run parameters: %s", e.what());
     }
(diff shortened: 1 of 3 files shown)


DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -134,7 +134,12 @@ namespace dd4hep::sim {
 
   void EDM4hepFileReader::registerRunParameters() {
     try {
-      auto *parameters = new RunParameters();
+      // get RunParameters or create new if not existent yet
+      auto* parameters = context()->run().extension<RunParameters>(false);
+      if (!parameters) {
+        parameters = new RunParameters();
+        context()->run().addExtension<RunParameters>(parameters);
```
andresailer, line 141: Line 151 should be dropped then, right?

andresailer (changes requested):
That one line needs to go. Otherwise looks good. Thanks!

andresailer (approved):

PR #1447, opened 2025-05-07
Geant4InputAction: add Alternative Stable Statuses to allow marking p…

### LINKED ISSUE #1446: proceed primary particle with custom generator status code in ddsim
To test the reconstruction performance of the EIC-ePIC detector under realistic environment, we combine several signal and beam background hepmc files to one merged hepmc file, and pass it to dd4sim for simulation.

By default, the particles from both signal and background files share the same generator status code (e.g. 1, 2...) in the merged file, therefore we don't know if a reconstructed event is associated with signal or background in later analysis.  A convenient way to deal with this would be modifying the status code for background particles (e.g. 1 --> 201, 2--> 202). However, those 200+ code won't proceed by dd4sim.

I presented this problem in the key4hep discussion on[ May 6th ](https://indico.cern.ch/event/1537262/#3-how-to-mark-particles-from-p). @andresailer suggested us to take use of the `alternativeDecayStatuses` option as implemented in https://github.com/AIDASoft/DD4hep/pull/1260.

To test this method, I prepared a sample hepmc3 file by replacing all generator status 1 with 201, ..., then ran with `--physics.alternativeDecayStatuses 201`.  ddsim complained:
```PrimaryHandler   INFO  +++++ G4PrimaryVertex at (+6.63e+00,-3.03e+01,+3.45e+02) [mm] +1.85e+03 [ns]
PrimaryHandler   INFO  +++++ G4PrimaryVertex at (+3.10e+01,-9.28e-01,+5.83e+02) [mm] +1.89e+03 [ns]
ParticleHandler  INFO  +++ Event 8 Begin event action. Access event related information.
EDM4hepOutput    WARN  +++ [Event:8] No valid MC truth info present. Is a Particle handler installed ?
Geant4Output2EDM4hep INFO  +++ Saving EDM4hep event 8 run -1.
GenerationInit   INFO  +++ Initializing event 10. Within run:0 event 10.
HEPMC3FileReader::moveToEvent INFO  Skipping the first 9 events
```

According to this line of code https://github.com/andresailer/DD4hep/blob/dc2485bf273d921754f7394bc5babedc839a1b53/DDG4/src/Geant4InputAction.cpp#L296, 

`--physics.alternativeDecayStatuses` sets a given status code to `G4PARTICLE_GEN_DECAYED` which is equivalent to status code 2, which won't make the particle go through dd4sim like a primary stable particle.

Is it possible to add a similar option to set
`  else if ( m_alternativePrimaryStatuses.count(genStatus) ) status.set(G4PARTICLE_GEN_STABLE);` ?

andresailer:
…articles as stable with statuses other than 1

These will be forwarded to Geant4 as stable particles

- [x] Testing needed

BEGINRELEASENOTES
- ddsim: add physics.alternativeStableStatus option to allow defining alternative stable status of MCParticles meant to be treated by Geant4, fixes #1446
ENDRELEASENOTES

### DIFF
--- a/DDG4/include/DDG4/Geant4InputAction.h
+++ b/DDG4/include/DDG4/Geant4InputAction.h
@@ -182,6 +182,9 @@ namespace dd4hep  {
       /// Property: set of alternative decay statuses that MC generators might use for unstable particles
       std::set<int> m_alternativeDecayStatuses = {};
 
+      /// Property: set of alternative stable statuses that MC generators might use for stable particles
+      std::set<int> m_alternativeStableStatuses = {};
+
       /// Perform some actions before the run starts, like opening the event inputs
       void beginRun(const G4Run*);
 
--- a/DDG4/python/DDSim/DD4hepSimulation.py
+++ b/DDG4/python/DDSim/DD4hepSimulation.py
@@ -459,6 +459,7 @@ class DD4hepSimulation(object):
         # this should never happen because we already check at the top, but in case of some LogicError...
         raise RuntimeError("Unknown input file type: %s" % inputFile)
       gen.AlternativeDecayStatuses = self.physics.alternativeDecayStatuses
+      gen.AlternativeStableStatuses = self.physics.alternativeStableStatuses
       gen.Sync = self.skipNEvents
       gen.Mask = index
       actionList.append(gen)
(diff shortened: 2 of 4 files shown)


andresailer:
Applied changes to the hepmc file in the testing folder, and I get the same result when updating the StableStatus set.

PR #1449, opened 2025-05-16
Prevent new TColor allocation in VisAttr::color()

veprbl:
We are observing a regression when using geoConvert (and a similar private tool) to do TGeo export.
### Expected:
image
### Actual:
image
(certain color replaced with black)

This traces to an upstream change https://github.com/root-project/root/pull/14063.
It appears that after this change many of our colors will get allocated as new instead of being picked from the pallete. They will still work in teveDisplay and geoDisplay, but not when saving to file (to show in JSROOT or other display).

BEGINRELEASENOTES
- Fix a regression with certain colors not saved to .root files when using ROOT 6.32.0+.

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/Objects.cpp
+++ b/DDCore/src/Objects.cpp
@@ -380,7 +380,7 @@ int VisAttr::color() const {
 void VisAttr::setColor(float alpha, float red, float green, float blue) {
   Object& o  = object<Object>();
   const auto num_before = gROOT->GetListOfColors()->GetLast();
-  // Set tolerance high enough to always lookup from existing pallete. This
+  // Set tolerance high enough to always lookup from existing palette. This
   // helps to preserve colors when saving TGeo to .root files.
   TColor::SetColorThreshold(1.0f/31.0f);
   Int_t col  = TColor::GetColor(red, green, blue);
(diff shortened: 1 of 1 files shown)


DDCore/src/Objects.cpp
```diff
@@ -379,7 +379,16 @@ int VisAttr::color() const {
 /// Set object color
 void VisAttr::setColor(float alpha, float red, float green, float blue) {
   Object& o  = object<Object>();
+  const auto num_before = gROOT->GetListOfColors()->GetLast();
+  // Set tolerance high enough to always lookup from existing pallete. This
+  // helps to preserve colors when saving TGeo to .root files.
+  TColor::SetColorThreshold(1.0f/31.0f);
```
veprbl, line 385: There is no obvious way to retrieve the current value via the public API, so this is done without a revert.

DDCore/src/Objects.cpp
```diff
@@ -379,7 +379,16 @@ int VisAttr::color() const {
 /// Set object color
 void VisAttr::setColor(float alpha, float red, float green, float blue) {
   Object& o  = object<Object>();
+  const auto num_before = gROOT->GetListOfColors()->GetLast();
+  // Set tolerance high enough to always lookup from existing pallete. This
```
andresailer, line 383:
```suggestion
  // Set tolerance high enough to always lookup from existing palette. This
```

MarkusFrankATcernch:
I do not like the printout to be WARNING. 
Most users use WARNING as a threshold, and will receive plenty of output ie. whenever allocating a new color.
This is not good and probably will in all application create additional noise.

PR #1454, opened 2025-05-22
CI: add gcc15 dbg dev3 builds

andresailer:
This adds builds using the Debug build mode, since there were errors triggered in the tests when DD4hep was build with Debug mode. Probably this is slower than the other instances of the CI builds.
To be seen and decided.

Also should decide which older builds to drop.


andresailer:
The gcc15-dbg test took 2h59m. Some other test to 2h40m, so not that much slower at least.

andresailer:
We will drop everything older than LCG_106.

PR #1456, opened 2025-05-26
fix: avoid exit codes in thisdd4hep.sh for `set -e` support

veprbl:
BEGINRELEASENOTES
- Wrap calls to grep in thisdd4hep.sh to make it source'able in scripts with `set -e`

ENDRELEASENOTES

Edge case is when $path_prefix is already in $path_value:
```output
# echo a | grep -v a; echo $?
1
```
if there is another path it just works
```
# (echo a; echo b) | grep -v a; echo $?
b
0
```


andresailer:
Thanks @veprbl !

PR #1483, opened 2025-08-04
Remove unused function in Geant4VolumeManager.cpp

jmcarcell:
BEGINRELEASENOTES
- Remove unused function in Geant4VolumeManager.cpp
- Fix the C++ includes (including what is used in each file)

ENDRELEASENOTES

It's declared in the .cpp file and it's not used there so it's not used anywhere
else.


andresailer (approved):
Thanks!

MarkusFrankATcernch:
Removing this function was not the most grandiose step....
These bit masking operations are always a bit tricky and I deliberately kept it 
as a good example (mostly for myself!) to remember how to properly do the masking.

jmcarcell:
If you point me to where I'm happy to some place I'm happy to document it. I see the same pattern happens in other places: https://github.com/AIDASoft/DD4hep/blob/master/DDCore/src/IDDescriptor.cpp#L142, so maybe we could find another place instead of unused code that is compiled.

MarkusFrankATcernch:
@jmcarcell 
This  is all true, but these volume managers were quite some head-ache. 
Also that the encoding has to go reverse was not at all clear from the begin. Until it really worked this was far from obvious.....
That bit bit 31 must be mapped to 0 and bit 0 is the 'highest' bit had to be learned with pain.
If this gets screwed up things go very quickly ballistic and is not obvious to debug.

andresailer:
> @jmcarcell This is all true, but these volume managers were quite some head-ache. Also that the encoding has to go reverse was not at all clear from the begin. Until it really worked this was far from obvious..... That bit bit 31 must be mapped to 0 and bit 0 is the 'highest' bit had to be learned with pain. If this gets screwed up things go very quickly ballistic and is not obvious to debug.

This sounds like a bunch of great comments to add to the function! What and why and how this removed function was doing things is not obvious at all, so having an annotated version of it back in the repo would be great!

PR #1484, opened 2025-08-07
Clean up and modernize DetectorImp.cpp

jmcarcell:
BEGINRELEASENOTES
- Clean up and modernize DetectorImp.cpp: Remove the unused struct TypePreserve, avoid calling find and then emplace in maps and call only emplace instead, use nullptr instead of 0, do not check pointers before deleting since deleting nullptr is a no-op, and simplify when possible.

ENDRELEASENOTES

While checking some code in this file I saw one thing then another... Certainly more things could be done.

### DIFF
--- a/DDCore/src/DetectorImp.cpp
+++ b/DDCore/src/DetectorImp.cpp
@@ -63,42 +63,26 @@ namespace {
 
   std::recursive_mutex  s_detector_apply_lock;
 
-  struct TypePreserve {
-    DetectorBuildType& m_t;
-    TypePreserve(DetectorBuildType& t)
-      : m_t(t) {
-    }
-    ~TypePreserve() {
-      m_t = BUILD_NONE;
-    }
-  };
-
   struct Instances  {
     std::recursive_mutex  lock;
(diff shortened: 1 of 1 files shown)


DDCore/src/DetectorImp.cpp
```diff
@@ -334,31 +313,32 @@ void DetectorImp::declareParent(const std::string& detector_name, const DetEleme
...
-             "Are you correctly building detector %s?",
-             de.name());
-    }
-    if ( (*i).second.isValid() )  {
-      Volume vol = (*i).second.volume();
-      if ( vol.isValid() )  {
-        return vol;
-      }
+  if ( !de.isValid() )   {
+    except("DD4hep","Detector: Attempt access mother volume of invalid detector [Invalid-handle]");
```
andresailer, line 317:
```suggestion
    except("DD4hep","Detector: Attempt to access mother volume of invalid detector [Invalid-handle]");
```

andresailer (approved):
Looks good to me.

PR #1494, opened 2025-09-02
CI: mac: drop cmake, already installed

andresailer:
BEGINRELEASENOTES
- CI: mac: drop cmake, already installed

ENDRELEASENOTES


andresailer:
It fixes the macOS CI builds (the test result is another test there that fails).

PR #1498, opened 2025-09-09
Modernize code related to Surfaces and the SurfaceManager

jmcarcell:
BEGINRELEASENOTES
- Modernize code related to Surfaces and the SurfaceManager. Simplify for loops, define default destructors with `=default`, pass arguments by const reference when possible
- Make `DetectorSurfaces` not inherit from `DetElement` since it is not used nor needed

ENDRELEASENOTES

### DIFF
--- a/DDRec/include/DDRec/DetectorSurfaces.h
+++ b/DDRec/include/DDRec/DetectorSurfaces.h
@@ -25,23 +25,19 @@ namespace dd4hep {
      * @date Apr, 10 2014
      * @version $Id$
      */
-    class DetectorSurfaces: virtual public DetElement {
+    class DetectorSurfaces {
 
     public:
-      typedef DetElement DetElement;
-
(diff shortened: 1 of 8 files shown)


andresailer:
Hi @MarkusFrankATcernch 

Could you have a look at these changes? Any reason things were the way they were and shouldn't be changed?

MarkusFrankATcernch:
@andresailer 
Short answer:no idea, Frank should know ;-)

Longer answer: 
I do not know why `DetectorSurfaces` have to be a `DetElement`. I think this is the main question.
If this `is-a`-relagtionship is not mandatory, then the changes of Juan are all correct. 
If  `DetectorSurfaces` have to be a `DetElement` then not. 

@gaede : do you remember  why this relationship was introduced?

MarkusFrankATcernch (approved):

PR #1503, opened 2025-09-18
[WIP] Add flag for fast simulation to MCParticle

tmadlener:
BEGINRELEASENOTES
- Add a flag to the MCParticle to track whether it has been handled by a fast simulation model

ENDRELEASENOTES

This is the start of this work. The goal in the end is to have an easy way in analysis to check whether an MCParticle has been handled by a fast simulation model or not. This will obviously also require EDM4hep and LCIO to acquire a similar flag, but I would like to finalize a few things here first before going there.

Currently there are a few open questions:
- Should we have multiple of these flags to disambiguate whether a particle has been fast simulated only in parts of the detector (i.e. probably `FAST_SIMULATION_TRK` and `FAST_SIMULATION_CALO`? Or do we simply name this current flag appropriately to `_CALO` (as that is the main / only use case for this)?
- Should setting this flag be responsibility of implementors of actual fast sim models, or can we still handle it on the DD4hep side? (The only place where we could conceivably do that would be in the `Geant4FastSimShowerModel::check_trigger`, but since killing the primary track is also in user responsibility, this should probably be as well)
- Does this flag need to do something else, other than keeping track of the fact that a particle has been fast simulated? (I don't think it does, but maybe others do)?

Tagging @peter-mckeown @gaede to solicit some inupt explicitly

- [ ] Needs https://github.com/key4hep/EDM4hep/pull/474

### DIFF
--- a/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
+++ b/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
@@ -458,6 +458,7 @@ void Geant4Output2EDM4hep::saveParticles(Geant4ParticleMap* particles)    {
       mcp.setHasLeftDetector(             mask.isSet(G4PARTICLE_SIM_LEFT_DETECTOR) );
       mcp.setStopped(                     mask.isSet(G4PARTICLE_SIM_STOPPED) );
       mcp.setOverlay(                     false );
+      mcp.setHandledByFastSim(            mask.isSet(G4PARTICLE_SIM_FAST_SIMULATION) );
 
       //fg: if simstatus !=0 we have to set the generator status to 0:
       if( mcp.isCreatedInSimulation() )


peter-mckeown:
Hi @tmadlener, thanks for starting this. Here's my two cents:

- I would say splitting this into `FAST_SIMULATION_TRK` and `FAST_SIMULATION_CALO` is probably the most generic approach, although calo is currently the main use case as you say
- I think it would be good to have some way of ensuring the user does not forget to set this flag, as this could cause confusion later (in particular if full simulation was not performed with detailed mode). Whether this is best done inside DD4hep by querying if `Geant4FastSimShowerModel::check_trigger` is passed as you say or externally I am not sure (I assume technically there is no blocker, since it is just setting a flag based on this condition? The only case I can imagine where having the trigger inside of DD4hep is restrictive is if the user doesn't kill the primary, but I don't think that is really a valid use case- and a flag on the primary wouldn't help then anyway).
- We have previously considered trying to add some kind of 'fast bit' that would keep track of whether `Geant4FastSimShowerModel::check_trigger` had failed at some point in the MC lineage of a given particle (to prevent unwanted fast sim triggers). However, I think this is trying to solve a different problem and is also a lot more involved, so probably out of scope for this

tmadlener:
I think I am missing something conceptual here entirely. The way I understand it the sensitive action is what transforms the G4Steps into dd4hep hits. However, the `Geant4FastSimShowerModel` essentially produces `G4Steps` and it is not a sensitive action but something "orthogonal". Hence, it also never passes through the constructor you pointed me to and I also don't think it becomes part of the sequence(?). 

It can be assigned to a region in the detector and if the trigger for it applies it will produce (G4) hits, e.g. via the `GFlashHitMaker`

https://github.com/AIDASoft/DD4hep/blob/dac8db64bc1eda281a0005d5728c4363afcfbb10/DDG4/plugins/Geant4GFlashShowerModel.cpp#L221-L222

Or via the one here

https://github.com/key4hep/DDML/blob/aa49b5f3fa42b7a16a72f1945871b0b1379b0351/include/DDML/FastMLShower.h#L185 

(which is [this one here](https://github.com/key4hep/DDML/blob/main/include/DDML/Geant4FastHitMakerGlobal.h)).

At least with the one in DDML this one also passes through `Geant4ParticleHandler::mark` once more. However, at least as far as I can see in case it is handled by a fast sim model, it doesn't have a sensitive action attached. I have now done this (some more extensive testing still needs to be done).

Issue #1529, opened 2025-12-01
Next podio tag will break EDM4hep reader


tmadlener:
https://github.com/AIDASoft/podio/pull/868

Makes `Frame::get` throw on non-existing collections. This will make ddsim fail with a fatal error when trying to do this

https://github.com/AIDASoft/DD4hep/blob/f2fa261e9382a0db6021beb700b19928e0a16881/DDG4/edm4hep/EDM4hepFileReader.cpp#L230-L237

jmcarcell:
Basically this means that building DD4hep 1.33 and below with podio greater than 1.6 will not be able to read EDM4hep files that do not have `edm4hep::GeneratorEventParametersCollection`, if one wants to specify this in the version requirements (for example, in Spack).

tmadlener:
I don't think the underlying issue is solved yet. And looking at the code again it looks like it will get somewhat messy if we want to keep things working for older podio versions. I think a clean option would be to require `EDM4hep > 1.0` once that is available and that would then also imply a minimum podio version. That way quite a bit of pre-processor handling could go. Do we want to be this radical?

andresailer:
> working for older podio versions

Do you mean older DD4hep versions (with newer podio)? Or is there yet another problem with older podio versions?

jmcarcell:
But
- DD4hep greater than 1.33: No problems since we check if the collections are in `frame.getAvailableCollections()` before doing anything, this works for any version of podio
- DD4hep 1.33 or older:
  - podio greater than 1.6: throw if the file doesn't have `edm4hep::GeneratorEventParametersCollection` or `edm4hep::EventHeaderCollection`, works fine if the file has both collections as they will be found
  - podio 1.6 or older: doesn't throw and creates empty collections instead if the collections are not there, fine if the collections are there

Note that the check for `edm4hep::GeneratorEventParametersCollection` only applies for EDM4hep 0.99.3, so if it's an older EDM4hep the same points above apply but removing this collection.

Right? So the issue happens with DD4hep 0.33 or older and podio greater than 1.6.

PR #1530, opened 2025-12-02
Fix the EDM4hep reader since frame.get throws if a collection is not found

jmcarcell:
BEGINRELEASENOTES
- EDM4hep reader: Prepare for podio frame.get to throw an exception if a collection is not found

ENDRELEASENOTES

Fix https://github.com/AIDASoft/DD4hep/issues/1529

### DIFF
--- a/DDG4/edm4hep/EDM4hepFileReader.cpp
+++ b/DDG4/edm4hep/EDM4hepFileReader.cpp
@@ -190,7 +190,7 @@ namespace dd4hep::sim {
     m_currEvent = event_number;
     podio::Frame frame = m_reader.readFrame("events", event_number);
     const auto& primaries = frame.get<edm4hep::MCParticleCollection>(m_collectionName);
-    const auto availableCollections = frame.getAvailableCollections();
+    const auto& availableCollections = frame.getAvailableCollections();
     int eventNumber = event_number, runNumber = 0;
 #if PODIO_BUILD_VERSION >= PODIO_VERSION(1, 6, 0)
     if (primaries.hasID()) {


DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -190,63 +190,64 @@ namespace dd4hep::sim {
...
-        try {
-          Geant4Context* ctx = context();
-          EventParameters *parameters = new EventParameters();
-          parameters->setRunNumber(0);
-          parameters->setEventNumber(event_number);
-          parameters->ingestParameters(frame.getParameters());
-          ctx->event().addExtension<EventParameters>(parameters);
-        } catch(std::exception &) {}
+      if (std::find(availableCollections.begin(), availableCollections.end(), m_eventHeaderCollectionName) !=
+          frame.getAvailableCollections().end()) {
```
andresailer:
```suggestion
          availableCollections.end()) {
```
?

DDG4/edm4hep/EDM4hepFileReader.cpp
```diff
@@ -190,63 +190,64 @@ namespace dd4hep::sim {
     m_currEvent = event_number;
     podio::Frame frame = m_reader.readFrame("events", event_number);
     const auto& primaries = frame.get<edm4hep::MCParticleCollection>(m_collectionName);
+    const auto availableCollections = frame.getAvailableCollections();
```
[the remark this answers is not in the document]
jmcarcell: `getAvailableCollections()` creates a vector of strings (can't avoid this) and then there is copy ellision: https://stackoverflow.com/questions/12953127/what-are-copy-elision-and-return-value-optimization/12953150#12953150 so `&` doesn't do anything here, but it doesn't hurt either.

DDG4/edm4hep/EDM4hepFileReader.cpp
andresailer: Well, since `getAvailableCollection()` creates a new vector every time it is called `getAvailableCollections().end() != getAvailableCollections().end()`, which makes for some hilarious debugging.

PR #1554, opened 2026-01-16
New SIMPLE_MODE hit creation mode.

tassiell:
BEGINRELEASENOTES
New SIMPLE_MODE hit creation mode. It saves the PreStepPoint position and momentum information and the StepLength.
The previous hit creation (with averaged information) has been moved to MEDIUM_MODE and set as the default.
This SIMPLE_MODE is more robust against information loss for tracks with significant bending.
This hit creation mode is particularly important for tracking systems when the hit step length is of the order of a few millimeters and the track curvature is not negligible, as for example in the IDEA Drift Chambers and other gaseous detectors.
ENDRELEASENOTES

### DIFF
--- a/DDCore/python/dd4hep_base.py
+++ b/DDCore/python/dd4hep_base.py
@@ -13,29 +13,116 @@ import importlib
 import types
 import logging
 
-
 logger = logging.getLogger(__name__)
 
 
+# ---------------------------------------------------------------------------
+def dd4hep_directories(g4=True):
+  """
+  Return tuple of relevant dd4hep directories: (dd4hep, rootsys, geant4)
+  """
+  import os
+  dd4hep = os.getenv("DD4hepINSTALL", "/usr")
+  rootsys = os.getenv("ROOTSYS", "/usr")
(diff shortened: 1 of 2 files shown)


DDG4/src/Geant4SensDetAction.cpp
```diff
@@ -106,7 +106,7 @@ Geant4Sensitive::Geant4Sensitive(Geant4Context* ctxt, const std::string& nam, De
     except("DDG4: Detector element for %s is invalid.", nam.c_str());
   }
   declareProperty("UseVolumeManager", m_useVolumeManager = true);
-  declareProperty("HitCreationMode",  m_hitCreationMode = SIMPLE_MODE);
+  declareProperty("HitCreationMode",  m_hitCreationMode = MEDIUM_MODE);
```
andresailer: This changes the default for all Geant4Sensitive, not just the Geant4SimpleTrackerAction. Which is why I think fixing the Geant4TrackerWeightedAction is less disruptive.

PR #1555, opened 2026-01-19
Allow to define user actions within the Geant4ParticleHandler processing

MarkusFrankATcernch:
BEGINRELEASENOTES
The Geant4ParticleHandler to propagate Monte-Carlo information to the DDG4 output record has the deficiency that
it is difficult for users to enhance the functionality. Effectively there is only one single user action possible and typically
set to the `Geant4TCUserParticleHandler`.
In this PR we allow to extend the functionality to a sequence of user actions. This functioanlity should enable the handling
of a usage which was pointed out in PR https://github.com/AIDASoft/DD4hep/pull/1503.

In the AlephTPC example we show an example how to enhance the `Geant4UserParticleHandler` functionality.
Execute:
```
$> python ../../DD4hep/examples/AlignDet/scripts/AlephTPC.py -mask
```
to execute an example. See the python script `DD4hep/examples/AlignDet/scripts/AlephTPC.py` for the setup of such enhancements.
ENDRELEASENOTES

### DIFF
--- a/DDG4/include/DDG4/Geant4ParticleHandler.h
+++ b/DDG4/include/DDG4/Geant4ParticleHandler.h
@@ -97,10 +97,8 @@ namespace dd4hep {
       bool m_printEndTracking    { false };
       /// Property: Flag to keep all particles generated
       bool m_keepAll             { false };
-      /// Property: Use extended user track handling in Geant4ParticleHandler::mark(const G4Track* track)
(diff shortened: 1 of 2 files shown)


MarkusFrankATcernch:
@andresailer Andre, please check Geant4ParticleHandler. For me it should be backwards compatible unless before (wronly) multiple handlers were assigned.

PR #1565, opened 2026-02-05
feat: support `ddsim --outputConfig.useRNTuple=true` in EDM4hep output

wdconinc:
This PR adds support to DDSim and the Geant4Output2EDM4hep plugin to write RNTuple output if supported by the podio installation. If requested but not supported, the user is warned and output falls back to TTree output.

The technical approach is with variants to limit the number of preprocessor directives that would be needed for every function that cascades to a podio call.

The determination of whether RNTuple support is available is at compile-time based on `__has_include` (C++17) since there is no clear podio CMake config variable (e.g. `PODIO_IO_BACKENDS`) that consistently exposes whether RNTuple is supported or not.

BEGINRELEASENOTES
- feat: support `ddsim --outputConfig.useRNTuple=true` in EDM4hep output

ENDRELEASENOTES

### DIFF
--- a/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
+++ b/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
@@ -301,15 +301,15 @@ void Geant4Output2EDM4hep::beginRun(const G4Run* run)  {
   if ( !fname.empty() && !m_file )   {
 #   if __has_include( <podio/RNTupleWriter.h> )
     if (m_rntuple) {
-      m_file = std::make_unique<writer_t>(podio::RNTupleWriter(fname));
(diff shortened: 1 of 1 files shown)


jmcarcell:
We have a generic `Reader` and `Writer` that we should use not to have to deal with the different types, and then we don't have to worry if podio was compiled with RNTuple support or not. In addition the env variable `PODIO_DEFAULT_WRITE_RNTUPLE` can be used: https://github.com/AIDASoft/podio/blob/master/doc/reading_writing.md?plain=1#L62. I had these almost ready (Andre and I talked about this today, before I thought we already had support for both reading and writing rntuples) so I'm going to pick up your changes in Python and open a PR so we can see.

wdconinc:
Defer to #1566.

PR #1566, opened 2026-02-05
Use the generic podio Reader and Writer to add support for RNTuples

jmcarcell:
BEGINRELEASENOTES
-  Use the generic `podio::Reader` and `podio::Writer` when podio is new enough to read and write EDM4hep files
-  Add a new parameter for ddsim: `--OutputConfig.useRNTuple` to overwrite the default behaviour of the Writer

ENDRELEASENOTES

Changes for the new parameter from ddsim copied from https://github.com/AIDASoft/DD4hep/pull/1565. 

Todo: 
- [x] Where to add a parameter to enable rntuples for the `DigiEdm4hepOutputProcessor`?

### DIFF
--- a/DDG4/python/DDSim/Helper/OutputConfig.py
+++ b/DDG4/python/DDSim/Helper/OutputConfig.py
@@ -159,7 +159,7 @@ class OutputConfig(ConfigHelper):
     return
 
   def _configureEDM4HEP(self, dds, geant4):
-    logger.info("++++ Setting up EDM4hep %s Output ++++", "RNTuple" if self.useRNTuple else "ROOT")
+    logger.info("++++ Setting up EDM4hep %s Output ++++", "RNTuple" if self.useRNTuple else "ROOT::TTree")
     e4Out = geant4.setupEDM4hepOutput('EDM4hepOutput', dds.outputFile)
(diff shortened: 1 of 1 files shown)


DDDigi/io/DigiEdm4hepInput.cpp
```diff
@@ -17,6 +17,9 @@
 
 // podio/edm4hep include files
 #include <podio/podioVersion.h>
+#if PODIO_BUILD_VERSION >= PODIO_VERSION(1, 0, 0)
+#include <podio/Reader.h>
+#else
 #if PODIO_BUILD_VERSION >= PODIO_VERSION(0, 99, 0)
```
wdconinc: I'd merge these with `#elif` to avoid the double `#endif` at the end.

andresailer:
@MarkusFrankATcernch 

Is there a way to to add a parameter to enable rntuples for the DigiEdm4hepOutputProcessor to switch between TTree and RNTuple output?

MarkusFrankATcernch:
DigiEdm4hepOutputProcessor has no provision for RNTuples.
But: Nobody is using DDDigi - do we have to extend it's functionality ?
Changes can easily be done if some client comes up on the horizon....

PR #1587, opened 2026-03-05
Fields/FieldTypes: performance related cleanups 

davidlange6:
The time spent getting field values in the ILD detector configuration caught my eye.  These changes unfortunately do not make a large impact (seems virtual method overheads instead, but tbc)

Performance related updates
- avoid sqrts when not needed
- avoid unnecessary creation of Position objects when they are just taken apart again

Changes that would change results if methods were used (likely they are not)
- ```OverlayedField::electromagneticField``` did not initialize all field members properly. Added 
- Some ```CartesianField::value``` methods did not add the current field value to the existing one. Changed to be all consistent (I think)

BEGINRELEASENOTES
- Fields/FieldTypes: performance related cleanups 
ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/Fields.h
+++ b/DDCore/include/DD4hep/Fields.h
@@ -213,9 +213,7 @@ namespace dd4hep {
     }
 
     /// Returns the 3 magnetic field components (x, y, z) if many components are present
-    void combinedMagnetic(const double* pos, double* field) const   {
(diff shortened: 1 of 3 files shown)


andresailer:
Stale, so closing it. The sqrt-removal is probably sensible to do. For the rest a more complete approach would be nice.

PR #1590, opened 2026-03-09
Fix more memory leaks

jmcarcell:
flagged by the leak sanitizer. Continuation of https://github.com/AIDASoft/DD4hep/pull/1586. These are a bit more complex.

BEGINRELEASENOTES
- `DetectorData.h`: fix a leak when defining multiple times the same constant. This is the case, for example, in the test `t_ClientTests_converter_vis_MagnetFields` where we have constants (like `world_side`) defined at the top level `MagnetFields.xml` and also in the included `SiDConstants.xml`. Because they are being put in a map, the second time the same constant appears emplacing will fail and there is a warning for that. However, that handle will never be deleted since it will never belong to the map.
- `DD4hep.cmake`: Allow listcomponents to run with the leak sanitizer during the build process since there are leaks coming from ROOT
- `StandardPlugins.cpp`: Make the `DD4hepUI` object static instead of allocating one with `new` that will never get deleted. Fixes the test `t_DDDigi_colored_noise`

ENDRELEASENOTES

For `DetectorData.h`, I can not call `e.destroy()` because it's const. Maybe it shouldn't be const?

### DIFF
--- a/DDCore/src/plugins/StandardPlugins.cpp
+++ b/DDCore/src/plugins/StandardPlugins.cpp
@@ -251,10 +251,11 @@ DECLARE_APPLY(DD4hep_Rint,run_interpreter)
  */
 static long root_ui(Detector& description, int /* argc */, char** /* argv */) {
   char cmd[256];
+  static dd4hep::detail::DD4hepUI s_ui(description);
   std::snprintf(cmd, sizeof(cmd),
-                "dd4hep::detail::DD4hepUI* gDD4hepUI = new "
-                "dd4hep::detail::DD4hepUI(*(dd4hep::Detector*)%p);",
-                (void*)&description);
+                "dd4hep::detail::DD4hepUI* gDD4hepUI = "
+                "(dd4hep::detail::DD4hepUI*)%p;",
+                (void*)&s_ui);
   gInterpreter->ProcessLine(cmd);
   printout(ALWAYS,"DD4hepUI",
            "Use the ROOT interpreter variable gDD4hepUI "
--- a/DDCore/include/DD4hep/DetectorData.h
+++ b/DDCore/include/DD4hep/DetectorData.h
@@ -75,6 +75,7 @@ namespace dd4hep {
               printout(WARNING,"Detector",
                        "+++ Object '%s' is already defined. New value will be ignored",
                        n.c_str());
+              delete e.ptr();
             }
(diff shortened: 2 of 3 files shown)


DDCore/src/plugins/StandardPlugins.cpp
MarkusFrankATcernch, line 1: No. Leave the "new". The variable is exported to the ROOT prompt and may be deleted there.
If static this will lead to a SEGV, because it is deleted again when all static destructors are called at shutdown.

jmcarcell: We agreed during the calls in the DD4hep developers meeting that this is OK, unlike the other change.

jmcarcell:
There were a few concerns about the changes in `StandardPlugins.cpp`. I have tested the following:
```
$ geoPluginRun -input ../DDDetectors/compact/SiD.xml -ui -interpreter
```
and then in the interpreter:
```
root [0]    std::cout << "gDD4hepUI is at: " << gDD4hepUI << std::endl;
gDD4hepUI is at: 0x563dda644720
root [1]    std::cout << "World volume: " << gDD4hepUI->detectorDescription()->worldVolume().name() << std::endl;
World volume: world_volume
root [2]    std::cout << "Constant world_side: " << gDD4hepUI->detectorDescription()->constantAsString("world_side") << std::endl;
Constant world_side: 30000*mm
```
But more importantly, using `.q` exits cleanly. Also calling without `-interpreter` exits cleanly. I think this is correct and all the tests pass. If there are any more paths that are useful to test I'm happy to, but so far I haven't found any issues and checked that the output is the same without these changes.
In the end this leak is not very important since it's leaked at exit when the program stops running but it allows to run tests where otherwise the sanitizer would flag them.

It seems also that DD4hepUI is only loaded when used, before:
```
$ time geoPluginRun -input ../DDDetectors/compact/SiD.xml -ui

real    0m1.260s
user    0m1.117s
sys     0m0.129s
```
after the changes:
```
$ time geoPluginRun -input ../DDDetectors/compact/SiD.xml -ui
...
real    0m0.329s
user    0m0.262s
sys     0m0.064s
```

MarkusFrankATcernch:
No need to have the discussion again ;-). 
Thanks for testing though!
Go ahead and merge!

PR #1602, opened 2026-03-28
feat: add Halton sequences support to isotrope generator

wdconinc:
Halton sequences are low-discrepancy sequences that fill phase space with faster variance reduction (1/N) than standard uniform point picking with PRNGs (1/sqrt(N)). It's not cheating statistics since you lose the Poisson statistical properties between two consecutive events. This technique is often referred to as RQMC, randomized quasi-Monte Carlo.

This adds scrambled Halton sequence support to the isotrope generators (where inter-event statistics are not considered since they don't represent real experimental running conditions). The scrambling uses Cranley-Patterson rotation, which is sufficient to remove correlations in three dimensional phase space sampling.

The sequences are scrambled with the random seed, so different runs with different seeds will produce different sequences. This also then allows statistical treatment to determine the errors on aggregate quantities (see note in ddsim help).

The various distributions are modified to take a sampler function that can either use PRNG or Halton sequences. For FFbar this is not possible since it uses an accept/reject algorithm that only works for PRNG.

Additional implementation notes:
- `getParticleMomentumUniform` needed to get the sample passed to it, so it was renamed to `sampleMomentum`. That leaves parent `Geant4ParticleGenerator::getParticleMomentumUniform` without any use or re-implementation in DD4hep, I think.
- `haltonValue` is written to use a pure integer loop with conversion to double only on return.

Since a picture (from wikipedia) is worth 1000 words:

| PRNG | RQMC |
|----------|-----------|
|image | image |

BEGINRELEASENOTES
- add Halton sequences support to isotrope generator

ENDRELEASENOTES

### DIFF
--- a/DDG4/include/DDG4/Geant4IsotropeGenerator.h
+++ b/DDG4/include/DDG4/Geant4IsotropeGenerator.h
@@ -17,13 +17,36 @@
 // Framework include files
 #include <DDG4/Geant4ParticleGenerator.h>
 
+// C/C++ include files
+#include <array>
+#include <cstdint>
+#include <functional>
+#include <mutex>
+
 /// Namespace for the AIDA detector description toolkit
 namespace dd4hep {
 
   /// Namespace for the Geant4 based simulation part of the AIDA detector description toolkit
   namespace sim {
+
+    // Forward declarations
+    class Geant4Random;
+
     /// Generate particles isotrop in space around origine (0,0,0)
     /**
+     *  Supports both standard PRNG-based sampling and Randomized Quasi-Monte Carlo
+     *  sampling via scrambled Halton sequences (Cranley-Patterson rotation).
+     *
+     *  When Halton mode is enabled (property "Halton" = true), the three sampling
+     *  dimensions use low-discrepancy Halton sequences in bases 2 (phi), 3 (theta),
+     *  and 5 (momentum), each additively scrambled by a per-dimension shift derived
+     *  from the property "HaltonSeed".  This provides superior phase-space coverage
+     *  relative to a PRNG while remaining reproducible and independent across jobs
+     *  (using different seeds or "HaltonOffset" values).
+     *
+     *  Note: the "ffbar" distribution is incompatible with Halton mode because its
+     *  acceptance-rejection loop cannot be driven by a fixed per-particle Halton point.
+     *
      *  \author  M.Frank
      *  \version 1.0
      *  \ingroup DD4HEP_SIMULATION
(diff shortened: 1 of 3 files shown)


DDG4/src/Geant4IsotropeGenerator.cpp
```diff
@@ -29,64 +29,105 @@ Geant4IsotropeGenerator::Geant4IsotropeGenerator(Geant4Context* ctxt, const std:
...
+/// Compute the radical inverse of index in the given base (Halton value in [0,1))
+double Geant4IsotropeGenerator::haltonValue(uint64_t index, int base)  {
+  uint64_t num   = 0;
+  uint64_t denom = 1;
+  while ( index > 0 )  {
+    denom *= base;
+    num    = num * base + (index % base);
+    index /= base;
+  }
+  return static_cast<double>(num) / static_cast<double>(denom);
```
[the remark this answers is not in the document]
wdconinc: Floating point math is intentionally avoided here to avoid rounding accumultaion, and overflow risk is acceptable for any reasonable index values.

Since `denom` is always a power of `base`, and `index` gets divided by `base` accordingly, the overflow happens when `index` is one power away from overflowing itself, i.e. essentially when we have more than 2<sup>64</sup> events. That's not a reasonable number of events to simulate.

The overflow margin decreases as phase space dimensionality (here 3) increases to large values. For a phase space dimension of 6543, with prime number 2<sup>16</sup>-1 = 65537, we would start overflowing sooner since 65537<sup>4</sup> overflows `uint64_t`. So we would be limited to 65537<sup>3</sup> = 2.8 x 10<sup>14</sup> events. Still very reasonable.

DDG4/include/DDG4/Geant4IsotropeGenerator.h
```diff
@@ -17,13 +17,36 @@
...
     /**
+     *  Supports both standard PRNG-based sampling and Randomized Quasi-Monte Carlo
+     *  sampling via scrambled Halton sequences (Cranley-Patterson rotation).
+     *
+     *  When Halton mode is enabled (property "Halton" = true), the three sampling
+     *  dimensions use low-discrepancy Halton sequences in bases 2 (phi), 3 (theta),
+     *  and 5 (momentum), each additively scrambled by a per-dimension shift derived
+     *  from the property "HaltonSeed".  This provides superior phase-space coverage
+     *  relative to a PRNG while remaining reproducible and independent across jobs
+     *  (using different seeds or "HaltonOffset" values).
```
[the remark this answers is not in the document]
wdconinc: Right, this was changed from an earlier dedicated seed to using the main Geant4Random seed.
```suggestion
     *  and 5 (momentum), each additively scrambled by a per-dimension shift derived
     *  from the Geant4Random seed.  This provides superior phase-space coverage
     *  relative to a PRNG while remaining reproducible and independent across jobs
     *  (using different seeds or "HaltonOffset" values).
```

DDG4/python/DDSim/Helper/Gun.py
```diff
@@ -39,6 +39,20 @@ def __init__(self):
...
+                          "superior phase-space coverage. The scrambling shifts are seeded from\n"
+                          "the simulation's Geant4Random engine (controlled by --random.seed).\n\n"
+                          "Note: the standard 1/sqrt(N) error estimate assumes i.i.d. samples and\n"
+                          "does NOT apply here. To estimate statistical errors, run M independent\n"
+                          "replications with different random seeds and use the spread across runs.\n\n"
+                          "Incompatible with distribution='ffbar': acceptance-rejection sampling\n"
+                          "cannot be driven by a fixed per-particle Halton point."}
+    self.halton = False
+    self._haltonOffset_EXTRA = {'help': "Starting index in the Halton sequence.\n\n"
+                                "Set to k*N for parallel-job partitioning (job k, N events each)."}
```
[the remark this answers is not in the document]
wdconinc: Proposed change too wordy.
```suggestion
                                "Set to k*N*m for parallel jobs (job k, N events, multiplicity m)."}
```

DDG4/src/Geant4IsotropeGenerator.cpp
```diff
@@ -29,64 +29,105 @@ Geant4IsotropeGenerator::Geant4IsotropeGenerator(Geant4Context* ctxt, const std:
...
+/// Sample momentum from [m_momentumMin, m_momentumMax] using a pre-computed [0,1) value h
+void Geant4IsotropeGenerator::sampleMomentum(double h, double& momentum) const  {
+  if ( m_energy != -1 )  {
+    momentum = m_energy;
+    return;
+  }
+  if ( m_momentumMax < m_momentumMin )
+    momentum = m_momentumMin + (momentum - m_momentumMin) * h;
+  else
+    momentum = m_momentumMin + (m_momentumMax - m_momentumMin) * h;
```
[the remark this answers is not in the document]
wdconinc: This follows the original pattern in `DDG4/src/Geant4ParticleGenerator.cpp`. Prefer to leave unchanged for reduced likelihood of unintentionally changed behavior.

DDG4/python/DDSim/Helper/Gun.py
```diff
@@ -39,6 +39,20 @@ def __init__(self):
                           "If not None, it will overwrite the setting of momentumMin and momentumMax"}
     self.energy = None
 
+    self._halton_EXTRA = {'help': "Use scrambled Halton sequence (RQMC) for particle gun sampling.\n\n"
```
andresailer: use triple quotes, and textwrap dedent?

DDG4/python/DDSim/Helper/Gun.py
```diff
@@ -39,6 +39,20 @@ def __init__(self):
...
+                          "Replaces the standard PRNG with a low-discrepancy sequence that gives\n"
+                          "superior phase-space coverage. The scrambling shifts are seeded from\n"
+                          "the simulation's Geant4Random engine (controlled by --random.seed).\n\n"
+                          "Note: the standard 1/sqrt(N) error estimate assumes i.i.d. samples and\n"
```
andresailer: what does i.i.d. stand for? This docstring is already long enough that it can be spelled out?

wdconinc: "Independent and identically distributed". It's a common abbreviation in English statistics teachings, but it's true that's not necessarily helpful for people who have not encountered it before.

```suggestion
                          "Note: the standard 1/sqrt(N) error estimate assumes independent\n"
                          "and identically distributed samples and\n"
```

DDG4/python/DDSim/Helper/Gun.py
wdconinc: You mean something like this?
```suggestion
    self._halton_EXTRA = {'help': dedent("""\
        Use scrambled Halton sequence (RQMC) for particle gun sampling.\n\n
```

andresailer: then you can wrap the whole block in one set if triple quote marks, and you can remove the explicit \n's

DDG4/python/DDSim/Helper/Gun.py
```diff
@@ -39,6 +40,29 @@ def __init__(self):
...
+
+            Replaces the standard PRNG with a low-discrepancy sequence that gives
+            superior phase-space coverage. The scrambling shifts are seeded from
+            the simulation's Geant4Random engine (controlled by --random.seed).
+            
+            Note: the standard 1/sqrt(N) error estimate assumes independent and
+            identically distributed (i.i.d.) samples and does NOT apply here.
+            To estimate statistical errors, run M independent replications with
+            different random seeds and use the spread across runs.
+
```
wdconinc:
```suggestion

            Note: the standard 1/sqrt(N) error estimate assumes independent and
            identically distributed (i.i.d.) samples and does NOT apply here.
            To estimate statistical errors, run M independent replications with
            different random seeds and use the spread across runs.

```

MarkusFrankATcernch:
@wdconinc @andresailer Is this MR supposed to be merged now?

andresailer:
> @wdconinc @andresailer Is this MR supposed to be merged now?

Yes, good to go!

andresailer (approved):

PR #1603, opened 2026-03-30
Remove unused template function calling non existing function.

ahennequ:
Remove unused template function calling non existing function.

This function is causing a build error when trying to compile with nvcc < 13.2. It works fine with other compilers (eg. gcc, clang) because the template function is never instantiated, so the compiler should be ignoring it. This is a bug in nvcc which doesn't ignore it and throws a compilation error. This issue is fixed in nvcc 13.2 but is present in all prior versions.

As part of the Gaudi-Allen convergence work we want to be able to compile, or at least parse DD4HEP headers with nvcc (for the host part).

FYI @sponce 

BEGINRELEASENOTES

* Shapes.h: Remove unused template function calling non existing function. (fix nvcc build)

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/Shapes.h
+++ b/DDCore/include/DD4hep/Shapes.h
@@ -950,11 +950,6 @@ namespace dd4hep {
                 double zneg, double zpos, int nsegments, double totphi)
     {  this->make(nam, twist_angle, rmin, rmax, zneg, zpos, nsegments, totphi);  }
 
-    /// Constructor to create a new identified tube object with attribute initialization
-    template <typename A, typename B, typename DZ>
-    TwistedTube(const std::string& nam, const A& a, const B& b, const DZ& dz)
-    {  this->make(nam, _toDouble(a), _toDouble(b), _toDouble(dz));   }
-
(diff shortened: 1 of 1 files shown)


andresailer:
Dear @ahennequ 

Thank you for your contribution, could you please keep the BEGIN/ENDRELEASENOTES in the PR description?

ahennequ:
> Dear @ahennequ
> 
> Thank you for your contribution, could you please keep the BEGIN/ENDRELEASENOTES in the PR description?

I've added them back

MarkusFrankATcernch:
TwistedTube is a construct from CMS. The fact this constructor is not used in the DD4hep code 
does not mean it is unused. Hence, if we want to remove it for nvcc, we need a proper #if !defined(...)  ... #endif

andresailer:
I think the point is that the make(str, dbl, dbl, dbl) function does not exist, so even cms cannot use this templated function?

MarkusFrankATcernch:
@andresailer I missed this one. Then of course the constructor can be suppressed.

PR #1605, opened 2026-04-10
Fix undefined behaviour in the BitFieldCoder

jmcarcell:
BEGINRELEASENOTES
- BitFieldCoder: Fix undefined behaviour if width is 32 (or greater) and the bitfield is unsigned. Then `1 << width` is undefined behaviour.

ENDRELEASENOTES

Note that above 1LL is being used for the signed branch.

### DIFF
--- a/DDCore/src/segmentations/BitFieldCoder.cpp
+++ b/DDCore/src/segmentations/BitFieldCoder.cpp
@@ -49,7 +49,7 @@ namespace dd4hep{
         _maxVal =  ( 1LL << ( _width - 1 ) ) - 1 ;
       
       } else {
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
Well spotted Dr. Watson!
Sub-bitfields with width > 32 are very rare,
but debugging things like this once hits are created wrongly is a true nightmare!

MarkusFrankATcernch:
@jmcarcell Can this MR be merged ?

jmcarcell:
Yes @MarkusFrankATcernch! The CI failure seems unrelated, we see a couple of tests that crash also in the LCG builds relatively often. Hopefully we investigate that soon...

PR #1608, opened 2026-04-13
Use subdetector-specific VolumeManager to avoid ambiguities.

scott-snyder:
When looking up a volume by VolumeID from VolumeManager, one needs to be careful of the possibility of VolumeIDs that are duplicated between different detector systems.  In that case, one need to first fetch the subdetector-specific VolumeManager and then do the lookup there.

In ALLEGRO, there are duplicate Volume IDs between the HCal and Si wrapper. Adjust the code here to work correctly in this case.

BEGINRELEASENOTES
- SDActions: Fix possible issues in looking up volumes by IDs in the case where VolumeIDs may be duplicated across subdetectors.
ENDRELEASENOTES


MarkusFrankATcernch:
@scott-snyder 
VolumeID duplicates between different subdetectors should only happen if the "system" id is the same.
Is this perhaps the reason you see duplicates between the HCal and Si ?
Otherwise duplicates should only be possible within the same subdetector.

scott-snyder:
> @scott-snyder VolumeID duplicates between different subdetectors should only happen if the "system" id is the same. Is this perhaps the reason you see duplicates between the HCal and Si ? Otherwise duplicates should only be possible within the same subdetector.

No, it can also happen with distinct system IDs if the bit width assigned to the system ID is different in two subsystems.   Here, HCAL_Barrel has an ID of 8 and a bit width of 4:
```
      <id>system:4,layer:5,row:9,theta:9,phi:10</id>
```

while SiWr_Disks has an ID of 24 and a bit width of 5:
```
<constant name="GlobalTrackerReadoutID" type="string" value="system:5,side:-2,layer:3,module:16,sensor:6"/>
```

So there can be (and are) duplicates between SiWr volumes and HCal volumes with the low bit of layer set.

I have pending changes to update the HCal so that the simulation produces the proper positions for the hits (like i did earlier for EcalB), but this issue was resulting in wrong results being silently produced in about 10% of runs.

This issue of duplicate IDs has been discussed in https://github.com/key4hep/k4geo/issues/568 and in a key4hep meeting.  The consensus was that it would be a very good idea to have a consistent system field bit width across all subsystems, with 8 bits sounding like a good value.  However, initial trials showed that likely requires nontrivial changes to the tracking code, so i don't want to wait for that to be resolved.

Also, comments in VolumeManager.h imply that this sort of duplication was forseen and expected:
```
   *  There are 2 working modes:
   *
   *  1) Starting from the top level element, the next level is addressed
   *  separately and the volumes are stored in a 2-level 'tree'. Placements
   *  this way are partitioned e.g. 'by subdetector'. This tries to keep
   *  the number of mapped elements at a reasonable level.
   *
   *  2) All placements are mapped to the top-level element. Good for small
   *  setups. Note, that the system field of all IDDescriptors of the
   *  subdetectors must have the same length to ensure the uniqueness of the
   *  placement keys.
```

But as long as the bit width of the system field depends on the system, there's no reliable way to extract the system from an arbitrary VolumeID,

MarkusFrankATcernch:
@scott-snyder 
Ohh. Yes of course! 

But we always said that the width of the system ID is fixed experiment wide 
for all subdetectors for exactly this reason!
But if this MR sort of overcomes this deficiency, the better. 

A problem though stays:
if during reconstruction you get an arbitrary hit with a CellID, 
the assignment to a subdetector stays ambiguous and this you cannot resolve without 
additional information like the container the hit is in etc.

Hence, we still strongly discourage having a different width of the system ID in different subdetectors.

PR #1610, opened 2026-04-13
Fix for building with CMAKE_INSTALL_MODE=ABS_SYMLINK.

scott-snyder:
Need to make scripts executable in the repo.

BEGINRELEASENOTES
- Tweak script permissions so that they work when installed as symlinks (CMAKE_INSTALL_MODE=ABS_SYMLINK).

ENDRELEASENOTES


DDG4/python/DDSim/bin/ddsim.in.py
andresailer, line 1: This file needs some variable replaced. How does this work with installing as symlink?

jmcarcell: `configure_file` does the substitution and then the symlinks can point to the build directory.

PR #1615, opened 2026-04-15
Remove checks for nullptrs before deleting, since deleting is a no-op

jmcarcell:
BEGINRELEASENOTES
- Remove checks for nullptrs before deleting, since deleting nullptr is a no-op

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/Parsers/Primitives.h
+++ b/DDCore/include/Parsers/Primitives.h
@@ -331,14 +331,12 @@ namespace dd4hep {
     }
     /// Helper to delete objects from heap and reset the pointer. Saves many many lines of code
     template <typename T> inline void deletePtr(T*& ptr) {
-      if (0 != ptr)
-        delete ptr;
+      delete ptr;
       ptr = 0;
     }
     /// Helper to delete objects from heap and reset the pointer. Saves many many lines of code
     template <typename T> inline void deleteObject(T* ptr) {
-      if (0 != ptr)
-        delete ptr;
+      delete ptr;
     }
     /// Helper to delete objects from heap and reset the pointer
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
NO. 
This is no good! Double delete leads to SEGV. 
Maybe `delete (nullptr);` is a noop, but the change at line 340 is a pure desaster.

jmcarcell:
But there is no double deletion.
- Before this PR: Only gets deleted when not nullptr
- Now: Always gets deleted, but `delete nullptr` doesn't do anything, so nothing changes. 
All tests are passing, if there was double deletion some tests would not pass since these are used in many places.

MarkusFrankATcernch:
@jmcarcell Not true. the deletion at line 340 does unconditionally delete whatever is at the pointer. 
The fact that it just not happens now is pure luck. `deletePtr` resets the external pointer to null. `deleteObject` does not. Hence it could be called twice outside of the control of the function.

MarkusFrankATcernch:
Another point is: we do not have to optimize to the last instruction. 
Geometry and detector description by nature are rather static.
Better play always safe.

jmcarcell:
> @jmcarcell Not true. the deletion at line 340 does unconditionally delete whatever is at the pointer. The fact that it just not happens now is pure luck. `deletePtr` resets the external pointer to null. `deleteObject` does not. Hence it could be called twice outside of the control of the function.

True if called twice there will be a double delete for a non-null ptr, but this is true with the changes in this PR and without them. The changes in the PR do not change the behaviour that one sees for both a `nullptr` and a non-null ptr in both functions.

andresailer (approved):
After further discussion, we all agree that this does not change behaviour.

PR #1620, opened 2026-04-24
Improve material properties. Introduce Geant4-custom properties

MarkusFrankATcernch:
BEGINRELEASENOTES
- Improve the interface to access material properties from materials if used:
```
    /// Access the number of properties attached to the material (if any)
    std::size_t numProperties()  const;
    /// Access to tabular properties of the material by index
    Property    property(std::size_t index)  const;
    /// Access to tabular properties of the material by name
    Property    property(const char* name)  const;
    /// Access to tabular properties of the material by name
    Property    property(const std::string& name)  const;
    /// Access string property value from the material table
    std::string propertyRef(const std::string& name, const std::string& default_value="");
    /// Access the number of const properties attached to the material (if any)
    std::size_t numConstProperties()  const;
    /// Access to const properties of the material by index
    double      constProperty(std::size_t index)  const;
    /// Access to tabular properties of the material
    double      constProperty(const std::string& name)  const;
    /// Access string property value from the material table
    std::string constPropertyRef(const std::string& name, const std::string& default_value="");
```
See for examples the area: examples/ClientTests/src/MaterialTester_geo.cpp

- In Geant4 special tags are used if the properties should be treated in a special way by Geant4:
  (see examples/OpticalSurfaces/compact/ReadMaterialProperties.xml)
  -- the tag `Geant4-ignore` hides the material property entirely from the Geant4 materials
  -- the tag `Geant4-custom` injects the property into the  Geant4 material. It may be used in user code 
     and accessed from regular Geant4 materials.
  -- the tag `Geant4-plugin` hides the property from Geant4 and allows the usage inside Geant4 plugins.

Users are e.g. RICH reconstruction and Channeling experiments. [Request from W,Pokorski]

ENDRELEASENOTES

### DIFF
--- a/DDCore/include/DD4hep/Objects.h
+++ b/DDCore/include/DD4hep/Objects.h
@@ -301,14 +301,22 @@ namespace dd4hep {
     double intLength() const;
     /// Access the fraction of an element within the material
     double fraction(Atom atom) const;
-    /// Access to tabular properties of the material
-    Property property(const char* name)  const;
-    /// Access to tabular properties of the material
-    Property property(const std::string& name)  const;
+    /// Access the number of properties attached to the material (if any)
+    std::size_t numProperties()  const;
+    /// Access to tabular properties of the material by index
+    Property    property(std::size_t index)  const;
+    /// Access to tabular properties of the material by name
+    Property    property(const char* name)  const;
+    /// Access to tabular properties of the material by name
+    Property    property(const std::string& name)  const;
     /// Access string property value from the material table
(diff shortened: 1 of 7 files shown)


pikacic:
FTR, this PR makes DD4hep incompatible with Geant4 < 11.

andresailer:
> FTR, this PR makes DD4hep incompatible with Geant4 < 11.

Hi @pikacic Do you still need compatibility with Geant4 < 11? Since we don't test this in the CI any longer...

PR #1625, opened 2026-04-29
Fix bug: add offset for boundary in Segmentation

### LINKED ISSUE #1624: Bug in Segmentation::positionToBin with vector of bin boundaries
### Check duplicate issues.

- [x] Checked for duplicates

### Goal

Dear developer, 

  I noticed there might be a bug in dd4hep [Segmentation::positionToBin](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/src/segmentations/Segmentation.cpp#L162) function: 
 The bin number is calculated with `position-offset` [Line 189](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/src/segmentations/Segmentation.cpp#L189), but the protection for hit outside of boundaries is calculated with position directly ([L171](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/src/segmentations/Segmentation.cpp#L171), [L179](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/src/segmentations/Segmentation.cpp#L179)). 
  With a non-zero offset value this function can not work. 

This is a very easy bug to fix, so if you allow, I can submit a PR for this. 

Best regards,
Fangyi

Alma linux 9

### compiler

GCC 14,2,0

### ROOT Version

6.38.04

### DD4hep Version

Github master branch

### Reproducer

None

### Additional context

_No response_

qwert2333:
BEGINRELEASENOTES
- Fix bug in the [Segmentation::positionToBin](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/src/segmentations/Segmentation.cpp#L162) function for irregular spaced binnings. The protection for hit outside of boundaries was calculated with position, the offset is missed. This affects PolarGridRPhi2 for the R coordinate with offset != 0. Fixes issue #1624 

ENDRELEASENOTES

### DIFF
--- a/DDCore/src/segmentations/Segmentation.cpp
+++ b/DDCore/src/segmentations/Segmentation.cpp
@@ -168,19 +168,19 @@ namespace dd4hep {
       if(fabs(position/cellBoundaries.back()-1.0)  < 3e-12) return int(cellBoundaries.size()-2);
 
       // hits outside cannot be treated
-      if(position < cellBoundaries.front()) {
+      if(position < cellBoundaries.front()+offset) {
         std::stringstream err;
         err << std::setprecision(20) << std::scientific;
         err << "Hit Position (" << position << ") is below the acceptance"
-            << " (" << cellBoundaries.front() << ") "
+            << " (" << cellBoundaries.front()+offset << ") "
             << "of the segmentation";
         throw std::runtime_error(err.str());
       }
-      if(position > cellBoundaries.back() ) {
+      if(position > cellBoundaries.back()+offset ) {
         std::stringstream err;
         err << std::setprecision(20) << std::scientific;
         err << "Hit Position (" << position << ") is above the acceptance"
-            << " (" << cellBoundaries.back() << ") "
+            << " (" << cellBoundaries.back()+offset << ") "
             << "of the segmentation";
         throw std::runtime_error(err.str());
       }
--- a/DDCore/src/segmentations/Segmentation.cpp
+++ b/DDCore/src/segmentations/Segmentation.cpp
@@ -160,33 +160,35 @@ namespace dd4hep {
     }
     /// Helper method to convert a 1D position to a cell ID given a vector of binBoundaries
     int Segmentation::positionToBin(double position, std::vector<double> const& cellBoundaries, double offset) {
-
+      const double relPosition = position - offset;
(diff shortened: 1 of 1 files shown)


DDCore/src/segmentations/Segmentation.cpp
andresailer: you would need offset also here

DDCore/src/segmentations/Segmentation.cpp
```diff
@@ -168,19 +168,19 @@ namespace dd4hep {
       if(fabs(position/cellBoundaries.back()-1.0)  < 3e-12) return int(cellBoundaries.size()-2);
```
andresailer: and here
```suggestion
      if(fabs(relPosition/cellBoundaries.back()-1.0)  < 3e-12) return int(cellBoundaries.size()-2);
```

andresailer (changes requested):
At least like this the changes are not fully consistent.

DDCore/src/segmentations/Segmentation.cpp
andresailer:
```suggestion
   const double relPosition = position - offset;
```
and then use relPosition on the rest of the function.

andresailer:
```suggestion
      if(fabs(relPosition/cellBoundaries.front()-1.0) < 3e-12) return 0;
```

DDCore/src/segmentations/Segmentation.cpp
```diff
@@ -168,19 +168,19 @@ namespace dd4hep {
...
 
       // hits outside cannot be treated
-      if(position < cellBoundaries.front()) {
+      if(position < cellBoundaries.front()+offset) {
```
andresailer:
```suggestion
      if(relPosition < cellBoundaries.front()) {
```

DDCore/src/segmentations/Segmentation.cpp
andresailer:
```suggestion
                                                                 relPosition);
```

andresailer (changes requested):

andresailer (approved):
Thanks!

andresailer:
We have some problems with the dev3 stacks at the moment (Due to an issue between Gaudi and ROOT master branches)

PR #1626, opened 2026-04-30
feat: add support for python geometry plugins

wdconinc:
This PR adds a `DECLARE_DETELEMENT` plugin `DD4hep_PythonDetector` that delegates geometry construction to a Python function specified by module/function/pythonpath XML attributes. This enables geometry to be distributed as pure Python packages without any per-project C++ compilation.

This PR also adds `DDCore/python/dd4hep_geo_plugin.py` providing the cppyy workarounds and type aliases needed by plugin authors (to be imported with `from dd4hep_geo_plugin import *`).

This PR also ports the IronCylinder example (the simplest I could find) from a C++ example to a minimal Python example, and includes it as a test.

BEGINRELEASENOTES
-  feat: add support for python geometry plugins

ENDRELEASENOTES

Note: This is for discussion; as I understand it from discussion with @jmcarcell there may be more computational intensive stuff done in the geometry plugins in FCC so this may impact initialization time in production and may add a translation step in project transition from prototype to production. In the case of EIC we have a challenge where incoming workforce is generally more familiar with python and this address that.

### DIFF
--- a/DDCore/python/dd4hep_geo_plugin.py
+++ b/DDCore/python/dd4hep_geo_plugin.py
@@ -1,0 +1,190 @@
+# ==========================================================================
+#  AIDA Detector description implementation
+# --------------------------------------------------------------------------
+# Copyright (C) Organisation europeenne pour la Recherche nucleaire (CERN)
+# All rights reserved.
+#
+# For the licensing terms see $DD4hepINSTALL/LICENSE.
+# For the list of contributors see $DD4hepINSTALL/doc/CREDITS.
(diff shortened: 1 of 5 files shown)


wdconinc:
Aha, interesting, https://github.com/wdconinc/DD4hep/actions/runs/25173235648/job/73798348230 fails since this requires ROOT v6.34 for `TPython::SwapWithObjAtAddr`. I'll make this functionality conditional on ROOT version.

PR #1627, opened 2026-05-04
fix: correct off-by-one error in dd4hep::detail::update_hash64(std::string)

veprbl:
BEGINRELEASENOTES
- Fixed a bug which caused update_hash64 to ignore last character of a std::string. This results in a change to geometry hash values.
ENDRELEASENOTES

before:
```c++
root [0] #include <DD4hep/Primitives.h>
root [1] dd4hep::detail::hash64(std::string("aaa"))
(unsigned long long) 620444549055354551
root [2] dd4hep::detail::hash64(std::string("aab"))
(unsigned long long) 620444549055354551
root [3] dd4hep::detail::hash64("aab")
(unsigned long long) 16653391238245862383
```
after:
```c++
root [0] #include <DD4hep/Primitives.h>
root [1] dd4hep::detail::hash64(std::string("aab"))
(unsigned long long) 16653391238245862383
root [2] dd4hep::detail::hash64(std::string("aaa"))
(unsigned long long) 16653392337757490594
```

### DIFF
--- a/DDCore/src/Primitives.cpp
+++ b/DDCore/src/Primitives.cpp
@@ -208,9 +208,7 @@ unsigned long long int dd4hep::detail::update_hash64(unsigned long long int hash
 /// 64 bit hash update function
 unsigned long long int dd4hep::detail::update_hash64(unsigned long long int hash, const void* key, std::size_t len)  {
   const unsigned char* str = (const unsigned char*)key;
-  if ( len > 0 )  {
-    for ( ; --len; ++str) hash = FNV1a_64::doByte(hash, *str);
-  }
+  for ( ; len--; ++str) hash = FNV1a_64::doByte(hash, *str);
   return hash;
 }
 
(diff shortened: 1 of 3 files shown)


andresailer:
Could you also update the test hash values?

veprbl:
Done.

DDCore/src/Primitives.cpp
```diff
@@ -208,9 +208,7 @@ unsigned long long int dd4hep::detail::update_hash64(unsigned long long int hash
 /// 64 bit hash update function
 unsigned long long int dd4hep::detail::update_hash64(unsigned long long int hash, const void* key, std::size_t len)  {
   const unsigned char* str = (const unsigned char*)key;
-  if ( len > 0 )  {
-    for ( ; --len; ++str) hash = FNV1a_64::doByte(hash, *str);
-  }
+  for ( ; len--; ++str) hash = FNV1a_64::doByte(hash, *str);
```
jmcarcell:
```suggestion
  for (std::size_t i = 0; i < len; ++i) hash = FNV1a_64::doByte(hash, str[i]);
```
What about this version where we don't have to think whether `--` applies before or after the check?

veprbl: Done, and it works.

DDCore/src/Primitives.cpp
jmcarcell, line 21:
```suggestion
#include <cstddef>
#include <cstring>
```
For `std::size_t`.

jmcarcell (approved):

PR #1629, opened 2026-05-13
Refactor Geant4Output2ROOT

jmcarcell:
BEGINRELEASENOTES
- Use `std::unique_ptr` when possible, return earlier to unindent big blocks and simplify when possible.

ENDRELEASENOTES

In preparation for https://github.com/AIDASoft/DD4hep/pull/1240, a few changes need to be done in `Geant4Output2ROOT` so it's a good time to do this.

In addition, I think assignments like this one were a bug: https://github.com/AIDASoft/DD4hep/blob/13d881a61f948a7b7a6501aa3617f6a7a84778b5/DDG4/src/Geant4Output2ROOT.cpp#L41

I have checked a few histograms and the output is the same before and after this change.

### DIFF
--- a/DDG4/include/DDG4/Geant4Output2ROOT.h
+++ b/DDG4/include/DDG4/Geant4Output2ROOT.h
@@ -16,6 +16,8 @@
 // Framework include files
 #include <DDG4/Geant4OutputAction.h>
 
+#include <memory>
+
 class TFile;
 class TTree;
 class TBranch;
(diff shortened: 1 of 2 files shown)


andresailer (approved):
Looks good to me.

jmcarcell:
I can rebase https://github.com/AIDASoft/DD4hep/pull/1240 after

PR #1636, opened 2026-05-28
fix: honor G4OpticalParameters process activation flags in DDG4

wdconinc:
The Geant4OpticalPhotonPhysics DDG4 plugin previously unconditionally constructed and registered all four optical processes (G4OpAbsorption, G4OpRayleigh, G4OpMieHG, G4OpBoundaryProcess) regardless of the G4OpticalParameters activation flags set by the user.

This patch makes the plugin (for Geant4 >= 10.7) check each process against G4OpticalParameters::GetProcessActivation() before adding it to the G4ProcessManager.  Processes that are deactivated are safely deleted instead.  G4OpBoundaryProcess is always registered since it controls all surface interactions.

Additionally, OpWLS and OpWLS2 are now registered when activated, completing the set of optical processes that can be controlled via
```
  /process/optical/processActivation <ProcessName> true/false
```

The log message now dynamically lists the actually-registered processes instead of a hardcoded string.

BEGINRELEASENOTES
- honor G4OpticalParameters process activation flags in DDG4

ENDRELEASENOTES

### DIFF
--- a/DDG4/plugins/Geant4OpticalPhotonPhysics.cpp
+++ b/DDG4/plugins/Geant4OpticalPhotonPhysics.cpp
@@ -34,6 +34,8 @@
 #include <G4OpRayleigh.hh>
 #include <G4OpMieHG.hh>
 #include <G4OpBoundaryProcess.hh>
+#include <G4OpWLS.hh>
+#include <G4OpWLS2.hh>
 #include <G4ParticleDefinition.hh>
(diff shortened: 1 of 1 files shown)


wdconinc:
https://github.com/AIDASoft/DD4hep/actions/runs/26594571685/job/78361956607?pr=1636 can't find ROOT for some reason...

PR #1646, opened 2026-06-18
Use Clang 22 instead of 19 in CI, soon unsupported

jmcarcell:
BEGINRELEASENOTES
- Use Clang 22 instead of 19 in CI, soon unsupported

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -24,7 +24,7 @@ jobs:
               "dev4/x86_64-el9-gcc13-opt",
               "dev4/x86_64-ubuntu2404-gcc13-opt",
               "dev3/x86_64-el9-gcc13-opt",
-              "dev3/x86_64-el9-clang19-opt",
+              "dev3/x86_64-el9-clang22-opt",
               "dev3/x86_64-el9-gcc15-dbg"]
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -34,28 +34,28 @@ jobs:
...
           echo "::group::CMakeConfig"
           # Common CMake options
           source ../.github/scripts/common_cmake_opts.sh
-          if [[ ${{ matrix.LCG }} =~ gcc13|clang19 ]]; then
-            echo "::group::CMakeConfig C++20"
+          if [[ ${{ matrix.LCG }} =~ gcc15|clang22 ]]; then
+            echo "::group::CMakeConfig C++23"
             cmake \
               $COMMON_CMAKE_OPTS \
               -DCMAKE_BUILD_TYPE=Release \
```
andresailer: You need to switch also the build types.

jmcarcell: Is there any logic? Before this PR: C++23 -> Debug, 20 -> Release, 17 -> Release, what should I use?
Edit: I kept it as it was.

andresailer: The build type just has to match the build type of the stack sourced.

PR #1648, opened 2026-07-04
feat: add region selection for G4HepEm plugin

wdconinc:
In using the G4HepEm plugin, we encountered a usability concern, namely that the G4Cerenkov process in our RICH detectors is also disabled for electrons (and positrons). This is a show stopper for us.

In this PR we add region selection to the G4HepEm plugin, by specifying the property HepEmRegions. A thin tracking manager wrapper is added before the G4HepEmTrackingManager to do the actual region selection.

Care is taken to avoid that WoodcockRegions and HepEmRegions are disjoint sets (i.e. warn the user when this is the case).

Note: we have to implement our own trackWithGeant4 since with a custom G4VTrackingManager this abstract base class does not cascade through to existing G4TrackingManager.

BEGINRELEASENOTES
- feat: add region selection for G4HepEm plugin

ENDRELEASENOTES

### DIFF
--- a/DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
+++ b/DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
@@ -147,6 +147,11 @@ namespace dd4hep {
         }
 
         void trackWithGeant4(G4Track* track) {
+          // We are inside a custom G4VTrackingManager::HandOverOneTrack() path here,
+          // but with track hand-off to Geant4 after all (since not in a HepEm region).
+          // In this path, G4EventManager does not execute its normal post-ProcessOneTrack()
+          // switch (see G4EventManager::DoProcessing, switch(istop) in the non-custom branch),
+          // so we must mirror that track-status/secondary/trajectory handling here.
           auto* eventManager = G4EventManager::GetEventManager();
           auto* trackManager = eventManager->GetTrackingManager();
           auto* stackManager = eventManager->GetStackManager();
--- a/DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
+++ b/DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
@@ -153,12 +153,12 @@ namespace dd4hep {
 
           trackManager->ProcessOneTrack(track);
 
-          G4TrackStatus istop = track->GetTrackStatus();
+          G4TrackStatus track_status = track->GetTrackStatus();
           G4VTrajectory* trajectory =
               trackManager->GetStoreTrajectory() == 0 ? nullptr : trackManager->GimmeTrajectory();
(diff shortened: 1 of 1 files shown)


DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
```diff
@@ -73,6 +87,123 @@ namespace dd4hep {
...
+            eventManager->StackTracks(secondaries);
+            delete trajectory;
+            delete track;
+            break;
+          }
+        }
+
+        bool fTrackInAllRegions;
+        std::unordered_set<G4int> fHepEmRegionIDs;
+        std::unique_ptr<G4HepEmTrackingManager> fHepEmTrackingManager;
```
andresailer: Why not `m_` for class variables?

DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
```diff
@@ -73,6 +87,123 @@ namespace dd4hep {
...
+        }
+
+        void trackWithGeant4(G4Track* track) {
+          auto* eventManager = G4EventManager::GetEventManager();
+          auto* trackManager = eventManager->GetTrackingManager();
+          auto* stackManager = eventManager->GetStackManager();
+
+          trackManager->ProcessOneTrack(track);
+
+          G4TrackStatus istop = track->GetTrackStatus();
```
andresailer: I find this variable name confusing. is it i_stop or is_top? or something else? Can we put status in the variable name?

DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
```diff
@@ -73,6 +87,123 @@ namespace dd4hep {
...
+          G4VTrajectory* trajectory =
+              trackManager->GetStoreTrajectory() == 0 ? nullptr : trackManager->GimmeTrajectory();
+          G4TrackVector* secondaries = trackManager->GimmeSecondaries();
+
+          switch (istop) {
```
andresailer: This could use some comments?

DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
wdconinc: `istop` is what Geant4 uses in the original code path that this is mirroring. But I agree that we can mirror logically, without using confusing variable names.

DDG4/g4hepem/Geant4HepEmTrackingPhysics.cpp
wdconinc: I added a comment inline at the top of this entire function, which is the fallback path to Geant4 code that we have to replicate here in essence since we are in a custom tracking manager.

andresailer (approved):
Thanks for addressing my comments!

PR #1649, opened 2026-07-07
g4PolarAngleScan

s6anloes:
BEGINRELEASENOTES
- Added `g4PolarAngleScan` to scan geometry material budget along the polar angle 

ENDRELEASENOTES

This PR introduces a new script in `DDG4/python/bin/` to run a material scan of the geometry in theta/cos(theta)/eta using geantinos and the `Geant4MaterialScanner/MaterialScan` step action.
The produced output is a root file which contains stacked histograms of the material depth, radiation length, interaction length (the stack is grouped by material), as well as individual histograms for each material.
In addition, the stacked histograms are exported as PDF and PNG.

The configurable parameters are inspired by [k4geo/utils/material_scan.py](https://github.com/key4hep/k4geo/blob/main/utils/material_scan.py) and [k4geo/utils/material_plots.py](https://github.com/key4hep/k4geo/blob/main/utils/material_plots.py).
One important difference is that the scripts in `k4geo` use the TGeo geometry, whereas this script uses the Geant4 geometry. There have been reports of problems with hyperboloid shapes in the TGeo material scan, which is why I'm introducing this Geant4 based script.

The script uses the ddsim particle gun to shoot geantinos in random directions.
Initial attempts with a Geant4 macro card, fixed directions, and repeated `/run/beamOn` commands were really slow in some of the detectors (e.g. the IDEA fibre dual-readout calorimeter). 
I didn't trace the problem deeply, I suspect between the runs the geometry is re-closed and -opened, causing long overhead times. 
This is entirely circumvented by using just one run with the ddsim particle gun, at the cost of not having the exact same number of events in every bin along the polar angle. I have switched to the Halton sequence option in the gun to mitigate this slightly.

The script can be run for instance with

`g4PolarAngleScan -c GEOMETRY.xml -o polar_scan.root --angleDef theta --minValue 10 --maxValue 170 -b 2 --eventsPerBin 100 -i Vacuum Air --noPilot --t 600 --saveRawData --seed 123456 --colors dimgrey firebrick gold mediumblue yellowgreen turquoise`

typically done only with one sub-detector to keep the number of materials overseeable.
Future work will aim to have it run with a full detector concept and group by sub-detector instead of material.


DDG4/python/bin/g4PolarAngleScan.py
```diff
@@ -0,0 +1,790 @@
...
+import array
+import bisect
+import ROOT
+from collections import defaultdict
+import threading
+from tqdm import tqdm
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawTextHelpFormatter,
+    description=(
```
andresailer: triple quotes and textwrap.dedent please.

DDG4/python/bin/g4PolarAngleScan.py
```diff
@@ -0,0 +1,790 @@
...
+        pass
+
+    print(f'WARNING: could not resolve color "{c}"')
+    return None
+
+
+def build_color_list(args_colors):
+    """
+    Create list from user input and fill with default colors if needed.
+    Ensure no dupicated colors in the final list.
```
andresailer:
```suggestion
    Ensure no duplicated colors in the final list.
```

DDG4/python/bin/g4PolarAngleScan.py
s6anloes: Like this in the newest commit?

andresailer: looks much more readable! Thanks!

DDG4/python/bin/g4PolarAngleScan.py
```diff
@@ -0,0 +1,791 @@
...
+import re
+import argparse
+import subprocess
+import math
+import array
+import bisect
+import ROOT
+from collections import defaultdict
+import threading
+from tqdm import tqdm
```
andresailer: tqdm is not a dependency for DD4hep, while it will usually be in the stack, this is not necessarily true in all cases. If you could make this optional?

DDG4/python/bin/g4PolarAngleScan.py
```diff
@@ -0,0 +1,791 @@
...
+    "--outputFile",
+    "-o",
+    dest="output",
+    default="materialScan.root",
+    type=str,
+    help="name of output root file with extension (used also as prefix for exported canvases)",
+    )
+
+parser.add_argument(
+    "--angleDef",
```
andresailer, line 82: add [choices](https://docs.python.org/3/library/argparse.html#choices) to the argument.

DDG4/python/bin/g4PolarAngleScan.py
```diff
@@ -0,0 +1,791 @@
...
+    dest="angleDef",
+    default="theta",
+    type=str,
+    help="angle definition to use: eta, theta, cosTheta or thetaRad. Default: theta",
+    )
+
+parser.add_argument("--minValue", dest="minValue", default=10, type=float, help="minimum value for eta/theta/cosTheta")
```
andresailer: what is the unit for min/maxValue? Does this work with `"1*rad"` for example?

DDG4/python/bin/g4PolarAngleScan.py
```diff
@@ -0,0 +1,791 @@
...
+print("geometry file:", args.compact)
+
+if args.steering is not None and not os.path.isfile(args.steering):
+    print("ERROR: cannot find requested ddsim steering file", args.steering, file=sys.stderr)
+    exit(1)
+print("ddsim steering file:", args.steering)
+
+angleDef = str(args.angleDef)
+if angleDef not in ANGLE_AND_DISTRIBUTION_DEFS.keys():
+    print("ERROR: unknown angle definition", angleDef, ". Choose from ",
```
andresailer: you don't need this if you have `choices` above.

DDG4/python/bin/g4PolarAngleScan.py
```diff
@@ -0,0 +1,791 @@
...
+if not math.isclose(actual_binning, binning, abs_tol=1e-6):
+    print(f"WARNING: bin width adjusted from {binning} to {actual_binning:.6g} "
+          f"to fit exactly {nBins} bins in [{minValue}, {maxValue}]")
+
+bin_edges = [minValue + i * actual_binning for i in range(nBins + 1)]
+bin_edges[-1] = maxValue  # avoid floating point rounding issues for the last bin edge
+
+
+def find_bin(angle_value, bin_edges):
+    ib = bisect.bisect_right(bin_edges, angle_value) - 1
```
andresailer: Do you always have uniform bins?
Can't you just

$$\text{index} = \lfloor \frac{\text{angleValue} - \text{minValue}}{\text{actualBinning}} \rfloor + 1$$

andresailer (commented):
I hope we never have to change the format of the material scan output...

s6anloes:
> I hope we never have to change the format of the material scan output...

Yes I agree. Are you suggesting we refactor the parsing, since the  [g4GraphicalScan](https://github.com/AIDASoft/DD4hep/blob/285e3402e58054f905ffae8702e90ad762555dd0/DDG4/python/bin/g4GraphicalScan.py#L342) does the same thing?
This way we'd only have to update in one place at least

DDG4/python/bin/g4PolarAngleScan.py
s6anloes: Indeed, it does not work with `"1*rad"`. The value is expected to be a float already in the correct units. I have updated the help text to clarify this

DDG4/python/bin/g4PolarAngleScan.py
s6anloes: Yes, you're right. I have simplified

DDG4/python/bin/g4PolarAngleScan.py
s6anloes: Optional now with default `False` (not showing progress bar)

andresailer:
> Yes I agree. Are you suggesting we refactor the parsing, since the [g4GraphicalScan](https://github.com/AIDASoft/DD4hep/blob/285e3402e58054f905ffae8702e90ad762555dd0/DDG4/python/bin/g4GraphicalScan.py#L342) does the same thing?
This way we'd only have to update in one place at least


That would be nice, probably.

s6anloes:
I have created a new module which does the parsing, and adapted `g4PolarAngleScan.py` and `g4GraphicalScan.py` accrodingly.
I have tested that both scripts still produce the expected results, still I invite @danieljeans to have a look himself and make sure.

Also added an option to suppress the PDF and PNG output

DDG4/python/bin/MaterialScanner_Parser.py
```diff
@@ -0,0 +1,152 @@
...
+#
+# For the licensing terms see $DD4hepINSTALL/LICENSE.
+# For the list of contributors see $DD4hepINSTALL/doc/CREDITS.
+#
+# ==========================================================================
+"""
+Parser for the output of the Geant4MaterialScanner/MaterialScan.
+Used by g4GraphicalScan.py and g4PolarAngleScan.py to extract the material budget along a geantino track.
+In case of any updates to the output format of the scanner, only this file needs to be adapted.
+This is a pure parser, how the infromation is used differs between g4GraphicalScan and g4PolarAngleScan.
```
andresailer:
```suggestion
This is a pure parser, how the information is used differs between g4GraphicalScan and g4PolarAngleScan.
```

DDG4/CMakeLists.txt
```diff
@@ -205,9 +205,11 @@ get_filename_component(PYTHON_INTERPRETER_BINARY ${Python_EXECUTABLE} NAME)
 configure_file(python/DDSim/bin/ddsim.in.py ${CMAKE_BINARY_DIR}/bin/ddsim @ONLY)
 install(PROGRAMS ${CMAKE_BINARY_DIR}/bin/ddsim DESTINATION bin)
 
-install(PROGRAMS python/bin/g4MaterialScan.py  DESTINATION bin RENAME g4MaterialScan)
-install(PROGRAMS python/bin/g4GeometryScan.py  DESTINATION bin RENAME g4GeometryScan)
-install(PROGRAMS python/bin/g4GraphicalScan.py DESTINATION bin RENAME g4GraphicalScan)
+install(FILES python/bin/MaterialScanner_Parser.py DESTINATION bin)
```
andresailer:
```suggestion
install(FILES python/bin/MaterialScanner_Parser.py DESTINATION ${DD4HEP_PYTHON_INSTALL_DIR})
```

andresailer (commented):
Thanks for the refactor. The new file shouldn't be in `bin` though. We can put it where other dd4hep python files are installed.

s6anloes:
Implemented @andresailer comments and rebased

PR #1653, opened 2026-07-21
DetectorSensitives: assign sensitive detector to reflected object if …

### LINKED ISSUE #1650: No hit produced when mirror reflection
Hello,
- I am implementing a detector for EIC using DD4hep.
- The detector is in TWO pieces, backward and forward. Forward should be the mirror image of backward.
- When I do apply the mirror reflection, `npsim` produces no hit in the mirrored piece (while le GL viewer looks nice, as expected).
- When I instead apply a rotation through the origin, everything being otherwise the same, `npsim` does produce hits.
- In short:
```
Transform3D tr = restoreZ * Rotation3D(1,0,0,0,1,0,0,0,-1) * rot * origin2Centre * tilt; // Does not work
Transform3D tr = restoreZ * RotationZYX(0, M_PI, 0)        * rot * origin2Centre * tilt; // Works
```

Do I miss something? Is there some action I should take to enable the hits in the mirrored image?
Regards,
Yann Bedfer

andresailer:
…the logical volume was sensitive

BEGINRELEASENOTES
- DDG4: DetectorSensitives: assign sensitive detector to reflected object if the original logical volume was sensitive, fixes #1650 

ENDRELEASENOTES

So the G4ReflectionFactory

https://github.com/AIDASoft/DD4hep/blob/8e15ffeef91c2674a0d1bbd158949c4397c3d3de/DDG4/src/Geant4Converter.cpp#L1164-L1170

Creates a new G4LogicalVolume (...+"_refl"), but we only assign the sensitive detector to the original logical volume. And here in that conversion we have not yet assigned the sensitive detector to the logical volume, so we cannot just copy it over from the original LogicalVolume.

@MarkusFrankATcernch : is there a more sensible way of doing this without checking the reflectionfactory for all entries? Although that maybe is a small number of reflected logical volumes, so the overhead is small?

### DIFF
--- a/DDG4/plugins/Geant4DetectorSensitivesConstruction.cpp
+++ b/DDG4/plugins/Geant4DetectorSensitivesConstruction.cpp
@@ -12,6 +12,9 @@
 //
 //==========================================================================
 
+#include "G4ReflectionFactory.hh"
+#include "G4LogicalVolume.hh"
+
 // Framework include files
 #include <DDG4/Geant4DetectorConstruction.h>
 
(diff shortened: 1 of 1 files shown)


MarkusFrankATcernch:
Aaargh! My bad!

This means CMS was never using reflected sensitive volumes...
Since this feature was developed originally for you @ianna : Is this true ?
Otherwise you will need to apply some corrections.

andresailer:
Is CMS using this part of DD4hep for their simulation?

MarkusFrankATcernch:
No. This must be the reason they have not seen it. 
They only use the Geant4Converter and this one does not deal with the
sensitive volumes.

Issue #1660, opened 2026-08-21
OutputAction and contexts


andresailer:
https://github.com/AIDASoft/DD4hep/blob/1a7e214e60f982c1886cef61e9404969a4dc9fd5/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L387-L407

>      // In multithreaded running, the run is present in only one of the contexts 

In multithreaded running with the outputAction being a shared action the context that is part of the shared action is actually the "master" context, so the run is not present in the context used here, ever. The run is present in all the worker contexts.

We need to get the worker context from the thread and get the run from that, but then we store one run for each thread, which is not what we want?

tmadlener:
For this particular case we use the run context mainly to transport the parameters along, right? Could we put that into the "master" context as well? (Maybe as a duplicate from what we put into the run context?)

andresailer:
> Could we put that into the "master" context as well?

Yeah, good idea, that would be an alternative solution. I guess we won't have multiple runs in flight.

andresailer:
I started something in #1663, but I will try with putting the run into master-context and getting it from there next week.

andresailer:
~~Putting the run in the master context works in the file reader, but~~ in the output stage there is no run in the master context in multithreaded mode, only in single threaded mode. I am not sure why that is the case.

No, no run in the master context in the reader, it was just catching the exception and I didn't read all the output.

andresailer:
> Could we put that into the "master" context as well?

There is no "run" in the master context in multi-threaded mode. Run only goes to each worker thread.
But the run parameters are only added to one of the worker contexts, so that part was kind of correct in the comment. Now I can write the run only when there are parameters.

PR #1661, opened 2026-08-21
Geant4Output2EDM4hep: always use the metadata category for metadata

### LINKED ISSUE #1659: EDM4hep writer creates two categories for metadata
I just realized that the EDM4hep output action writes a "metadata" category

https://github.com/AIDASoft/DD4hep/blob/1a7e214e60f982c1886cef61e9404969a4dc9fd5/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L331-L338

as well as another metadata frame into a "meta" category

https://github.com/AIDASoft/DD4hep/blob/1a7e214e60f982c1886cef61e9404969a4dc9fd5/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L397-L405

This was originally introduced in https://github.com/AIDASoft/DD4hep/pull/1371

Naively I would have assumed there to be only one location for this sort of data. From the looks of it the main (only?) thing that ends up in the `metadata` category is the CellID encoding strings. On the other hand things that were ingested from an EDM4hep file go into the `meta` category, so essentially they change category via `ddsim`.

Ingestion happens here:

https://github.com/AIDASoft/DD4hep/blob/1a7e214e60f982c1886cef61e9404969a4dc9fd5/DDG4/edm4hep/EDM4hepFileReader.cpp#L161-L170

This looks like a bug we have overlooked when adding the EDM4hep ingestion. (Also we should probably switch to `podio::Categories` for common category names)

andresailer:
BEGINRELEASENOTES
- Geant4Output2EDM4hep: always use the metadata category for metadata, fixes #1659 

ENDRELEASENOTES

### DIFF
--- a/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
+++ b/DDG4/edm4hep/Geant4Output2EDM4hep.cpp
@@ -333,6 +333,12 @@ void Geant4Output2EDM4hep::saveFileMetaData() {
   for (const auto& [name, encodingStr] : m_cellIDEncodingStrings) {
     metaFrame.putParameter(podio::collMetadataParamName(name, CellIDEncoding), encodingStr);
   }
+  if (context()->runPtr() != nullptr) {
+    FileParameters* parameters = context()->run().extension<FileParameters>(false);
+    if ( parameters ) {
+      parameters->extractParameters(metaFrame);
+    }
+  }
   G4AutoLock protection_lock(&action_mutex);
   m_file->writeFrame(metaFrame, "metadata");
 }
@@ -391,17 +397,6 @@ void Geant4Output2EDM4hep::saveRun(const G4Run* run)   {
         parameters->extractParameters(runHeader);
       }
       m_file->writeFrame(runHeader, "runs");
-    }
-  }
-  {
-    // In multithreaded running, the run is present in only one of the contexts
-    if (context()->runPtr() != nullptr) {
-      podio::Frame metaFrame {};
-      FileParameters* parameters = context()->run().extension<FileParameters>(false);
-      if ( parameters ) {
-        parameters->extractParameters(metaFrame);
-      }
-      m_file->writeFrame(metaFrame, "meta");
     }
   }
 }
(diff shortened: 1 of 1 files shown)


DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -333,8 +333,14 @@ void Geant4Output2EDM4hep::saveFileMetaData() {
   for (const auto& [name, encodingStr] : m_cellIDEncodingStrings) {
     metaFrame.putParameter(podio::collMetadataParamName(name, CellIDEncoding), encodingStr);
   }
+  if (context()->runPtr() != nullptr) {
```
andresailer, line 336: This is just a copy/paste from the other location, but actually see #1660

andresailer:
A bit suspicious that this test fails with
```
     Geant4Output2EDM4hep INFO  +++ Saving EDM4hep event 0 run -1.
    free(): invalid pointer
    [FATAL] (SignalHandler) Handle signal: 6 [SIGABRT] Old action:0x7f064ced06e0 Mem:0x1358 Code:FFFFFFFA
    GenerationInit   WARN  +++ Finished run 0 after 1 events (1 events in total)
     *** Break *** abort
```

But weird it only fails in one stack.

tmadlener:
this is vaguely reminiscent of https://github.com/key4hep/k4geo/pull/628#issuecomment-5297113161. Could it be some threading issue?

tmadlener (commented):
This now looks like it does what it should do. All CellID encodings and the parameters that were already present in the input should end up in the output. Can we easily adapt some test to confirm this? It looks like the `ZH250_ISR.edm4hep.root` file that we use in tests for the reader doesn't contain any `metadata` to begin with.

andresailer:
> this is vaguely reminiscent of [key4hep/k4geo#628 (comment)](https://github.com/key4hep/k4geo/pull/628#issuecomment-5297113161). Could it be some threading issue?

Yes, seems to be similar. Also of course annoying that it passed when I retried. But I guess I need to set up a key4hep based build locally and re-run a few times hoping to get lucky.

andresailer:
Since I will work on #1660 as well, I will merge this here and open a new PR and also add some tests. Since this exception happened before in that test, it is probably not caused by the changes here.

PR #1663, opened 2026-08-21
Geant4Output2EDM4hep: get run from worker context

### LINKED ISSUE #1660: OutputAction and contexts

andresailer:
This will need some more work, ~~the current status writes one run per thread.~~

BEGINRELEASENOTES
- Geant4OutputAction: In multi-threaded mode, make sure the ingested runparameters and fileparameters end up in the output, fixes #1660
  -  Fixed locking endRun action of the outputaction
- Geant4Output2LCIO: fix locking of output writing in multi-threaded mode
- Geant4Output2ROOT: fix locking of output writing in multi-threaded mode
- Tests: added comparison of run and metadata trees if they exist in the output

ENDRELEASENOTES

TODO

- [x] Output2ROOT
- [x] Output2LCIO: ~~writes one run per thread~~
- ~~Common refactor?~~


tmadlener:
> the current status writes one run per thread.

I'm assuming a "trivial" solution employing `std::call_once` doesn't do the trick? Might be a bit trickier if there are multiple runs as well as multiple threads...

tmadlener (commented):
This looks deceptively simple in the end, but I think following your comments / explanations it should be correct.

Somewhat related: Can we handle multiple runs in flight in a multi-threaded context correctly in DD4hep/ddsim? I think at least from the podio side things should keep working, as long as all runs are added to a worker context at some point.

andresailer:
> Somewhat related: Can we handle multiple runs in flight in a multi-threaded context correctly in DD4hep/ddsim? I think at least from the podio side things should keep working, as long as all runs are added to a worker context at some point.

Unless I already forgot again: run information is read when the file is opened. There is no sequential processing of files. Only one run per file, so no treatment of multiple runs in parallel.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -389,15 +393,12 @@ void Geant4Output2EDM4hep::saveRun(const G4Run* run)   {
...
-        parameters->extractParameters(runHeader);
-      }
-      m_file->writeFrame(runHeader, podio::Category::Run);
-    }
+  RunParameters* parameters = workerCtx->run().extension<RunParameters>(false);
+  if ( parameters ) {
+    // only one of the worker contexts (or the master context) has run parameters, since only one reader is created
+    // which registers the run parameters
+    parameters->extractParameters(runHeader);
+    m_file->writeFrame(runHeader, podio::Category::Run);
```
andresailer: I have to check. Are there always runParameters, or are there cases where we don't have runParameters, but we still want to write the runheader.

andresailer:
> deceptively

Indeed

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -368,6 +369,9 @@ void Geant4Output2EDM4hep::commit( OutputContext<G4Event>& /* ctxt */)   {
 /// Callback to store the Geant4 run information
 void Geant4Output2EDM4hep::saveRun(const G4Run* run)   {
   G4AutoLock protection_lock(&action_mutex);
+  // This action is shared, so by default it has the "master" context, we need to get the threaded context
+  auto* workerCtx = context()->kernel().worker(Geant4Kernel::thread_self(), false).workerContext();
```
andresailer: Sometimes the kernel is not the master kernel (even though everything should be the master kernel?) And then we don't have the workers we should have.

andresailer: So both LCG_106 and key4hep nightlies fail with something like
```
    Geant4Kernel     ERROR DDG4: The worker instance 1843390016l does not exist!
```

Event though earlier:
```
    Geant4Kernel     INFO  +++ Created worker instance id=1843390016l
```

These are different versions of geant4 (11.4 and 11.2.1 (LCG_106))

It seems to be deterministic. I can reproduce the issue with LCG_106 locally. And also with my LCG_109 based dev stack. I don't know why it doesn't fail in the other builds.

tmadlener: > I don't know why it doesn't fail in the other builds.

without having looked and just to exclude the obvious: all the other stacks build `DDG4` with EDM4hep enabled?

andresailer: Yeah, we only turn on EDM4hep for 106 for some reason...

DDG4/include/DDG4/Geant4Action.h
```diff
@@ -127,6 +130,8 @@ namespace dd4hep {
       PropertyManager    m_properties   {   };
       /// Reference count. Initial value: 1
       long               m_refCount     { 1 };
+      ///Synchronisation lock
+      mutable G4Mutex m_mutex = G4MUTEX_INITIALIZER;
```
andresailer: From DD4hep meeting discussion: Rather put this in the Geant4EventAction where it is needed, and not in all actions.

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
```diff
@@ -367,7 +361,9 @@ void Geant4Output2EDM4hep::commit( OutputContext<G4Event>& /* ctxt */)   {
 
 /// Callback to store the Geant4 run information
 void Geant4Output2EDM4hep::saveRun(const G4Run* run)   {
-  G4AutoLock protection_lock(&action_mutex);
+  // This action is shared, so by default it has the "master" context, we need to get the threaded context
+  auto* workerCtx = context()->kernel().worker(Geant4Kernel::thread_self(), false).workerContext();
```
andresailer: From DD4hep meeting discussion: Instead of having the RunParameters in one of the many runs. Use a shared object that is added to all the run instance of the same run

andresailer:
EventReader test fails only in 107a
106: no EDM4hepFileReader, podio is too old
107a: issue with
```
16: Warning in <TClass::Init>: no dictionary for class podio::root_utils::CollectionWriteInfo is available
```
(podio 1.2.0)

109: this warning is not present (podio 1.7.0)

Still weird why it started failing so, need to investigate.

andresailer:
> Still weird why it started failing so, need to investigate.

Because we just turned on EDM4hep for the CI

tmadlener:
What is the file version with which the input file has been written? (`podio-dump` on the input file should tell you). Since podio doesn't make forward compatibility guarantees, it's possible that the input file is too new for certain stacks.

andresailer:
> possible that the input file is too new for certain stacks.

106
---
 podio-dump crashes

107a
----

```
TClass::Init:0: RuntimeWarning: no dictionary for class podio::root_utils::CollectionWriteInfo is available
input file: ../DDTest/inputFiles/ZH250_ISR.edm4hep.root
            (written with podio version: 1.6.0)

datamodel model definitions stored in this file: 
 - edm4hep (0.99.99)

Frame categories in this file:
Name    Entries
------  ---------
ERROR: Cannot print category 'events' (not present in file)
$ rootls ../DDTest/inputFiles/ZH250_ISR.edm4hep.root
events  podio_metadata  runs
```
109a
----
```
podio-dump DDTest/inputFiles/ZH250_ISR.edm4hep.root 
input file: DDTest/inputFiles/ZH250_ISR.edm4hep.root
            (written with podio version: 1.6.0)

datamodel model definitions stored in this file:
 - edm4hep (0.99.99)

Frame categories in this file:
Name    Entries  
------  -------  
runs    1        
events  3        
################################### events: 0 ####################################
Collections:
Name (s = subset collection)  ValueType                          Size  ID          
----------------------------  ---------------------------------  ----  --------    
EventHeader                   edm4hep::EventHeader               1     d793ab91    
GeneratorEventParameters      edm4hep::GeneratorEventParameters  1     a193a48b    
MCParticles                   edm4hep::MCParticle                125   a1cba250    

Parameters:
Name  Type  Elements  
----  ----  --------  
```

DDG4/include/DDG4/Geant4Action.h
andresailer: Also needed in EventActionSequence, but moved out of Geant4Action

andresailer:
I am going to disable the edm4hep filereader test for podio < 1.6

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
andresailer: Now the runparameters and fileparameters from any thread are stored until we are ready to close the file. An error is emitted if we happen to have more than one of them (so we should get test failures).

DDTest/python/compare_simulation_output.py
```diff
@@ -350,6 +354,108 @@ def compare_hit_collections(file1, file2, tolerance=1e-9, tree_name='EVENT'):
...
+                    print(
+                        f"[{tree_name}] Entry {entry_idx}, branch '{b_name}' value mismatch"
+                    )
+                    is_identical = False
+
+    return is_identical
+
+
+def compare_root_metadata_and_runs(f1, f2):
+    """Compares 'metadata', 'runs', and 'podio_metadata' trees using pure PyROOT.
```
andresailer:
```suggestion
    """Compares 'metadata' and 'runs' trees using pure PyROOT.
```

DDG4/edm4hep/Geant4Output2EDM4hep.cpp
andresailer: This won't work if we have multiple different runs with different run parameters in flight, but that probably needs a major'ish refactoring when that happens.

DDTest/python/compare_simulation_output.py
```diff
@@ -350,6 +355,100 @@ def compare_hit_collections(file1, file2, tolerance=1e-9, tree_name='EVENT'):
     return True
 
 
+def compare_tree_contents(tree1, tree2, tree_name="tree"):
```
andresailer, line 358: the changes below were written by ♊ 

```
 to the "compare_simulation_output.py" in dd4hep add a function to compare metadata and runs frames of podio. 
```
Second prompt to re-write in pure ROOT.

DDG4/plugins/Geant4EventSeed.cpp
```diff
@@ -41,7 +41,6 @@ Geant4EventSeed::Geant4EventSeed(Geant4Context* c, const std::string& typ) : Gea
 									     m_type(typ),
 									     m_initialised(false)
 {
-  Geant4Action::runAction().callAtBegin(this,&Geant4EventSeed::begin);
```
andresailer, line 44: This was not needed, because the event seeder was added to the sequence, now begin is just called once for it.

DDG4/lcio/Geant4Output2LCIO.cpp
```diff
@@ -268,9 +264,14 @@ void Geant4Output2LCIO::saveRun(const G4Run* run)  {
   rh->parameters().setValue("DD4HEPVersion", versionString());
   rh->setRunNumber(m_runNo);
   rh->setDetectorName(context()->detectorDescription().header().name());
-  auto* parameters = context()->run().extension<RunParameters>(false);
-  if (parameters) {
-    parameters->extractParameters(*rh);
+  try {
```
andresailer: Can drop the exception catching again, I think.

andresailer:
Any more comments @tmadlener @wdconinc ? I dumped my patch into :gemini: and fixed a couple typos and bad use of unique_ptr::release instead of reset.

DDG4/src/Geant4EventAction.cpp
```diff
@@ -130,8 +127,8 @@ Geant4EventAction* Geant4EventActionSequence::get(const std::string& nam) const
 /// Add an actor responding to all callbacks. Sequence takes ownership.
 void Geant4EventActionSequence::adopt(Geant4EventAction* action) {
   if (action) {
-    G4AutoLock protection_lock(&event_action_mutex);
     action->addRef();
+    G4AutoLock protection_lock(mutex());
```
tmadlener, line 131: Moving the lock to after the ref-count increase does not change behavior, because the ref-counting already does the appropriate locking, resp. is in the end independent of adopting the action?

DDG4/include/DDG4/Geant4EventAction.h
```diff
@@ -57,6 +61,8 @@ namespace dd4hep {
     protected:
       /// Define standard assignments and constructors
       DDG4_DEFINE_ACTION_CONSTRUCTORS(Geant4EventAction);
+      /// Synchronisation lock
+      mutable G4Mutex m_mutex = G4MUTEX_INITIALIZER;
```
tmadlener, line 65: Just to confirm my understanding of the changes. This now makes every action carry a mutex, but behavior essentially doesn't change too much, because we just replace a static mutex for each output action with this one.

Are there other occurences of the same pattern for other actions?

DDG4/include/DDG4/Geant4Kernel.h
```diff
@@ -137,8 +137,6 @@ namespace dd4hep {
 
       /// Parent reference
       Geant4Kernel*      m_master         { nullptr };
-      /// Thread context reference
-      Geant4Context*     m_threadContext  { nullptr };
```
tmadlener, line 141: This is removed, because it is unused (essentially independent of the other changes)?

DDG4/lcio/Geant4Output2LCIO.cpp
```diff
@@ -248,7 +245,6 @@ void Geant4Output2LCIO::endRun(const G4Run* /*run*/)  {
 void Geant4Output2LCIO::commit( OutputContext<G4Event>& /* ctxt */)   {
   lcio::LCEventImpl* e = context()->event().extension<lcio::LCEventImpl>();
   if ( m_file )   {
-    G4AutoLock protection_lock(&action_mutex);
```
tmadlener, line 251: This seems to be common for the ROOT and LCIO output (and also for EDM4hep given that we never locked there in the first place). It's not immediately clear to me why we do not have to lock here anymore. Was it never necessary and we just remove the lock now, or did the locking go to wherever this is called from?

DDG4/plugins/Geant4EventSeed.cpp
tmadlener, line 1: Are these changes related to the rest, or could we in principle move them to a separate PR? At least from the changes here, it's not immediately clear where the seeder was added to the sequence.

DDG4/src/Geant4Output2ROOT.cpp
```diff
@@ -141,7 +140,6 @@ void Geant4Output2ROOT::beginRun(const G4Run* run) {
 
 /// Fill single EVENT branch entry (Geant4 collection data)
 int Geant4Output2ROOT::fill(const std::string& nam, const ComponentCast& type, void* ptr) {
-  std::lock_guard<std::mutex> lock(s_rootMutex);
```
tmadlener, line 144: Similar question as for `commit` (in the LCIO output and here). Why is it possible to remove this lock now?

DDTest/python/compare_simulation_output.py
```diff
@@ -350,6 +355,100 @@ def compare_hit_collections(file1, file2, tolerance=1e-9, tree_name='EVENT'):
...
+                if val1 == val2:
+                    equal = True
+                else:
+                    try:
+                        equal = list(val1) == list(val2)
+                    except Exception:
+                        try:
+                            equal = str(val1) == str(val2)
+                        except Exception:
+                            equal = False
```
tmadlener, line 415: I suppose this construct would also fail if `val1` and `val2` can neither be converted to a `list` or a `str` but would still compare equal in the appropriate type. Probably doesn't matter because all of that should be covered by the first comparison?

DDG4/include/DDG4/Geant4EventAction.h
andresailer: Before we had a static mutex in the output actions, and a different on when we called the output action, which lead to races. Now we use the same mutex inherited from the base class.

DDG4/include/DDG4/Geant4Kernel.h
andresailer: This was not used anywhere, it confused me when I was looking for contexts to attach extensions to.

DDG4/lcio/Geant4Output2LCIO.cpp
andresailer: The lock was moved to the calling location, because now we use the same lock in the calling location and in the output action, no longer the static mutex of the output action.

DDG4/plugins/Geant4EventSeed.cpp
andresailer: The seeder was added to the sequence in a previous PR, so this was just called twice back to back. Yes, can be moved to a separate PR.

DDG4/src/Geant4Output2ROOT.cpp
andresailer: see above. calling shared lock instance in base class

DDG4/plugins/Geant4EventSeed.cpp
andresailer: Moved to #1675

DDG4/src/Geant4EventAction.cpp
andresailer: I think I assume this is not going to be called for the same action in parallel. Otherwise we should lock this on the `action->mutex()` and the next one on the actionSequence mutex (as we do now).

DDTest/python/compare_simulation_output.py
andresailer: I think just checking `val1 == val2` should be good enough, because they should be the same type and content.

Thanks for having a look at that generated code in this function!

andresailer: I changed this part of the code.

DDG4/src/Geant4EventAction.cpp
andresailer: I added locking, either it costs almost nothing, or it prevents issues.

PR #1665, opened 2026-08-24
remove redundant calculations from i_collect and scanPhysicalVolume 

davidlange6:
Found that considerable time for idea geometry was going into 

::Populator::scanPhysicalVolume
dd4hep::detail::VolumeManager_Populator::scanPhysicalVolume
dd4hep::detail::GeoHandler::i_collect

This PR reduces that considerably by removing duplicate calculations by caching some results.

BEGINRELEASENOTES
- remove redundant calculations from i_collect and scanPhysicalVolume 
ENDRELEASENOTES

### DIFF
--- a/DDCore/src/GeoHandler.cpp
+++ b/DDCore/src/GeoHandler.cpp
@@ -147,6 +147,13 @@ detail::GeoHandler& detail::GeoHandler::i_collect(const TGeoNode* /* parent */,
                                                   const TGeoNode*    current,
                                                   int level, Region rg, LimitSet ls)
 {
+  /// Early-exit if this node has already been collected at this level.
+  /// Daughters of a TGeoVolume are the same TGeoNode pointers for every placement of
+  /// that volume; without this guard the subtree would be traversed once per placement
+  /// (N times for a volume placed N times) with all but the first producing no output.
+  if ( !(*m_set_data)[level].emplace(current).second )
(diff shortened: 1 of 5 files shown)


MarkusFrankATcernch:
Thanks a lot for this PR. This looks very promising!

jmcarcell:
What validation has been done here? Are simulation results unchanged? And what differences in time are we talking about?

andresailer:
> This PR reduces that considerably by removing duplicate calculations by caching some results.

That should mean you have numbers. Why not post those numbers?

MarkusFrankATcernch:
@jmcarcell @andresailer 
This code was not the most efficient, because in order to compute all combination the code explored all branches for all volumes, which is not always necessary. The resulting lookups should be the same.

What he did was to cache intermediate results from multiply placed logical volumes:
- If a placed branch of a logical volume contains no sensitives a second placement is ignored and no computation of the sensitive pathes is done. This holkds for all the passive stuff
- The unordered map caches the refs (aka TGeoVolumes) to all non sensitive branches.

Just as a reminder: because the VolumeMgr was relatively slow to populate all sorts of tricks got introduced. For example, the regex-based sensitive assignment was only done to cope with this behaviour (well also because the VolumeMgr is not really scalable: it grows with the number of sensitive physical volumes).

PR #1670, opened 2026-09-03
fix(Geant4UIManager): do not skip BeamOn when PreRunCommands are set

jbeirer:
BEGINRELEASENOTES

- Geant4UIManager: fix `PreRunCommands` preventing the normal batch `BeamOn` call.
- Geant4UIManager: separate `PreRunCommands` from the `Commands` property.
- Geant4UIManager: execute `PostRunCommands` after the automatic batch run.

ENDRELEASENOTES

## Summary

PreRunCommands are intended to configure Geant4 immediately before starting a run. However, they currently share the same internal storage as the Commands property. Since Commands historically assumes that the supplied commands control the run themselves, setting any PreRunCommands also prevents the normal batch `BeamOn` from happening.

This separates the two properties while keeping the existing behaviour of Commands for backwards compatibility. `PreRunCommands` are now executed before the normal batch run without suppressing it, while `Commands` continue to suppress the automatic `BeamOn`. `PostRunCommands` are also executed after the automatic batch run.

### DIFF
--- a/DDG4/include/DDG4/Geant4UIManager.h
+++ b/DDG4/include/DDG4/Geant4UIManager.h
@@ -64,6 +64,8 @@ namespace dd4hep {
       std::vector<std::string> m_initializeCommands;
       /// Property: List of commands to be executed when the Geant4Kernel gets terminated
       std::vector<std::string> m_terminateCommands;
+      /// Property: Legacy list of commands controlling the run execution themselves
+      std::vector<std::string> m_commands;
       /// Property: List of commands to be executed BEFORE running
       std::vector<std::string> m_preRunCommands;
       /// Property: List of commands to be executed AFTER running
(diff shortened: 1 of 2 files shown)


DDG4/include/DDG4/Geant4UIManager.h
```diff
@@ -64,6 +64,8 @@ namespace dd4hep {
       std::vector<std::string> m_initializeCommands;
       /// Property: List of commands to be executed when the Geant4Kernel gets terminated
       std::vector<std::string> m_terminateCommands;
+      /// Property: Legacy list of commands controlling the run execution themselves
```
andresailer: Why legacy?

andresailer: you are still throwing legacy around in the release notes and description.

jbeirer: should be fixed now :)

PR #1678, opened 2026-09-17
DDSim: outputconfig: fix example for multi-threaded running for user …

### LINKED ISSUE #1677: Duplication in the events saved in the output file using ddsim
### Check duplicate issues.

- [ ] Checked for duplicates

### Goal

Dear all,

I am opening this issue because I noticed that duplicated events are produced when running the simulation with the steering file available in FCC-Config:

https://github.com/HEP-FCC/FCC-config/blob/main/FCCee/FullSim/IDEA/IDEA_o1_v03/SteeringFile_IDEA_o1_v03.py

I suspect that this behaviour may be related to the new version of DD4hep.

I found a possible solution that consists of modifying the steering file by replacing the following configuration: 

```
def Geant4Output2EDM4hep_DRC_plugin(dd4hepSimulation):
    from DDG4 import EventAction, Kernel

    evt_root = EventAction(
        Kernel(), "Geant4Output2EDM4hep_DRC/" + dd4hepSimulation.outputFile, True
    )
    evt_root.Control = True
    output = dd4hepSimulation.outputFile
    evt_root.Output = output
    evt_root.enableUI()
    Kernel().eventAction().add(evt_root)
    return None
```

with 

```
def Geant4Output2EDM4hep_DRC_plugin(dd4hepSimulation):
    from DDG4 import EventAction, Kernel

    shared = dd4hepSimulation.numberOfThreads > 1
    evt_root = EventAction(
        Kernel(), "Geant4Output2EDM4hep_DRC/" + dd4hepSimulation.outputFile, shared
    )
    evt_root.Control = True
    output = dd4hepSimulation.outputFile
    evt_root.Output = output
    evt_root.enableUI()
    Kernel().eventAction().add(evt_root)
    return None
```

Cheers,
Andrea

Alma9

### compiler

GCC 14.2.0

### ROOT Version

6.38.00

### DD4hep Version

master with key4hep 2026-09-10

### Reproducer

```
source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh --spack 2026-09-10

wget https://raw.githubusercontent.com/key4hep/k4geo/master/example/SteeringFile_IDEA_o1_v03.py

ddsim --enableGun --gun.distribution uniform \
            --gun.momentumMin "0.01*GeV" --gun.momentumMax "1.00*GeV" \
            --gun.particle pi+ \
            --gun.multiplicity 2 \
            --random.enableEventSeed --random.seed 42 \
            --numberOfEvents 10 \
            --compactFile $K4GEO/FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml \
            --steeringFile SteeringFile_IDEA_o1_v03.py \
            --part.minimalKineticEnergy "0.00*MeV" \
            --outputFile out_sim_edm4hep_gun.root   
```

### Additional context

_No response_

andresailer:
Addressing #1677 

BEGINRELEASENOTES
- DDSim: OutputConfig: fix example for multi-threaded running for user output plugin.
- Geant4EventAction: throw an exception when shared actions are used in single-threaded mode to prevent silent wrong behavior. Fixes #1677 

ENDRELEASENOTES

TODO:

- [x] : maybe the Kernel() can be turned into singleton and we can get away without breaking backward compatibility. Kind of succeeded.


andread3vita:
Thank you, @andresailer! Just to better understand: is this PR intended as a workaround, or is it the definitive solution? If it is the latter, should we propagate the same change to FCC-Config?

andresailer:
I will see if there is something to make this automatic, but for now you should definitely apply this in the idea steering file. The output plugins available in dd4hep directly also behave like this at the moment.

PS: "behave like this" means that they would write events twice if marked as shared in multi-threaded mode, but the ddsim configuration only makes them shared in multithreaded mode.

andresailer:
@andread3vita 

Please have a look at the changes for the user output plugin here. I am afraid this is the only way to make this work, but I would also love to be proven wrong.

andread3vita:
> @andread3vita
> 
> Please have a look at the changes for the user output plugin here. I am afraid this is the only way to make this work, but I would also love to be proven wrong.

I tested this PR and it fixes the issue. I have one question about the kernel: is it necessary to use the Geant4 kernel?
If so, I will need to propagate this change to all the steering files in FCCConfig and k4geo, since the plugin now requires two inputs instead of one

andresailer:
Yes one has to use this `kernel()` , i.e., everything uses the same `kernel()`

DDG4/python/DDSim/Helper/OutputConfig.py
```diff
@@ -97,15 +97,17 @@ def exampleUserPlugin(dd4hepSimulation):
         '''
         from DDG4 import EventAction, Kernel
         dd = dd4hepSimulation  # just shorter variable name
-        evt_root = EventAction(Kernel(), 'Geant4Output2ROOT/' + dd.outputFile, True)
+        # Only use shared=True in MT mode to avoid double-save in ST mode
+        shared = dd.numberOfThreads > 1
+        evt_root = EventAction(Kernel().worker(), 'Geant4Output2ROOT/' + dd.outputFile, shared)
```
andresailer, line 102: @andread3vita 

So there is a way to do this without changing the function signature. We have to just use the `worker()` from the kernel, which is what the `geant4.kernel()` actually corresponds to.

Sorry for changing things back and forth.

andread3vita: Good news! This means that this PR will not affect other repositories! Thank you!

PR #1681, opened 2026-09-21
DDCAD: remove CSG-based exporter

jmcarcell:
ROOT made RCsg and CsgOps.h private in [root-project/root#23421](https://github.com/root-project/root/pull/23421). DDCAD's ASSIMP writer uses that API to tessellate every exported shape, so it can no longer consume a current ROOT installation.

Question, what to do with this? This PR is the removal of what is using this CSG exporter. An alternative would be to make it a dependency of DD4hep.

BEGINRELEASENOTES
 - Remove the CSG-dependent CAD export path because ROOT's RCsg library is now private.

ENDRELEASENOTES


andresailer:
I am not convinced that simply removing the assimpwriter is the correct way forward. either we find an alternative library, or complain to ROOT to keep Rcsg public.

ferdymercury:
> we find an alternative library

side note: Csg was derived from https://github.com/dtecta/solid3 so it could be maybe that one.

side question: would it be ok to have it in ROOT/CsgOps.h rather than CsgOps.h ?

andresailer:
I may have some implementation that just uses publicly available ROOT interfaces, but I need to do some more testing.

andresailer:
my (and :gemini:'s) implementation: #1682

andresailer:
I won't get the other mess sorted, so I will merge this to get the builds green and then revert it to get the functionality back.

PR #1682, opened 2026-09-21
AssimpWriter: Drop need for rootcsg

andresailer:
BEGINRELEASENOTES
- AssimpWriter: replace use of now private ROOTCsg library (:gemini:)
- AssimpWriter: fix bug in filling color information for vertices. Previously only the color for the first vertex was written instead of for all of them.

ENDRELEASENOTES


andresailer:
While this passes all the tests, it doesn't correctly reproduce the mesh for boolean shapes and just dumps all of the boolean shapes participants as unions, and not intersections or subtractions.

DDCAD/src/ASSIMPWriter.cpp
```diff
@@ -102,136 +95,112 @@ namespace  {
...
-    auto right_mesh = std::unique_ptr<RootCsg::TBaseMesh>(make_mesh(right));
-    *glmat = &copy;
-
-    switch (oper) {
-    case TGeoBoolNode::kGeoUnion:
-      return RootCsg::BuildUnion(left_mesh.get(), right_mesh.get());
-    case TGeoBoolNode::kGeoIntersection:
-      return RootCsg::BuildIntersection(left_mesh.get(), right_mesh.get());
-    case TGeoBoolNode::kGeoSubtraction:
-      return RootCsg::BuildDifference(left_mesh.get(), right_mesh.get());
```
ferdymercury: Did you check if TGLScenePad::BuildComposite() could work for this purpose?
Otherwise, we could create functions in TGeoTesselated forwarding these calls maybe?

andresailer: I think rather the GetBuffer3D from TGeoCompositeShape should be forwarding these calls.

andresailer: The use case we have is wanting to tessellate a whole geometry consisting of arbitrarty shapes. So we go over the whole geometry tree and getting triangles. This fails for CompositeShapes because the boolean solids are not resolved.

